PR #14, opened 2017-03-22
External toolchain

qwattash:
Change compiler flags for the kernel build with the external clang toolchain
Forward external toolchain args to the freebsd install targets


RichardsonAlex:
Thanks!

Issue #26, opened 2018-03-19
Qemu build needs pixman dependency


xrayn:
Hi,

I just tried to build the sdk via `cheribuild.py sdk` on FreeBSD 11

```
FreeBSD  11.0-RELEASE-p1 FreeBSD 11.0-RELEASE-p1 #0 
r306420: Thu Sep 29 01:43:23 UTC 2016   
root@releng2.nyi.freebsd.org:/usr/obj/usr/src/sys/GENERIC  amd64
```
Apparently qemu is missing a dependency to pixman.
After installing it via `pkg install pixman` qemu compiled without any further issues.

Perhaps add it as an additional dependency in https://github.com/CTSRD-CHERI/cheribuild/blob/master/pycheribuild/projects/build_qemu.py
See attached console output.

```
Cloning into '/home/ar/cheri/qemu'...                                  
remote: Counting objects: 337590, done.                                      
remote: Total 337590 (delta 0), reused 0 (delta 0), pack-reused 337589                                                                                                                        
Receiving objects: 100% (337590/337590), 163.86 MiB | 2.05 MiB/s, done.                                                                                                                        
Resolving deltas: 100% (269961/269961), done.                                                                                                                                                  
Already up-to-date.                                                                                                          
Configuring QEMU ...                                                       
cd /home/ar/cheri/build/qemu-build && /home/ar/cheri/qemu/configure --target-list=cheri256-softmmu,cheri128-softmmu,cheri128magic-softmmu --disable-linux-user --disable-bsd-user --disable-xen
 --disable-docs --disable-rdma --disable-werror '--extra-cflags=-O3 -I/usr/local/include/glib-2.0 -I/usr/local/lib/glib-2.0/include -I/usr/local/include' --disable-stack-protector --disable-l
inux-aio --disable-kvm --prefix=/home/ar/cheri/output/sdk             
Saving build log to /home/ar/cheri/build/qemu-build/configure.log      
                                                                    
ERROR: pixman >= 0.21.8 not present. Your options:                    
         (1) Preferred: Install the pixman devel package (any recent      
             distro should have packages as Xorg needs pixman too).      
         (2) Fetch the pixman submodule, using:                     
             git submodule update --init pixman                        
                                                                     
Command "/home/ar/cheri/qemu/configure --target-list=cheri256-softmmu,cheri128-softmmu,cheri128magic-softmmu --disable-linux-user --disable-bsd-user --disable-xen --disable-docs --disable-rdm
[...]

PR #27, opened 2018-03-19
Dependencies for FreeBSD builds added

xrayn:
- Adds pixmap as a dependency for building of qemu

  - Fixes Issue #26 

- Added new dependencies a section in README
  - command line  added to enable successful build of sdk & qemu


xrayn:
fix is more complicated. It's a library and so not supported by current mechanisms

PR #29, opened 2018-04-06
elftoolchain: use out-of-tree build support

nwf:
Together with the patches at https://github.com/nwf/elftoolchain , this enables out-of-tree builds for elftoolchain.  Those may not be complete, but they worked for me. ;)

I care because I use a single collection of checkouts shared with a guest (FreeBSD/amd64) VM, and want, I think, the full `sdk` on the host as well as a `cheribsd-sdk` on the guest.


arichardson:
Thanks, that looks good. However, I don't think I can merge this until the patches are included in the upstream elftoolchain branch. For existing checkouts the build will suddenly start failing because they are not using your elftoolchain fork.

PR #34, opened 2018-05-15
A few small changes

nwf:
Mostly to elftoolchain build.  I found it necessary to run "make obj" on my FreeBSD VM, for example.  The rest are mostly superficial.

### DIFF
--- a/pycheribuild/projects/elftoolchain.py
+++ b/pycheribuild/projects/elftoolchain.py
@@ -91,8 +91,9 @@ class BuildElftoolchain(Project):
             version = get_program_version(Path(self.make_args.command), ("-V", "MAKE_VERSION"), regex=b"(.+)")[0]
             # The version of bmake shipped with ubuntu doesn't handle depedencies correctly
             if version > 20170101:
-                statusUpdate("Note: Working around old version of bmake: ", version)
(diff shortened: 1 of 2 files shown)


arichardson:
I believe I fixed the Cmake issue, is that change still required for you? Otherwise looks good to me

pycheribuild/projects/elftoolchain.py
```diff
@@ -91,8 +91,9 @@ def compile(self, **kwargs):
             version = get_program_version(Path(self.make_args.command), ("-V", "MAKE_VERSION"), regex=b"(.+)")[0]
             # The version of bmake shipped with ubuntu doesn't handle depedencies correctly
             if version > 20170101:
-                statusUpdate("Note: Working around old version of bmake: ", version)
                 is_old_broken_bmake = False
+            else :
```
arichardson: removeextra space before the `:`

nwf:
Trade you: dropped the cmake hunk but added one to disk_image.  The issue there is that wget can't create its temporary index files over smbfs, which leads to sadness for those of us crazy enough to have extra-files so mounted.

arichardson:
Thanks!

PR #38, opened 2018-08-02
Add bodiagsuite project

nwf:


arichardson:
Thanks! I guess at some point we will want a git repo for it?

nwf:
Probably.  I don't think I can create one in `CTSRD-CHERI`, but feel free to steal mine.

PR #41, opened 2018-11-03
Update cheribuild.py usage in README.

kwitaszczyk:
Currently README states that cheribuild.py builds the whole software stack for 256-bit CHERI by default. This behaviour was changed in 880bbe0.

While we're here, I'd like to update cheribuild.py usage output, fit it in 80 columns so that it doesn't require horizontal scroll on GitHub and use the tilde symbol instead of user-specific paths.


arichardson:
Thanks!

PR #46, opened 2019-06-18
DNM: minimal disk-image: use minimal rootfs

nwf:
This half-corrects a problem, in that it now points at the correct
rootfs dir, and some more blank directories are made, but the builds
will fail because not all the requisite targets for the files listed
in cheribsdbox.manifest are currently built by the cheribsd-minimal
target.

The PR is not, as such, ready to merge, but I would like to make sure
I'm not barking up the wrong tree.


arichardson:
cheribsd-minimal doesn't really work. I should probably delete it. Disk-image-minimal just uses a subset of the normal disk image and that works quite well.
If FreeBSD hat a sane build system I could build only cheribsdbox but that seems unlikely any time soon.

PR #48, opened 2019-07-21
snmalloc build target

nwf:


pycheribuild/projects/cross/snmalloc.py
```diff
@@ -0,0 +1,87 @@
+#
+# Copyright (c) 2017 Alex Richardson
```
arichardson: You probably want to fix this.

pycheribuild/projects/cross/snmalloc.py
```diff
@@ -0,0 +1,87 @@
...
+# All rights reserved.
+#
+# This software was developed by SRI International and the University of
+# Cambridge Computer Laboratory under DARPA/AFRL contract FA8750-10-C-0237
+# ("CTSRD"), as part of the DARPA CRASH research programme.
```
arichardson: ECATS?

pycheribuild/projects/cross/snmalloc.py
```diff
@@ -0,0 +1,87 @@
...
+    projectName = "snmalloc"
+    repository = GitRepository("https://github.com/Microsoft/snmalloc")
+    crossInstallDir = CrossInstallDir.CHERIBSD_ROOTFS
+    appendCheriBitsToBuildDir = True
+    supported_architectures = [CrossCompileTarget.CHERI, CrossCompileTarget.NATIVE, CrossCompileTarget.MIPS]
+    defaultOptimizationLevel = ["-O2"]
+    default_build_type = BuildType.DEBUG
+    defaultCMakeBuildType = "Debug"
+
+    dependencies = ["cheribsd-sysroot"]
```
arichardson: This should be implicit for CrossCompileCMakeProject

pycheribuild/projects/cross/snmalloc.py
```diff
@@ -0,0 +1,87 @@
...
+            self.pagemap_rederive = True
+
+        if self.pagemap_rederive:
+            self.pagemap_pointers = True
+
+        self.COMMON_FLAGS.append("-DSNMALLOC_PAGEMAP_POINTERS=%d"   % (1 if self.pagemap_pointers else 0))
+        self.COMMON_FLAGS.append("-DSNMALLOC_PAGEMAP_REDERIVE=%d"   % (1 if self.pagemap_rederive else 0))
+        self.COMMON_FLAGS.append("-DSNMALLOC_CHERI_SETBOUNDS=%d"    % (1 if self.cheri_bounds     else 0))
+        self.COMMON_FLAGS.append("-DSNMALLOC_QUARANTINE_DEALLOC=%d" % (1 if self.quarantine       else 0))
+        self.COMMON_FLAGS.append("-DSNMALLOC_REVOKE_QUARANTINE=%d"  % (1 if self.revoke           else 0))
```
arichardson: Are there CMake options for these flags? If so I would use self.add_cmake_option() instead (which would handle boolean values)

nwf: I hadn't added them, but I can.

jrtc27: If they're real boolean values (i.e. you're not relying on `__bool__()` or `None`) then `%d` will do the coercion for you (and even if it didn't, `int(foo)` does).

pycheribuild/projects/cross/snmalloc.py
```diff
@@ -0,0 +1,87 @@
...
+# SUCH DAMAGE.
+#
+import shutil
+
+from .crosscompileproject import *
+from ...utils import getCompilerInfo, runCmd, IS_FREEBSD
+
+class SNMalloc(CrossCompileCMakeProject):
+    projectName = "snmalloc"
+    repository = GitRepository("https://github.com/Microsoft/snmalloc")
```
arichardson:
```suggestion
    repository = GitRepository("https://github.com/nwf/snmalloc")
```

pycheribuild/projects/cross/snmalloc.py
```diff
@@ -0,0 +1,87 @@
...
+            self.cheri_bounds     = True
+            self.pagemap_rederive = True
+
+        if self.cheri_bounds:
+            self.pagemap_rederive = True
+
+        if self.pagemap_rederive:
+            self.pagemap_pointers = True
+
+        self.COMMON_FLAGS.append("-DSNMALLOC_PAGEMAP_POINTERS=%d"   % (1 if self.pagemap_pointers else 0))
```
arichardson:
```suggestion
        self.COMMON_FLAGS.append("-DSNMALLOC_PAGEMAP_POINTERS=%d"  % self.pagemap_pointers)
```

PR #49, opened 2019-08-06
Add --benchmark-fpga-extra-args

nwf:
Useful for specifying the cable id when a system has multiple attached
FPGAs, for example: --benchmark-fpga-extra-args="-c 2"


arichardson:
Thanks! It might be awkward if someone want to add additonal options to the runbench command but for now that should be fine. Eventually we can stop using beri-fpga-bsd-boot.py and then a single option will be fine

PR #50, opened 2019-09-02
beri-fpga-bsd-boot: make -k an argument to parser

brettferdosi:
It is currently an argument to runbench but used in the boot_bsd()
function, which is called by bootonly as well as runbench.

Making it an argument to parser is the clean solution but will break existing scripts - perhaps it is better to add it as an argument to runbench, bootonly, and whatever other subcommands use it?


arichardson:
I think this is a good change but I'm reluctant to update all Jenkins jobs.

Does it work if we have it both as an argument to the root parser and the runbench parser?
That way we get the sensible new behaviour, and the Jenkins jobs that use runbench still work.

brettferdosi:
Seems to work locally.

PR #51, opened 2019-09-25
beri-fpga-bsd-boot: on panic, type "bt" at the debug prompt

nwf:
Untested; does this look OK?


arichardson:
Should be fine as long as panicstr is not a regex. I've been planning to merge most of this script with the QEMU boot one. Once that's done you'll get the `bt` feature for free.

But until then this hopefully helps.

PR #52, opened 2019-10-18
Support to compile newlib for RTEMS and RISC-V with Clang

heshamelmatary:


pycheribuild/projects/cross/newlib_rtems_riscv.py
```diff
@@ -0,0 +1,119 @@
...
+from ...utils import statusUpdate, IS_MAC, runCmd
+from ...config.loader import ComputedDefaultValue
+from pathlib import Path
+import tempfile
+
+
+class BuildNewlibRtemsRiscv(CrossCompileAutotoolsProject):
+    repository = GitRepository("https://github.com/CTSRD-CHERI/newlib")
+    projectName = "newlib-rtems-riscv"
+    make_kind = MakeCommandKind.GnuMake
```
arichardson: You probably want to set gitBranch here?

heshamelmatary:
> In general seems fine, I just wonder if we can share code with the existing newlib target?
That will need at least two new flags passed to cheribuild: one to compile for RISC-V, and the other to compile for RTEMS. Additional -march and -mabi might also be needed for future RISC-V targets, but that will take considerable refactoring effort to cheribuild

jrtc27:
I think Alex's thinking is that any RISC-V-specific bits should go in BuildNewlibBaremetal, and any RTEMS-specific bits should go in BuildNewlibRtemsRiscv, which should be a subclass of BuildNewlibBaremetal.

jrtc27:
Also I believe `projectName = "newlib-rtems-riscv"` is going to mean that you have newlib-rtems-riscv ie newlib-rtems-riscv-riscv as well as newlib-rtems-riscv-mips and newlib-rtems-riscv-cheri projects, which you probably don't want?

pycheribuild/projects/cross/newlib_rtems_riscv.py
```diff
@@ -0,0 +1,119 @@
...
+    rtems = True
+    add_host_target_build_config_options = False
+    # TODO Make this -O2 once a bug with LLVM/Clang relaxation is fixed
```
jrtc27: I thought our believe was that it was because you compiled without `-mno-relax` and the linker used was LLD. Yes, `-mno-relax` currently messes up `%pcrel_lo` on master, but the fact that everything works at `-O0` is purely coincidence and has a 50% chance of failure (assuming this is a `.balign 4` that may or may not be aligned correctly pre-relaxation and gets put into `mtvec`.).

heshamelmatary: > I think Alex's thinking is that any RISC-V-specific bits should go in BuildNewlibBaremetal, and any RTEMS-specific bits should go in BuildNewlibRtemsRiscv, which should be a subclass of BuildNewlibBaremetal.

RTEMS has its own GCC-newlib toolchain, so technically newlib-rtems isn't baremetal (and it doesn't use libgloss). That said, I am fine making it a subcalss of baremetal as it'll make things way easier for me

heshamelmatary:
Closed in favor of https://github.com/CTSRD-CHERI/cheribuild/pull/62

PR #54, opened 2019-11-07
Several beri-fpga-bsd-boot fixes

nwf:


beri-fpga-bsd-boot.py
```diff
@@ -694,7 +694,7 @@ def common_boot (kernel_img=args.kernel_img,addr=args.kernel_addr,bitfile=args.b
 #################
 
 def main():
-    if args.interact or args.subcmd == "console":
+    if vars(args).get("interact",False) or args.subcmd == "console":
```
arichardson: This could probably be `getattr(args, "interact", False)` instead

beri-fpga-bsd-boot.py
```diff
@@ -518,7 +518,7 @@ def do_network_on(console: boot_cheribsd.CheriBSDInstance, args, timeout=300):
     console.expect_prompt()
     # No longer needed? console.run('/sbin/ifconfig {} polling'.format(ifc))
     console.sendline('/sbin/dhclient {}'.format(ifc))
-    console.expect(["turning network on","bound to .* -- renewal in .*"], timeout=60)
+    console.expect(["bound to .* -- renewal in .*\."], timeout=60)
```
arichardson: Good catch! I saw some odd failures in the past, probably caused by this.

PR #57, opened 2020-01-07
llvm: use internal c++ goodies

nwf:
Add libcxx, libcxxabi, compiler-rt, and libunwind to the llvm target.
This avoid the need to use the host equivalents, and works around the
fact that newish GCC's bits/regex.tcc names a variable "__output", which
we take to be a reserved word.

### DIFF
--- a/pycheribuild/projects/llvm.py
+++ b/pycheribuild/projects/llvm.py
@@ -73,7 +73,7 @@ class BuildLLVMBase(CMakeProject):
                      "needed by default (e.g. llvm-mca, llvm-pdbutil)")
         cls.build_everything = cls.add_bool_option("build-everything", default=False,
             help="Also build documentation,examples and bindings")
-        cls.use_llvm_cxx = cls.add_bool_option("use-tree-cxx", default=False,
+        cls.use_llvm_cxx = cls.add_bool_option("use-in-tree-cxx-libs", default=False,
             help="Use in-tree, not host, C++ runtime")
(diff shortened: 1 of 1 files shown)


jrtc27:
Doesn't this pose problems if you use that compiler to link against things that use libstdc++?

jrtc27:
(Ditto libunwind and compiler-rt/libgcc)

nwf:
As far as I can tell, the resulting compiler builds at least cheribsd fine.  I could believe there are problems I haven't yet tripped over, tho'.

nwf:
AFAICT the resulting compiler builds `cheribsd` just fine, at least?

arichardson:
I think we probably want to hide this behind a flag since the default should be to build against the host C++ standard library

pycheribuild/projects/llvm.py
```diff
@@ -73,6 +73,10 @@ def setup_config_options(cls, useDefaultSysroot=True):
                      "needed by default (e.g. llvm-mca, llvm-pdbutil)")
         cls.build_everything = cls.add_bool_option("build-everything", default=False,
             help="Also build documentation,examples and bindings")
+        cls.use_llvm_cxx = cls.add_bool_option("use-tree-cxx", default=False,
```
arichardson:
```suggestion
        cls.use_llvm_cxx = cls.add_bool_option("use-in-tree-cxx-libs", default=False,
```

PR #62, opened 2020-03-03
Generalize Newlib and add support to build it for RTEMS RISCV64 purecap

heshamelmatary:


pycheribuild/config/target_info.py
```diff
@@ -567,6 +567,61 @@ def get_rootfs_target(self) -> "Project":
...
+        """Additional linker flags that need to be passed when building an executable (e.g. custom linker script)"""
+        return [""]
+
+    @property
+    def is_baremetal(self):
+        return False
+
+    @property
+    def is_rtems(self):
+        return True
```
arichardson: A default implementation returning false is needed in the base classes (see jenkins failure)

pycheribuild/config/target_info.py
```diff
@@ -729,6 +784,10 @@ class CompilationTargets(object):
     FREEBSD_AARCH64 = CrossCompileTarget("aarch64", CPUArchitecture.AARCH64, FreeBSDTargetInfo)
     FREEBSD_X86_64 = CrossCompileTarget("x86_64", CPUArchitecture.X86_64, FreeBSDTargetInfo)
 
+    # RTEMS targets
+    RTEMS_NEWLIB_RISCV64 = CrossCompileTarget("rtems-riscv64", CPUArchitecture.RISCV64,
```
arichardson: I would probably call this rtems-riscv64-purecap so we can add a hybrid/plain riscv version later.
```suggestion
    RTEMS_NEWLIB_RISCV64 = CrossCompileTarget("rtems-riscv64-purecap", CPUArchitecture.RISCV64,
```

heshamelmatary: I don't think we will need to build newlib in vanilla and/or hybrid mode for RTEMS-CHERI-RISC-V in cheribuild, that's why I omitted it and assumed that target is always gonna be purecap

pycheribuild/projects/cross/newlib.py
```diff
@@ -118,31 +120,40 @@ def configure(self):
...
+              # FIXME: these don't seem to work
+              "--enable-serial-build-configure",
+              "--enable-serial-target-configure",
+              "--enable-serial-host-configure",
+          ])
+        elif self.target_info.is_rtems:
+          self.configureArgs.extend([
+              "--enable-newlib-io-c99-formats",
+              "--disable-libgloss",
+              "--disable-static", "--enable-shared",
```
arichardson: I'm a bit surprised by the --enable-shared? Don't we wan't it static?

heshamelmatary: Nice catch. Yeah, definitely should be static. I fail to remember where/why did these flags come from, but they don't seem to be doing what they should (i.e., I always get static libs). I deleted them.

PR #64, opened 2020-03-04
Add preliminary support for RTEMS and rework compiler-rt 

heshamelmatary:


pycheribuild/projects/project.py
```diff
@@ -2573,7 +2573,7 @@ def configure(self, **kwargs):
             # CMAKE_CROSSCOMPILING will be set when we change CMAKE_SYSTEM_NAME:
             # This means we may not need the toolchain file at all
             # https://cmake.org/cmake/help/latest/variable/CMAKE_CROSSCOMPILING.html
-            system_name = "Generic" if self.target_info.is_baremetal else "FreeBSD"
+            system_name = "Generic" if self.target_info.is_baremetal or self.target_info.is_rtems else "FreeBSD"
```
arichardson: This is fine for now, but longer-term we should add `self.target_info.cmake_system`

pycheribuild/config/target_info.py
```diff
@@ -791,6 +849,9 @@ class CompilationTargets(object):
     # RTEMS targets
     RTEMS_NEWLIB_RISCV64 = CrossCompileTarget("rtems-riscv64-purecap", CPUArchitecture.RISCV64,
         NewlibRtemsTargetInfo, is_cheri_purecap=True)
+    RTEMS_RISCV64_PURECAP = CrossCompileTarget("rtems-riscv64-purecap", CPUArchitecture.RISCV64, RTEMSTargetInfo,
```
arichardson: Why do you need both rtems and rtems+newlib? I though newlib is always used?

pycheribuild/projects/cross/compiler-rt-builtins.py
```diff
@@ -0,0 +1,102 @@
+#
```
arichardson: I would be tempted to merge these two files since they are building the same project, just different subsets.

pycheribuild/projects/project.py
```diff
@@ -2622,7 +2623,10 @@ def configure(self, **kwargs):
             # CMAKE_CROSSCOMPILING will be set when we change CMAKE_SYSTEM_NAME:
             # This means we may not need the toolchain file at all
             # https://cmake.org/cmake/help/latest/variable/CMAKE_CROSSCOMPILING.html
-            system_name = "Generic" if self.target_info.is_baremetal else "FreeBSD"
```
arichardson, line 2625: This should probably be an abstract property in TargetInfo (similar to `cmake_processor_id`, but that change can be made later.

PR #65, opened 2020-04-03
Don't set XAS.

brooksdavis:
Setting it to an invalid value breaks native builds and we'd already
commented out setting it elsewhere.


brooksdavis:
A quick `cheribuild.py cheribsd-purecap` succeeded with this change.

PR #67, opened 2020-04-07
add target for poco c++ lib

dodsonmg:
Doesn't build successfully yet.

### DIFF
--- a/pycheribuild/projects/cross/poco.py
+++ b/pycheribuild/projects/cross/poco.py
@@ -1,0 +1,53 @@
+#
+# Copyright (c) 2017 Alex Richardson
+# All rights reserved.
+#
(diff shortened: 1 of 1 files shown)


pycheribuild/projects/cross/poco.py
```diff
@@ -0,0 +1,53 @@
+#
+# Copyright (c) 2017 Alex Richardson
```
arichardson, line 2: That's not quite right

dodsonmg: fixed.  2017->2020

dodsonmg:
This builds with the patch applied to cheribsd lib/libcxxrt proposed by @arichardson:
https://github.com/CTSRD-CHERI/cheribsd/pull/421

PR #68, opened 2020-04-07
Add -DWITH_DTRACE_TEST cmd line option

nicomazz:
in addition to -DWITH_TESTS dtrace also requires -DWITH_DTRACE_TESTS.

Other than this, there are minor fixes to boot_and_login.
The question is: should the boot be stopped when there is a USER_CHERI_EXCEPTION?
if so, then the change in boot_cheribsd/__init__.py should be reverted

### DIFF
--- a/pycheribuild/__main__.py
+++ b/pycheribuild/__main__.py
@@ -235,7 +235,8 @@ def real_main():
 def main():
     error = False
     try:
-        os.setpgrp()  # create new process group, become its leader
+        if os.getpgrp() != os.getpid():
+            os.setpgrp()  # create new process group, become its leader
         real_main()
     except KeyboardInterrupt:
         error = True
--- a/pycheribuild/boot_cheribsd/__init__.py
+++ b/pycheribuild/boot_cheribsd/__init__.py
@@ -545,13 +545,13 @@ def boot_and_login(child: CheriBSDInstance, *, starttime, kernel_init_only=False
         boot_expect_strings = [LOGIN, SHELL_OPEN, BOOT_FAILURE]
         i = child.expect(boot_expect_strings + ["DHCPACK from "] + FATAL_ERROR_MESSAGES, timeout=15 * 60,
             timeout_msg="timeout awaiting login prompt")
-        if i == len(boot_expect_strings):
+        if i == len(boot_expect_strings):  # DHCPACK from
             have_dhclient = True
             success("===> got DHCPACK")
             # we have a network, keep waiting for the login prompt
             i = child.expect(boot_expect_strings + FATAL_ERROR_MESSAGES, timeout=5 * 60,
                 timeout_msg="timeout awaiting login prompt")
-        if i == 0:
+        if i == boot_expect_strings.index(LOGIN):
             success("===> got login prompt")
             child.sendline("root")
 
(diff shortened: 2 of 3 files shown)


pycheribuild/boot_cheribsd/__init__.py
```diff
@@ -513,13 +513,13 @@ def boot_and_login(child: CheriBSDInstance, *, starttime, kernel_init_only=False
...
             timeout_msg="timeout awaiting login prompt")
-        if i == len(boot_expect_strings):
+        if i == len(boot_expect_strings):  # DHCPACK from
             have_dhclient = True
             success("===> got DHCPACK")
             # we have a network, keep waiting for the login prompt
             i = child.expect(boot_expect_strings + FATAL_ERROR_MESSAGES, timeout=5 * 60,
                 timeout_msg="timeout awaiting login prompt")
-        if i == 0:
+        if i == boot_expect_strings.index(LOGIN):
```
arichardson, line 522: That's a good idea, hardcoding the indices is very fragile.

arichardson:
I think --test should fail if we get a capability violation during boot, but other than that the test change looks good.

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -148,6 +148,11 @@ def setup_config_options(cls, **kwargs):
             cls.build_tests = cls.add_bool_option("build-tests", help="Build the tests too (-DWITH_TESTS)",
                 show_help=True, default=True)
 
+        if "build_dtrace_tests" not in cls.__dict__:
```
arichardson, line 151: You probably don't need this check since it's unlikely that a subclass would define it to be always true/false

nicomazz:
> I think --test should fail if we get a capability violation during boot, but other than that the test change looks good.

Yeah I fixed the problem and reverted the change!

nicomazz:
I completely removed the `--build-dtrace-tests` flag. Instead, when WITH_TESTS is on, also WITH_DTRACE_TESTS is on

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -563,8 +564,9 @@ def buildworld_args(self) -> MakeOptions:
     def kernel_make_args_for_config(self, kernconf: str, extra_make_args) -> MakeOptions:
         self._setup_make_args()  # ensure make args are complete
         kernel_options = self.make_args.copy()
-        if self.compiling_for_mips(include_purecap=True):
-            # Don't build kernel modules for MIPS
+        if self.compiling_for_mips(include_purecap=True) and not self.build_tests:
+            # Don't build kernel modules for MIPS, except if we are building tests
+            # (DTrace tests require the 'dtrace_test' loadable module)
             kernel_options.set(NO_MODULES="yes")
```
arichardson, line 570: I think we shouldn't bother building modules by default.
Maybe set `MODULES_OVERRIDE=dtrace_test` (https://www.freebsd.org/doc/handbook/kernelconfig-building.html) `if self.build_tests` and otherwise keep the NO_MODULES

arichardson:
Minor comments, but looks good to be merged once you rebase it.

pycheribuild/projects/cross/cheribsd.py
nicomazz: I'll change it so that it can be compiled directly into the kernel, and not as a module

pycheribuild/__main__.py
```diff
@@ -235,8 +235,7 @@ def real_main():
 def main():
     error = False
     try:
-        if os.getpgrp() != os.getpid():
-            os.setpgrp()  # create new process group, become its leader
+        os.setpgrp()  # create new process group, become its leader
```
arichardson: Did you mean to revert this change?

arichardson:
Looks good, but could you do a rebase to drop all the unnecessary commits?

PR #69, opened 2020-04-07
Add target for Eigen linear algebra libraries

dodsonmg:
[Eigen's website](http://eigen.tuxfamily.org/index.php?title=Main_Page)

### DIFF
--- a/pycheribuild/projects/cross/eigen.py
+++ b/pycheribuild/projects/cross/eigen.py
@@ -37,16 +37,3 @@ class BuildEigen(CrossCompileCMakeProject):
     repository = GitRepository("https://gitlab.com/libeigen/eigen.git")
     native_install_dir = DefaultInstallDir.IN_BUILD_DIRECTORY
     cross_install_dir = DefaultInstallDir.SYSROOT
-
-    # @classmethod
-    # def setup_config_options(cls, **kwargs):
-    #     super().setup_config_options(**kwargs)
-    
-    # def __init__(self, config: CheriConfig, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


pycheribuild/projects/cross/eigen.py
```diff
@@ -0,0 +1,52 @@
...
+# from ..project import ReuseOtherProjectRepository
+
+
+class BuildEigen(CrossCompileCMakeProject):
+    project_name = "eigen"
+    repository = GitRepository("https://gitlab.com/libeigen/eigen.git")
+    native_install_dir = DefaultInstallDir.IN_BUILD_DIRECTORY
+    cross_install_dir = DefaultInstallDir.SYSROOT
+
+    @classmethod
```
arichardson, line 41: None of these methods need to be defined if they just call the superclass.

pycheribuild/projects/cross/eigen.py
```diff
@@ -0,0 +1,52 @@
+#
+# Copyright (c) 2020 Alex Richardson
```
arichardson, line 2: You probably want your copyright here instead of listing me.

dodsonmg: Well, you wrote cheribuild and the [whatever].py script on which eigen.py was based.  I feel like my modification of the script is similar to providing a command line argument.

pycheribuild/projects/cross/eigen.py
dodsonmg: removed.

PR #75, opened 2020-04-17
Use virtio networking on non-mips.

brooksdavis:
This work well enough to get connectivity on riscv.


arichardson:
LGTM. Does it also work on MIPS?

brooksdavis:
yes, mips still works.

arichardson:
I meant whether virtio networking would also work on MIPS. If so, I guess we should use it there too.

brooksdavis:
> I meant whether virtio networking would also work on MIPS. If so, I guess we should use it there too.

No, it is broken on BE systems other than powerpc.

PR #77, opened 2020-04-20
Avoid unnecessary updates

brooksdavis:
Don't perform any updates if the current branch doesn't have an upstream.  Also, avoid an attempt to update if our current revision is equal to that of the upstream branch.  Most usefully, that avoids a prompt to stash outstanding changes and update when there is no work to do.

### DIFF
--- a/pycheribuild/projects/project.py
+++ b/pycheribuild/projects/project.py
@@ -1295,6 +1295,15 @@ class GitRepository(SourceRepository):
                     if current_project.query_yes_no("Update to correct URL?"):
                         runCmd("git", "remote", "set-url", "origin", self.url, runInPretendMode=True, cwd=src_dir)
 
+        # see if we have an upstream
+        head = runCmd("git", "symbolic-ref", "-q", "HEAD",
+                      captureOutput=True, cwd=src_dir, print_verbose_only=True, runInPretendMode=True).stdout.strip().decode("utf-8")
+        upstream_branch = runCmd("git", "for-each-ref", '--format=%(upstream:short)', head, "origin/mainline",
+                                 captureOutput=True, cwd=src_dir, print_verbose_only=True, runInPretendMode=True).stdout.strip().decode("utf-8")
+        if len(upstream_branch) == 0:
+            print(coloured(AnsiColour.blue, "No upstream to update from"))
+            return
+
         # make sure we run git stash if we discover any local changes
         has_changes = len(runCmd("git", "diff", "--stat", "--ignore-submodules",
                                  captureOutput=True, cwd=src_dir, print_verbose_only=True).stdout) > 1
--- a/pycheribuild/projects/project.py
+++ b/pycheribuild/projects/project.py
@@ -1304,6 +1304,18 @@ class GitRepository(SourceRepository):
             print(coloured(AnsiColour.blue, "No upstream to update from"))
             return
 
+	# make sure there's work to do before touching the repo
+        upstream_remote = upstream_branch.split("/", 1)[0]
+        print(coloured(AnsiColour.red, "upstream_remote " + upstream_remote))
+        runCmd("git", "fetch", upstream_remote)
+        head_rev = runCmd("git", "rev-parse", "HEAD",
+                          captureOutput=True, cwd=src_dir, print_verbose_only=True, runInPretendMode=True).stdout.strip().decode("utf-8")
+        upstream_rev = runCmd("git", "rev-parse", upstream_branch,
+                              captureOutput=True, cwd=src_dir, print_verbose_only=True, runInPretendMode=True).stdout.strip().decode("utf-8")
+        if (head_rev == upstream_rev):
+            print(coloured(AnsiColour.blue, "HEAD is up to date with", upstream_branch, "at", upstream_rev[:11]))
+            return
+
         # make sure we run git stash if we discover any local changes
         has_changes = len(runCmd("git", "diff", "--stat", "--ignore-submodules",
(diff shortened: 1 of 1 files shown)


arichardson:
This appears to be failing in Jenkins because it's running the commands when the source dir doesn't exist.
There should probably an early return after `self.ensure_cloned()`
```
if not src_dir.exists():
    return
```

brooksdavis:
@arichardson I don't really understand how to diagnose these failures which seem to be due to some aspect of the jenkins git checkout.

pycheribuild/projects/project.py
```diff
@@ -1293,6 +1293,26 @@ def update(self, current_project: "Project", *, src_dir: Path, default_src_dir:
...
+
+        upstream_remote = upstream_branch.split("/", 1)[0]
+        print(coloured(AnsiColour.red, "upstream_remote " + upstream_remote))
+        runCmd("git", "fetch", upstream_remote)
+        head_rev = runCmd("git", "rev-parse", "HEAD",
+                          captureOutput=True, cwd=src_dir, print_verbose_only=True, runInPretendMode=True).stdout.strip().decode("utf-8")
+        upstream_rev = runCmd("git", "rev-parse", upstream_branch,
+                              captureOutput=True, cwd=src_dir, print_verbose_only=True, runInPretendMode=True).stdout.strip().decode("utf-8")
+        if (head_rev == upstream_rev):
+          print(coloured(AnsiColour.blue, "HEAD is up to date with", upstream_branch, "at", upstream_rev[:11]))
```
arichardson, line 1313: indentation seems wrong.

pycheribuild/projects/project.py
```diff
@@ -1293,6 +1293,26 @@ def update(self, current_project: "Project", *, src_dir: Path, default_src_dir:
                     if current_project.query_yes_no("Update to correct URL?"):
                         runCmd("git", "remote", "set-url", "origin", self.url, runInPretendMode=True, cwd=src_dir)
 
+	# see if we have an upstream
```
arichardson, line 1296:
```suggestion
        # see if we have an upstream
```

pycheribuild/projects/project.py
brooksdavis: noticed and fixed in my force push

pycheribuild/projects/project.py
```diff
@@ -1293,6 +1295,27 @@ def update(self, current_project: "Project", *, src_dir: Path, default_src_dir:
...
+        # see if we have an upstream
+        head = runCmd("git", "symbolic-ref", "-q", "HEAD",
+                      captureOutput=True, cwd=src_dir, print_verbose_only=True, runInPretendMode=True).stdout.strip().decode("utf-8")
+        upstream_branch = runCmd("git", "for-each-ref", '--format=%(upstream:short)', head, "origin/mainline",
+                      captureOutput=True, cwd=src_dir, print_verbose_only=True, runInPretendMode=True).stdout.strip().decode("utf-8")
+        if len(upstream_branch) == 0:
+            print(coloured(AnsiColour.blue, "No upstream to update from"))
+            return
+
+	# make sure there's work to do before touching the repo
```
arichardson, line 1307:
```suggestion
        # make sure there's work to do before touching the repo
```

pycheribuild/projects/project.py
```diff
@@ -1293,6 +1295,27 @@ def update(self, current_project: "Project", *, src_dir: Path, default_src_dir:
                     if current_project.query_yes_no("Update to correct URL?"):
                         runCmd("git", "remote", "set-url", "origin", self.url, runInPretendMode=True, cwd=src_dir)
 
+        # see if we have an upstream
+        head = runCmd("git", "symbolic-ref", "-q", "HEAD",
+                      captureOutput=True, cwd=src_dir, print_verbose_only=True, runInPretendMode=True).stdout.strip().decode("utf-8")
+        upstream_branch = runCmd("git", "for-each-ref", '--format=%(upstream:short)', head, "origin/mainline",
+                      captureOutput=True, cwd=src_dir, print_verbose_only=True, runInPretendMode=True).stdout.strip().decode("utf-8")
```
brooksdavis, line 1302: A bit more digging suggested this might be a better command `git rev-parse --abbrev-ref --symbolic-full-name @{upstream}`.  I'm not sure when either set of commands became available.

arichardson: Right now I think the oldest OS that we try to support is Ubuntu 16.04. I just managed to run both of these on a Ubuntu 16.04 docker container, so I think the `git rev-parse --abbrev-ref --symbolic-full-name @{upstream}` command looks good.

brooksdavis: bah, upon further checking, that command fails with an error if no upstream is present and thus doesn't work.  I've reverted the previous change so we're back to the original.

arichardson: Maybe the following works? `rev-parse` returns 128 for me if there is no upstream, but I'm not sure if that is always the case, so I added the output check too:
```python
try:
  upstream_branch = runCmd("git", "rev-parse", "--abbrev-ref", "--symbolic-full-name", "@{upstream}",
       captureError=True, captureOutput=True, cwd=src_dir, print_verbose_only=True,
        runInPretendMode=True).stdout.strip().decode("utf-8")
except subprocess.CalledProcessError as e:
  if e.returncode == 128 or "no upstream configured" in e.stderr:
    print(coloured(AnsiColour.blue, "No upstream to update from"))
    return
  raise
```

PR #78, opened 2020-04-21
Skip pull --rebase if the current branch is ahead of upstream

arichardson:
This should avoid unnecessary rebuilds since `pull --rebase` might change
modification times of file.

I'm not sure if we need the `git rev-parse --abbrev-ref --symbolic-full-name @{upstream}` or whether `git fetch` will always do the right thing.

### DIFF
--- a/pycheribuild/projects/project.py
+++ b/pycheribuild/projects/project.py
@@ -1295,23 +1295,11 @@ class GitRepository(SourceRepository):
                     if current_project.query_yes_no("Update to correct URL?"):
                         runCmd("git", "remote", "set-url", "origin", self.url, runInPretendMode=True, cwd=src_dir)
 
-        # see if we have an upstream and fetch it
-        try:
-            upstream_branch = runCmd("git", "rev-parse", "--abbrev-ref", "--symbolic-full-name", "@{upstream}",
-                captureError=True, captureOutput=True, cwd=src_dir, print_verbose_only=True,
-                runInPretendMode=True, raiseInPretendMode=True).stdout.strip().decode("utf-8")
-            # make sure there's work to do before touching the repo
-            upstream_remote = upstream_branch.split("/", 1)[0]
-            current_project.verbose_print("Upstream remote is " + upstream_remote)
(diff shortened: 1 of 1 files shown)


jrtc27:
Really? Rebase is smart enough to not create new commit objects when it's a fast-forward operation, I would expect it to update the working copy lazily in that case... :(

arichardson:
> Really? Rebase is smart enough to not create new commit objects when it's a fast-forward operation, I would expect it to update the working copy lazily in that case... :(

I wonder if this is caused by my change to use `--rebase=merges`. I made that change because I got annoyed at pull --rebase flattening my upstream merges for QEMU...

jrtc27:
`git pull --rebase=merges` (i.e. `git rebase --rebase-merges`) isn't in Ubuntu 16.04, nor even 18.04. But yes, if you have merge commits, that will end up causing the merges to be replayed and I believe they will get new commit objects, though if you have no merges. Indeed it does also seem that it touches the timestamps; a normal non-interactive rebase will detect that there's nothing to do and not even run the operations, but since `--rebase[-=]merges` uses the interactive machinery behind the scenes it doesn't have that early bail-out and despite each `pick` being a fast-forward the timestamps are in fact updated rather than being lazier... _sigh_

arichardson:
I guess I could check if the local changes include a merge commit and only add the --rebase=merges flag if they do but that also seems quite fragile.

arichardson:
Regarding the --rebase=merges: I was going to use --rebase=preserve (and cheribuild does it for git < 2.18) but the latest documentation says this is deprecated in favour of --rebase=merges. However, I think --rebase=preserve also results in the same problems.

jrtc27:
> Regarding the --rebase=merges: I was going to use --rebase=preserve (and cheribuild does it for git < 2.18) but the latest documentation says this is deprecated in favour of --rebase=merges. However, I think --rebase=preserve also results in the same problems.

Yeah, that's deprecated; they all suck, but `--rebase=merges` sucks least, even if it does have some pretty stupid bugs.

PR #79, opened 2020-04-21
Initial attempt at CheriOS target info

arichardson:


nwf:
It certainly gets me further along, but I had to patch `cherios-llvm` to also include `host` target support, because otherwise `cmake`'s feature tests fail, because `cmake` does not apply `CMAKE_C_FLAGS_INIT` before running its feature tests, for what I'm sure are reasons.  I tried fiddling with the `cheribuild`-generated toolchain file and `CMAKE_USER_MAKE_RULES_OVERRIDE` and other similar terrors but was not able to convince `cmake` to not want a host-targeting compiler for feature extraction. :(

PR #80, opened 2020-04-29
Open GDB tmux pane automatically when debugging

nicomazz:
When the flag `--debug-kernel` is added, another pane is opened on the side, so we skip the command copypaste.
Then, I added an option to randomize the gdb port (since the default one was always used)


pycheribuild/projects/run_qemu.py
```diff
@@ -256,9 +259,17 @@ def gdb_command(main_binary, breakpoint=None, extra_binary=None) -> str:
                 self.info("If you would like to debug a userspace program (e.g. sbin/init):")
                 self.info("\t", coloured(AnsiColour.red, gdb_command(self.rootfs_path / "sbin/init", "main", path_to_kernel)), sep="")
             self.info("Launching QEMU in suspended state...")
+
+            if self.config.debug_kernel:
+                server = libtmux.Server()
+                session = server.list_sessions()[0] # hopefully
```
jrtc27, line 265: What if I have multiple sessions? What if there are no sessions? Both should be handled without doing the wrong thing or crashing with an exception.

pycheribuild/projects/run_qemu.py
```diff
@@ -28,6 +28,9 @@
 # SUCH DAMAGE.
 #
 import socket
+from random import randint
+
+import libtmux
```
arichardson, line 33: I try hard to make cheribuild work with only the python standard library installed (that's why I bundle dependencies in the 3rdparty directory). This should be
```python
try:
     import libtmux
 except ImportError:
    .... fall back to something else
```

Also it shouldn't be necessary unless you have `self.config.debug_kernel` set. If it's only used within one function, I'd import it there to reduce startup time (important for bash tab-completion speed).

pycheribuild/projects/run_qemu.py
```diff
@@ -222,18 +225,18 @@ def add_smb_dir(directory, target, share_name=None, readonly=False):
...
-
             def gdb_command(main_binary, breakpoint=None, extra_binary=None) -> str:
                 gdb_cmd = BuildGDB.getInstallDir(self, cross_target=CompilationTargets.NATIVE) / "bin/gdb"
                 # Set the sysroot to ensure that the .debug file is loaded from $SYSROOT/usr/lib/debug/boot/kernel
                 result = [gdb_cmd, main_binary, "--init-eval-command=set sysroot " + str(self.rootfs_path)]
                 # Once the file has been loaded set a breakpoint on panic() and connect to the remote host
                 if breakpoint:
                     result.append("--eval-command=break " + breakpoint)
-                result.append("--eval-command=target remote localhost:1234")
+                result.append(f"--eval-command=target remote localhost:{gdb_port}")
```
arichardson, line 239: Unfortunately cheribuild still needs to support python 3.5 so we can't use f-strings.

jrtc27: With Ubuntu 20.04 released a week ago, we could start to retire 16.04 support and mandate at least 18.04. 16.04 only has another year left until you have to pay for extended maintenance (except for personal use), and I very much hope that we can kill off the remaining 16.04 Jenkins slaves imminently.

arichardson: Even after killing the jenkins slaves, we might still have external users of cheribuild that are using an older (but still supported Ubuntu version). Currently increasing the minimum to 3.6 really doesn't add much other than f-strings, so I don't see an urgent need to move to 3.6 minimum.

For other projects increasing the minimum supported release might be more useful.
I don't think we need to support 16.04 for every target supported by cheribuild.

pycheribuild/projects/run_qemu.py
```diff
@@ -222,18 +223,18 @@ def add_smb_dir(directory, target, share_name=None, readonly=False):
         if self._hasPCI:
             qemuCommand += ["-device", "virtio-rng-pci"]
 
-        if self.config.wait_for_debugger:
-            self.info("QEMU is waiting for GDB to attach (using `target remote :1234`)."
-                      " Once connected enter 'continue\\n' to continue booting")
-
+        if self.config.wait_for_debugger or self.config.debug_kernel:
+            gdb_port = randint(1200,1300) if self.config.gdb_random_port else 1234
```
arichardson, line 227: I think it's probably fine to always use a random port, so I'd add default=True to the gdb_random_port option.

For reliability (minimizing race conditions) you should use  `gdb_socket_placeholder = find_free_port()` which opens a socket with a random port, and assign `gdb_port = gdb_socket_placeholder.port`.
To make the port available to QEMU run `gdb_socket_placeholder.socket.close()` just before launching QEMU.

nicomazz: I'm not sure I understood, this. can you fix it please?
Also, would be nice to automatically close the opened pane when qemu stops.. where is it better to close it, codewise?

nicomazz: Fixed  (:

pycheribuild/projects/run_qemu.py
```diff
@@ -259,9 +260,30 @@ def gdb_command(main_binary, breakpoint=None, extra_binary=None) -> str:
...
+                server = libtmux.Server()
+                if server is None:
+                    raise Exception("Tmux server not found")
+                sessions = server.list_sessions()
+                if len(sessions) != 1:
+                    raise Exception("There should be only one tmux session running")
+                session = server.list_sessions()[0]
+                window = session.attached_window
+                pane = window.split_window(vertical=True, attach=False)
+                pane.send_keys(command)
```
arichardson, line 274: I think it would be nice if we could just check the current tmux window:
It appears you can use tmux display-message to get the current session/window/pane ID. Not sure if libtmux has a wrapper for this.
```
$ tmux display-message -p '#{pane_id} #{pane_index} #{session_id} #{session_name} #{window_id} #{window_name}'
%5 1 $0 0 @3
```
It would be really nice if we could just split the current pane rather than the first one.

pycheribuild/config/chericonfig.py
```diff
@@ -222,6 +222,20 @@ def __init__(self, loader: ConfigLoaderBase, action_class):
         self.wait_for_debugger = loader.add_bool_option("wait-for-debugger", group=loader.run_group,
                                                         help="Start QEMU in the 'wait for a debugger' state when"
                                                              "launching CheriBSD,FreeBSD, etc.")
+
+        self.debug_kernel = loader.add_bool_option("debug-kernel",
```
arichardson, line 226: I'd prefer a different name for this option. Maybe `debugger-in-new-tmux-pane` but there might be a shorter and better name.

pycheribuild/projects/run_qemu.py
nicomazz: I think that now the current pane is split

arichardson:
I just tested this and it almost works as I would expect. Thanks!
If I run cheribuild inside tmux with this flag it will open a new pane with the debugger in the same window.
However if I run it outside of tmux (but there is a tmux session on this system) it will still open a new pane in that tmux session. 
I find that slightly surprising and would prefer an error message.

pycheribuild/projects/run_qemu.py
```diff
@@ -259,9 +260,30 @@ def gdb_command(main_binary, breakpoint=None, extra_binary=None) -> str:
...
+            def start_gdb_in_tmux_pane(command):
+                server = libtmux.Server()
+                if server is None:
+                    raise Exception("Tmux server not found")
+                sessions = server.list_sessions()
+                if len(sessions) != 1:
+                    raise Exception("There should be only one tmux session running")
+                session = server.list_sessions()[0]
+                window = session.attached_window
+                pane = window.attached_pane.split_window(vertical=True, attach=False)
```
arichardson, line 273: I would prefer a horizontal split, but I don't really mind.

arichardson:
But that should not be a blocker for this change since "--debugger-in-tmux-pane" is not on by default.

pycheribuild/projects/run_qemu.py
nicomazz: Now it is set based on the window size

nicomazz:
Now there is an error message if it isn't run from within a tmux session, and the direction is set based on the window size (:

arichardson:
Nice! Although the window direction doesn't seem to work for me.
I'll do some debugging.

arichardson:
Split direction logic fixed in https://github.com/CTSRD-CHERI/cheribuild/commit/0b219756408234eb2f08437725de5c15209bc4d9

PR #81, opened 2020-04-29
Add bbl dependency for run-riscv-minimal

arichardson:
I hope GitHub webui didn't mess up whitespace. Let's see what Jenkins says


arichardson:
Fixed in https://github.com/CTSRD-CHERI/cheribuild/commit/20a0498095037859c4e283b7851514ca6975fa32

PR #87, opened 2020-05-11
FETT OpenSSH and nginx linkage

brooksdavis:
Replacing #85

### DIFF
--- a/pycheribuild/config/target_info.py
+++ b/pycheribuild/config/target_info.py
@@ -562,11 +562,6 @@ class CheriBSDTargetInfo(FreeBSDTargetInfo):
     def get_rootfs_target(self) -> "Project":
         from ..projects.cross.cheribsd import BuildCHERIBSD
         xtarget = CompilationTargets.NONE
-        # Install the purecap targets into the hybrid rootfs:
-        if self.target.is_cheri_purecap([CPUArchitecture.MIPS64]):
-            xtarget = CompilationTargets.CHERIBSD_MIPS_HYBRID
-        elif self.target.is_cheri_purecap([CPUArchitecture.RISCV64]):
-            xtarget = CompilationTargets.CHERIBSD_RISCV_HYBRID
         return BuildCHERIBSD.get_instance(self.project, cross_target=xtarget)
 
 
--- a/pycheribuild/config/target_info.py
+++ b/pycheribuild/config/target_info.py
@@ -497,7 +497,7 @@ class CheriBSDTargetInfo(FreeBSDTargetInfo):
                 purecap_prefix="-purecap", hybrid_prefix="", nocheri_name="-mips")
         elif self.target.is_riscv(include_purecap=True):
             return self._sysroot_path(config.cheri_sdk_dir, separate_cheri_sysroots,
-                purecap_prefix="-riscv64c", hybrid_prefix="-riscv64c-hybrid", nocheri_name="-riscv64")
+                purecap_prefix="-riscv64-purecap", hybrid_prefix="-riscv64-hybrid", nocheri_name="-riscv64")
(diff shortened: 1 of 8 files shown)


pycheribuild/projects/cross/openssh.py
```diff
@@ -0,0 +1,77 @@
...
+    target = "fett-openssh"
+    project_name = "fett-openssh"
+    repository = GitRepository("https://github.com/CTSRD-CHERI/openssh-portable.git",
+                               default_branch="fett")
+
+    dependencies = ["fett-zlib", "fett-openssl"]
+
+    def configure(self, **kwargs):
+        # XXX: this is wrong, we should get the paths from our dependencies
+        # rather than assuming we have the same basic configs.
```
brooksdavis, line 67: I'd really like to be able to query the dependent objects, but I wasn't clear to me how to get ahold of them.

arichardson: You can do things like 

`openssl_dir = BuildFettOpenSSL.getInstallDir(self)` (not updated yet to use python style names).
If you need other properties `BuildFettOpenSSL.get_instance(self).foo`

pycheribuild/config/target_info.py
```diff
@@ -562,11 +562,6 @@ def pkgconfig_dirs(self) -> str:
     def get_rootfs_target(self) -> "Project":
         from ..projects.cross.cheribsd import BuildCHERIBSD
         xtarget = CompilationTargets.NONE
-        # Install the purecap targets into the hybrid rootfs:
-        if self.target.is_cheri_purecap([CPUArchitecture.MIPS64]):
-            xtarget = CompilationTargets.CHERIBSD_MIPS_HYBRID
-        elif self.target.is_cheri_purecap([CPUArchitecture.RISCV64]):
-            xtarget = CompilationTargets.CHERIBSD_RISCV_HYBRID
         return BuildCHERIBSD.get_instance(self.project, cross_target=xtarget)
```
brooksdavis, line 565: Without this the programs don't end up in the pure cap disk image.

arichardson: Since we are moving away from using hybrid in favour of rootfs, I think this is fine.
It was required back when the full purecap userspace wasn't stable: I used a hybrid world to test new purecap projects.
I tend to mount directories via smbfs rather than embedding in the rootfs, so installing both hybrid and purecap to the same rootfs shouldn't be required anymore.

pycheribuild/projects/cross/nginx.py
```diff
@@ -113,3 +119,10 @@ class BuildFettNginx(BuildNginx):
     project_name = "fett-nginx"
     repository = GitRepository("https://github.com/CTSRD-CHERI/nginx.git",
         default_branch="fett")
+    dependencies = ["fett-openssl"]
+
+    def configure(self):
+        openssl_dir = str(BuildFettOpenSSL.get_instance(self)._installPrefix)
+        self.configureEnvironment["NGX_OPENSSL_fett_path"] = str(BuildFettOpenSSL.get_instance(self).destdir) + openssl_dir
```
arichardson, line 126: Does `self.configureEnvironment["NGX_OPENSSL_fett_path"] = str(BuildFettOpenSSL.getInstallDir(self))` work?

arichardson: Ah never mind you need the install prefix for the rpath anyway.

pycheribuild/config/target_info.py
brooksdavis: Thinking this over (at 4am, thanks brain) I think we should probably make all the -purecap cross targets default to using the purecap sdk and rootfs, but perhaps should retain a fallback to using the hybrid version.

PR #88, opened 2020-05-11
compiler-rt: Use self.sed_bindir to set LLVM_CONFIG_PATH

heshamelmatary:
Otherwise it fails to find it in the build dir with this error:

llvm-config failed with status No such file or directory

### DIFF
--- a/pycheribuild/projects/cross/compiler-rt.py
+++ b/pycheribuild/projects/cross/compiler-rt.py
@@ -32,6 +32,7 @@ from .crosscompileproject import *
 from ..llvm import BuildCheriLLVM
 from ..project import ReuseOtherProjectDefaultTargetRepository
 from ...utils import classproperty
+from ...utils import setEnv, commandline_to_str, is_jenkins_build
 
 
 class BuildCompilerRt(CrossCompileCMakeProject):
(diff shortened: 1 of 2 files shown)


pycheribuild/projects/cross/compiler-rt.py
```diff
@@ -47,14 +47,14 @@ class BuildCompilerRt(CrossCompileCMakeProject):
...
+        if self.target_info.is_rtems() or self.target_info.is_baremetal():
             self.add_cmake_options(CMAKE_TRY_COMPILE_TARGET_TYPE="STATIC_LIBRARY") # RTEMS only needs static libs
             # Get default target (arch) from the triple
             self.add_cmake_options(COMPILER_RT_DEFAULT_TARGET_ARCH=self.target_info.target_triple.split('-')[0])
 
         self.add_cmake_options(
-            LLVM_CONFIG_PATH=BuildCheriLLVM.getBuildDir(self, cross_target=CompilationTargets.NATIVE) / "bin/llvm-config",
-            LLVM_EXTERNAL_LIT=BuildCheriLLVM.getBuildDir(self, cross_target=CompilationTargets.NATIVE) / "bin/llvm-lit",
+            LLVM_CONFIG_PATH=self.sdk_bindir / "llvm-config",
+            LLVM_EXTERNAL_LIT=self.sdk_bindir/ "llvm-lit",
```
arichardson, line 57: This is probably better, but the libc++ jenkins job will need adjusting.

heshamelmatary: Is that why the CI check if failing?

arichardson: The failing CI reminds me why I made this change. For the native build there is no sdk_bindir, so it should use it from the build dir. Maybe the best solution is to change it to "build dir by default", but "sdk dir" for Jenkins.

pycheribuild/projects/cross/compiler-rt.py
```diff
@@ -47,14 +47,14 @@ class BuildCompilerRt(CrossCompileCMakeProject):
...
-        if self.target_info.is_rtems():
+        if self.target_info.is_rtems() or self.target_info.is_baremetal():
             self.add_cmake_options(CMAKE_TRY_COMPILE_TARGET_TYPE="STATIC_LIBRARY") # RTEMS only needs static libs
             # Get default target (arch) from the triple
             self.add_cmake_options(COMPILER_RT_DEFAULT_TARGET_ARCH=self.target_info.target_triple.split('-')[0])
 
         self.add_cmake_options(
-            LLVM_CONFIG_PATH=BuildCheriLLVM.getBuildDir(self, cross_target=CompilationTargets.NATIVE) / "bin/llvm-config",
-            LLVM_EXTERNAL_LIT=BuildCheriLLVM.getBuildDir(self, cross_target=CompilationTargets.NATIVE) / "bin/llvm-lit",
+            LLVM_CONFIG_PATH=self.sdk_bindir / "llvm-config",
```
arichardson, line 56:
```suggestion
            LLVM_CONFIG_PATH=self.sdk_bindir / "llvm-config" if is_jenkins_build() else BuildCheriLLVM.getBuildDir(self, cross_target=CompilationTargets.NATIVE) / "bin/llvm-config",
```

arichardson:
I guess the condition need to be `if is_jenkins_build() and not self.compiling_for_host()`

PR #94, opened 2020-05-16
Add options to use the local FPGA build environment.

qwattash:
This makes it easier to setup a dev environment outside the CL, with a local quartus install.
- Add option to use a custom environment setup script instead of the cheri-cpu ones. This is useful if   you are not running within the CL NFS or have different NFS mount-points.
- Always build the qemu SLIRP from git as it may complain when looking for the system one.

### DIFF
--- a/pycheribuild/config/chericonfig.py
+++ b/pycheribuild/config/chericonfig.py
@@ -289,6 +289,9 @@ class CheriConfig(object):
             help="Perform a shallow `git clone` when cloning new projects. This can save a lot of time for large"
             "repositories such as FreeBSD or LLVM. Use `git fetch --unshallow` to convert to a non-shallow clone")
 
+        self.fpga_custom_env_setup_script = loader.add_path_option("beri-fpga-env-setup-script",
+            help="Custom script to source to setup PATH and quartus, default to using cheri-cpu/cheri/setup.sh")
+
         self.targets = None  # type: typing.Optional[typing.List[str]]
         self.FS = None  # type: Optional[FileSystemUtils]
         self.__optionalProperties = ["preferred_xtarget"]
--- a/pycheribuild/projects/cherisim.py
+++ b/pycheribuild/projects/cherisim.py
@@ -98,11 +98,14 @@ class BuildCheriSim(Project):
         return None
 
     def compile(self, **kwargs):
-        if not (self.sourceDir / "cheri/setup.sh").exists():
-            self.fatal("Could not find setup.sh, please set --cheri-sim/source-directory")
-        self.runShellScript("source setup.sh && " + commandline_to_str(self.get_make_commandline("sim", parallel=False)),
+        setup_sh = self.sourceDir / "cheri" / "setup.sh"
+        if self.config.fpga_custom_env_setup_script:
+            setup_sh = self.config.fpga_custom_env_setup_script
+        if not setup_sh.exists():
+            self.fatal("Could not find setup.sh, please set --cheri-sim/source-directory or --fpga-env-setup-script")
+        source_cmd = "source {setup_script}".format(setup_script=setup_sh)
+        self.runShellScript(source_cmd + " && " + commandline_to_str(self.get_make_commandline("sim", parallel=False)),
                             cwd=self.sourceDir / "cheri", shell="bash")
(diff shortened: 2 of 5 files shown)


pycheribuild/projects/cherisim.py
```diff
@@ -127,12 +127,21 @@ def __init__(self, config: CheriConfig):
         super().__init__(config)
         self.make_args.set(JTAG_ATLANTIC=1) # MUCH faster
 
+    @classmethod
+    def setup_config_options(cls, **kwargs):
```
arichardson, line 131: Could make this a global config option so you don't have to add it to `run-fpga` as well?

qwattash: Would that be visible to other projects as well?

arichardson: You could also do BuildBeriCtl.get_instance().env_script if you don't want to pollute the global options namespace. But since there are three projects that would need it, I think it's fine to make it a global option.

arichardson: BuildCheriSim should also use this setting.

qwattash: Right, I prefer the global option to reading other project's configs. I will also switch BuildCheriSim to use it.

pycheribuild/config/chericonfig.py
```diff
@@ -289,6 +289,9 @@ def __init__(self, loader: ConfigLoaderBase, action_class):
             help="Perform a shallow `git clone` when cloning new projects. This can save a lot of time for large"
             "repositories such as FreeBSD or LLVM. Use `git fetch --unshallow` to convert to a non-shallow clone")
 
+        self.fpga_custom_env_setup_script = loader.add_path_option("fpga-env-setup-script",
+            default="", help="Custom script to source to setup PATH and quartus, default to using cheri-cpu/cheri/setup.sh")
```
arichardson, line 293:
```suggestion
            help="Custom script to source to setup PATH and quartus, default to using cheri-cpu/cheri/setup.sh")
```

The default should already be `None`

pycheribuild/projects/cherisim.py
```diff
@@ -98,11 +98,13 @@ def clean(self):
...
-        self.runShellScript("source setup.sh && " + commandline_to_str(self.get_make_commandline("sim", parallel=False)),
-                            cwd=self.sourceDir / "cheri", shell="bash")
-        pass
+        setup_sh = self.sourceDir / "cheri" / "setup.sh"
+        if self.config.fpga_custom_env_setup_script:
+            setup_sh = self.config.fpga_custom_env_setup_script
+        if not setup_sh.exists():
+            self.fatal("Could not find setup.sh, please set --cheri-sim/source-directory or --fpga-env-setup-script")
+        self.runShellScript("source {setup_script} && " + commandline_to_str(self.get_make_commandline("sim", parallel=False)),
+                            setup_script=setup_sh, cwd=self.sourceDir / "cheri", shell="bash")
```
arichardson, line 107:
```suggestion
        self.runShellScript("source {setup_script} && {build_cmd}".format(setup_script=setup_sh, build_cmd=commandline_to_str(self.get_make_commandline("sim", parallel=False))),
                            cwd=self.sourceDir / "cheri", shell="bash")
```

pycheribuild/config/chericonfig.py
```diff
@@ -289,6 +289,9 @@ def __init__(self, loader: ConfigLoaderBase, action_class):
             help="Perform a shallow `git clone` when cloning new projects. This can save a lot of time for large"
             "repositories such as FreeBSD or LLVM. Use `git fetch --unshallow` to convert to a non-shallow clone")
 
+        self.fpga_custom_env_setup_script = loader.add_path_option("fpga-env-setup-script",
```
arichardson, line 292:
```suggestion
        self.fpga_custom_env_setup_script = loader.add_path_option("cheri-mips-fpga-env-setup-script",
```
In case we need something similar for riscv

arichardson:
LGTM, but we should probably use a name to indicate that it's only for bluespec CHERI-MIPS

jrtc27:
beri?

PR #95, opened 2020-05-20
Abort on missing explicitly-given config file

### LINKED ISSUE #76: Should error out if explicitly specified --config-file does not exist
Right now, it just raises a warning and continues on, which is the right idea if it hasn't been explicitly set, but is unexpected if it has.

nwf:
Fixes #76, but unsure if a stylistically acceptable approach.

### DIFF
--- a/pycheribuild/config/loader.py
+++ b/pycheribuild/config/loader.py
@@ -506,6 +506,12 @@ def dict_raise_on_duplicates(ordered_pairs):
             d[k] = v
     return d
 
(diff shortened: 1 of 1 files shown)


nwf:
Hmmm!  It isn't.  Let's try that again.

nwf:
Ah, I see.  Let's take a different tack.

arichardson:
This seems useful for other cases too. I only load the JSON config values of the argparse value is not equal to the default. But that breaks if you explicitly pass the default.

PR #97, opened 2020-05-22
point CMAKE_PREFIX_PATH to Find[PROJECT].cmake files

dodsonmg:
Function to populate `CMAKE_PREFIX_PATH` in `CrossToolchain.cmake` should include the subdirectory `libcheri/cmake` for `find_packages()` to find the right `Find[Project].cmake` file.


arichardson:
This should probably be in CheriBSD target info and only for purecap. But since we will stop installing to libcheri longer term, adding it here is fine for now.

PR #103, opened 2020-05-25
create a setup file to set cheri env vars after install

dodsonmg:
Building ROS creates a `setup.bash` file that should be `source`ed before running any ROS applications.  The `setup.bash` file sets environment variables (primarily `LD_LIBRARY_PATH`).

This PR creates an analogous file (`cheri_setup.csh`) after building ROS.  This file is created in `self.sourceDir` and should be `source`ed to set CHERI environment variables in CheriBSD (primarily `LD_CHERI_LIBRARY_PATH`) before running a ROS application.

### DIFF
--- a/pycheribuild/projects/cross/ros2.py
+++ b/pycheribuild/projects/cross/ros2.py
@@ -25,8 +25,7 @@
 #
 
 from .crosscompileproject import *
-# from ..project import ReuseOtherProjectRepository
-
+import shlex
 
 class BuildRos2(CrossCompileCMakeProject):
     project_name = "ros2"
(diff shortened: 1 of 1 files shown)


jrtc27:
ROS should learn about rpath...

dodsonmg:
Maybe, but it's a bit more complicated than that.

Normally, you build your main ROS libraries and tools in one workspace (which is what this cheribuild target does), and then you build your applications in a separate workspace.  Since there are several flavours of ROS that you might be building against, and since developers often have multiple flavours building on the same machine (e.g., for ROS1 and ROS2), they've designed the system such that you source the `setup.bash` file of the ROS flavour you want to build against before building your application workspace.

PR #105, opened 2020-05-27
Set WITH_CHERI in the CheriBSD build process.

brooksdavis:
The other didn't seem to stick and this seems to be more the right mechanism.


arichardson:
I'm surprised the other way didn't work but LGTM.

PR #106, opened 2020-05-28
Minor infrastructure improvements

brooksdavis:


brooksdavis:
meh, branched from the wrong place, will force push tomorrow.

brooksdavis:
I've fixed a dumb bug in "Abstract the prefixes to autoadd to the METALOG" so all the files get added.

PR #108, opened 2020-05-28
Use CHERI_* configs for CheriBSD on MIPS.

bsdjhb:
CheriBSD only supports 128-bit CHERI on MIPS and this is more inline
with RISC-V.

### DIFF
--- a/README.md
+++ b/README.md
@@ -501,8 +501,7 @@ Options for target 'cheribsd':
                         will break if not all dependencies have been built. (default: 'the value of the global
                         --freebsd-subdir options')
   --cheribsd/kernel-config CONFIG
-                        The kernel configuration to use for `make buildkernel` (default: CHERI_MALTA64 or
-                        CHERI128_MALTA64 depending on --cheri-bits) (default: 'target-dependent default')
+                        The kernel configuration to use for `make buildkernel` (default: 'target-dependent default')
   --cheribsd/no-debug-info
                         Do not pass make flags for building with debug info
   --cheribsd/no-auto-obj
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -501,8 +501,8 @@ Options for target 'cheribsd':
                         will break if not all dependencies have been built. (default: 'the value of the global
                         --freebsd-subdir options')
   --cheribsd/kernel-config CONFIG
-                        The kernel configuration to use for `make buildkernel` (default: CHERI_MALTA64 or
-                        CHERI128_MALTA64 depending on --cheri-bits) (default: 'target-dependent default')
+                        The kernel configuration to use for `make buildkernel` (default: CHERI_MALTA64)
```
bsdjhb, line 504: Should the MIPS-specific default clause just be removed entirely?

bsdjhb:
There are many other references to 128 around and this doesn't try to address those (e.g. the names of disk images, etc. for MIPS).

README.md
arichardson: Yes, I think that is probably best. It made sense a long time ago but no longer does.

arichardson:
> There are many other references to 128 around and this doesn't try to address those (e.g. the names of disk images, etc. for MIPS).

Yeah I kept those names around so that existing files can still be used without having to do a full rebuild. But if it's annoying I can remove them.

bsdjhb:
Mmm, the NO_SWAPPING fix for CHERI_MALTA64 at least does need to be pulled up into master first, yes.  I'm fine with delaying this until the next dev -> master merge.  Are we planning on doing one next week?  (I think our goal was to do dev -> master roughly weekly)

README.md
bsdjhb: Trimmed.

bsdjhb:
I think this is ok to merge after the recent merge to master in CheriBSD?

PR #111, opened 2020-06-03
Allow forcing clean build using a .require_clean_build in the git repo

arichardson:
This file is parsed by cheribuild to determine whether a clean build is
required. It will read the last non-empty, non-comment line as an integer and
compare that to the value read from this file at the last time.
The previous logic using timestamps was fragile and did not consider git
branches. Reading a file that is committed to the repository fixes this
issue:

I will commit the following file to CheriBSD once this change has landed:

```
# Add a new line containing a larger number than the last (e.g. YYYYMMDD) for
# every change that breaks ABI or requires a full rebuild for some other reason.
# Only the last entry is required, but having the full list may be useful for
# looking at frequency of ABI breakage, etc.

# Building all programs hybrid
# https://github.com/CTSRD-CHERI/cheribsd/commit/cc876df741e840569cc3032f2536fe8abbe39030
20200526
```

### DIFF
--- a/pycheribuild/projects/project.py
+++ b/pycheribuild/projects/project.py
@@ -1470,7 +1470,6 @@ class Project(SimpleProject):
     skipGitSubmodules = False
     compileDBRequiresBear = True
     doNotAddToTargets = True
-    full_rebuild_if_older_than = None  # type: datetime.datetime
 
     build_dir_suffix = ""   # add a suffix to the build dir (e.g. for freebsd-with-bootstrap-clang)
     add_build_dir_suffix_for_native = False  # Whether to add -native to the native build dir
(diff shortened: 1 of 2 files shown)


pycheribuild/projects/project.py
```diff
@@ -2499,25 +2529,25 @@ def process(self):
...
+                clean_counter_in_build_dir = int(last_clean_counter_path.read_text().strip())
+                if clean_counter_in_build_dir < required_clean_counter:
+                    self.info("Forcing full rebuild since clean counter in build dir (", clean_counter_in_build_dir,
+                        ") is less than required minimum ", required_clean_counter, sep="")
+                    self._force_clean = True
+                else:
+                    self.verbose_print("Not forcing clean build since clean counter in build dir",
+                        clean_counter_in_build_dir, "is >= required minimum", required_clean_counter)
+            except Exception as e:
+                self.warning("Could not parse", last_clean_counter_path, "-> assuming clean build is required.", e)
```
jrtc27, line 2548: We probably don't want to print a warning if the file doesn't exist; that should be handled gracefully as if the counter is -INF.

arichardson: Yeah I just had that warning there to see that it did the right thing without a debugger.
 Will fix

arichardson:
> Sounds good. I'd like the convention for the counter line include a comment with the reason. This seems to be supported in a skim of the code:
> 
> ```
> YYYYMMDD	# reason
> ```

Yeah, whitespace and everything after  # is ignored.

PR #114, opened 2020-06-06
ROS2 testing

dodsonmg:
First stab at adding testing for the ROS2 target.

This doesn't currently work.  If I don't load sysroot, the linker can't find `librt.so.1`.  If I do load sysroot, it can't find any of my libraries, even when LD_CHERI_LIBRARY_PATH appears to be set correctly.

### DIFF
--- a/pycheribuild/projects/cross/ros2.py
+++ b/pycheribuild/projects/cross/ros2.py
@@ -96,9 +96,9 @@ class BuildRos2(CrossCompileCMakeProject):
             fout.write("#!/bin/csh\n\n")
             fout.write("setenv LD_CHERI_LIBRARY_PATH " + LD_CHERI_LIBRARY_PATH + "\n\n")
             fout.write("setenv LD_LIBRARY_PATH " + LD_CHERI_LIBRARY_PATH)
-        
+
         return
-    
+
     def update(self):
         super().update()
         if not (self.sourceDir / "src").is_dir():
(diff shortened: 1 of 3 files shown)


test-scripts/run_ros2_tests.py
```diff
@@ -0,0 +1,42 @@
...
+#
+import argparse
+
+from run_tests_common import *
+
+
+def run_ros2_tests(qemu: boot_cheribsd.CheriBSDInstance, args: argparse.Namespace) -> bool:
+    boot_cheribsd.info("Running ROS2 tests")
+    boot_cheribsd.set_ld_library_path_with_sysroot(qemu)
+    boot_cheribsd.checked_run_cheribsd_command(qemu, "sh -xe ./source/run-ros2-tests.sh", timeout=240 * 60)
```
arichardson, line 36:
```suggestion
    boot_cheribsd.checked_run_cheribsd_command(qemu, "cd /source && sh -xe ./run-ros2-tests.sh", timeout=240 * 60)
```

pycheribuild/projects/cross/ros2.py
```diff
@@ -63,10 +66,31 @@ def _run_colcon(self, **kwargs):
         if self.config.verbose:
             cmdline.append("--event-handlers")
             cmdline.append("console_cohesion+")
-        return self.run_cmd(cmdline, cwd=self.sourceDir, **kwargs)
+        self.run_cmd(cmdline, cwd=self.sourceDir, **kwargs)
+
+    def _find_file(self, name, path):
+        for root, dirs, files in os.walk(path):
```
arichardson, line 72:
```suggestion
        for root, dirs, files in os.walk(str(path)):
```
Seems to be needed for python 3.5

PR #115, opened 2020-06-08
Use UFS2 for all rootfs images and enable soft updates journaling

### LINKED ISSUE #112: Enable journaling for UFS images

jrtc27:
Closes #112

Please test this on Linux/macOS; newfs claims that it needs to use tunefs for softupdates, I assume makefs is the same, and I don't know whether it'll work out of the box currently when bootstrapping.


arichardson:
Works fine on macOS and Linux. Will this reduce the probability that disk images become completely corrupted?

jrtc27:
It should, though it’s still possible.

PR #117, opened 2020-06-10
run_qemu: look for "ss" on Linux when reporting port usage

nwf:
And report when we can't find either "ss" or "netstat" on Linux, and
report (rather than silently saying nothing) if we're on an unknown
platform.

@anaaktge reported some rather ugly spew from a system lacking `netstat`:

```
sh -c 'netstat -tulpne | grep ":10005"'
sh: 1: netstat: not found
Fatal error: Command `sh -c 'netstat -tulpne | grep ":10005"'` failed with non-zero exit code 1. Working directory was /home/a
```

As `ss` is the new hotness on Linux (part of the `iproute2` suite), hopefully this clears that up.


arichardson:
Thanks!

PR #122, opened 2020-06-15
Add support for automatic variable initialization

### LINKED ISSUE #121: mechanism to enable init-all=zero in cheribuild
For cross building it should add `-ftrivial-auto-var-init=zero -enable-trivial-auto-var-init-zero-knowing-it-will-be-removed-from-clang` to cflags. For CheriBSD WITH_INIT_ALL_ZERO should be defined during buildworld.

It would be helpful if it was possible to enable this from the command line and in a class (e.g. the FETT target applications or some future cheribsd-fett project).

arichardson:
Fixes #121

### DIFF
--- a/pycheribuild/projects/cross/cheribsd.py
+++ b/pycheribuild/projects/cross/cheribsd.py
@@ -1194,6 +1194,8 @@ class BuildCHERIBSD(BuildFreeBSD):
 
         # Support for automatic variable initialization:
         # See https://github.com/CTSRD-CHERI/cheribsd/commit/57e063b20ec04e543b8a4029871c63bf5cbe6897
+        # Explicitly disable first (in case the defaults in the source tree change)
+        self.make_args.set_with_options(INIT_ALL_ZERO=False, INIT_ALL_PATTERN=False)
         if self.auto_var_init is AutoVarInit.ZERO:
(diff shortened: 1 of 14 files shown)


pycheribuild/projects/cross/cheribsd.py
```diff
@@ -1192,6 +1192,13 @@ def __init__(self, config: CheriConfig):
             self.make_args.set(CROSS_BINUTILS_PREFIX=str(self.sdk_bindir / "llvm-"))
             self.use_llvm_binutils = True
 
+        # Support for automatic variable initialization:
+        # See https://github.com/CTSRD-CHERI/cheribsd/commit/57e063b20ec04e543b8a4029871c63bf5cbe6897
+        if self.auto_var_init is AutoVarInit.ZERO:
+            self.make_args.set_with_options(INIT_ALL_ZERO=True)
+        elif self.auto_var_init is AutoVarInit.PATTERN:
+            self.make_args.set_with_options(INIT_ALL_PATTERN=True)
```
brooksdavis, line 1200: Do we want to explicitly support None here (e.g. set WITHOUT_INIT_ALL_*=True) in case the default changes in some project or another.

arichardson: That makes sense, will add.

Issue #119, opened 2020-06-10
QEMU state not cleaned up if there are errors in cheribuild


anaaktge:
After running into issues with cheribuild, which appear to be fixed by pull 117 (thanks!), I ran into the following issue below. It appears the previous broken run left behind broken state. Resetting the world appears to fix it, so it's not blocking me, although @nwf requested I file a bug anyway.

Does cheribuild have a mechanism to clean up after itself if its faults or exits with a fatal error? If not, it might benefit from as much clean up as feasible if there is a fatal condition. 

For reference: 
Populating `/home/a/cheri/output/cheribsd-mips-purecap128.img'
Image `/home/a/cheri/output/cheribsd-mips-purecap128.img' complete
/home/a/cheri/output/sdk/bin/qemu-img info /home/a/cheri/output/cheribsd-mips-purecap128.img
image: /home/a/cheri/output/cheribsd-mips-purecap128.img
file format: raw
virtual size: 2.55 GiB (2738880512 bytes)
disk size: 1.33 GiB
Built target 'disk-image-mips-purecap' in 73.83739161491394 seconds
Port 10005 usage information:
sh -c 'ss -tulpne | grep ":10005"'
tcp    LISTEN  0       1                                         0.0.0.0:10005                                       0.0.0.0:*                                   users:(("qemu-system-che",pid=3799,fd=10)) uid:1000 ino:6904009 sk:22 <->      
Fatal error: SSH forwarding port 10005 is already in use! Make sure you don't  already have a QEMU instance running or change the chosen port by setting the config option run-mips-purecap/ssh-forwarding-port

PR #126, opened 2020-06-17
Restore TTY and file descriptor flags after running processes

arichardson:
QEMU crashing or not shutting down properly because it was manually killed
can result in the current terminal behaving very strangely. To avoid this
problem, save the tcgetattr result and F_FLAGS values for stdin/stdout/stderr
before executing other processes and restore them again after.

With this change I get the following output if I kill -9 QEMU after this patch:
```
# Warning: TTY flags for <stdin> changed, resetting them
Previous state [25862, 5, 191, 2619, 15, 15, [b'\x03', b'\x1c', b'\x7f', b'\x15', b'\x04', b'\x00', b'\x01', b'\x00', b'\x11', b'\x13', b'\x1a', b'\x00', b'\x12', b'\x0f', b'\x17', b'\x16', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00']]
New state [24580, 5, 191, 2608, 15, 15, [b'\x03', b'\x1c', b'\x7f', b'\x15', b'\x04', 0, 1, b'\x00', b'\x11', b'\x13', b'\x1a', b'\x00', b'\x12', b'\x0f', b'\x17', b'\x16', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00', b'\x00']]
```

Partially addresses https://github.com/CTSRD-CHERI/cheribuild/issues/119

### DIFF
--- a/pycheribuild/utils.py
+++ b/pycheribuild/utils.py
@@ -46,7 +46,7 @@ import typing
 from pathlib import Path
 from subprocess import CompletedProcess
 
-from .colour import coloured, AnsiColour
+from .colour import AnsiColour, coloured
 
 # reduce the number of import statements per project  # no-combine
 __all__ = ["typing", "printCommand", "includeLocalFile", "CompilerInfo",   # no-combine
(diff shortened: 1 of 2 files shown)


arichardson:
@jrtc27 Does this look sensible, or is there some other state I should also be saving?

pycheribuild/utils.py
```diff
@@ -156,11 +158,13 @@ def _make_called_process_error(retcode, args, *, stdout=None, stderr=None, cwd=N
 
 def check_call_handle_noexec(cmdline: "typing.List[str]", **kwargs):
     try:
-        return subprocess.check_call(cmdline, **kwargs)
+        with keep_terminal_sane():
```
arichardson, line 161: I guess I could also do this at the start of __main__ to avoid some (minor overhead), but this way it should not mess up output if something goes wrong in the middle of a cheribuild execution.

arichardson:
Ah, jenkins does not like this: `termios.error: (25, 'Inappropriate ioctl for device')`. I guess I need to check isatty first.

PR #128, opened 2020-06-17
CheriBSD: Disable GOOGLETEST by default

arichardson:
GOOGLETEST takes many minutes to compile and link with an
assertions-enabled clang (some of the tests take 30 seconds to compile and
about 6 minutes to link). Since the only user of GOOGLETEST in CheriBSD is
capsicum-test, disable it by default.

It's marked as BROKEN_OPTIONS on MIPS so we weren't building it at all until
https://github.com/CTSRD-CHERI/cheribsd/pull/544 enabled C++ support for RISC-V.


jrtc27:
Could you add this as another boolean option just so it's easy to override?

arichardson:
> Could you add this as another boolean option just so it's easy to override?

Yeah that probably makes sense. Will update the PR.

PR #132, opened 2020-06-20
Add a flag to build FETT kernels

brooksdavis:


brooksdavis:
Oops, I thought I'd already pushed the malloc.conf change, I guess it's batched with the other one now.

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -1230,6 +1233,14 @@ def __init__(self, config: CheriConfig):
...
             else:
                 self.fatal("Unsupported architecture for FPGA kernels")
+        if self.buildFettKernels:
+            if self.compiling_for_riscv(include_purecap=True):
+                if self.crosscompile_target.is_hybrid_or_purecap_cheri():
+                    self.extra_kernels.append("CHERI_FETT")
+                else:
+                    self.extra_kernels.append("FETT")
+            else:
+                self.fatal("Unsupported architecture for FETT kernels")
```
arichardson, line 1243: Might be better to make this a warning instead of fatal so you can put it in the global cheribsd section in the JSON config. But I don't really mind either way.

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -1179,6 +1179,9 @@ def setup_config_options(cls, install_directory_help=None, **kwargs):
         cls.buildFpgaKernels = cls.add_bool_option("build-fpga-kernels", show_help=True, _allow_unknown_targets=True,
                                                  only_add_for_targets=fpga_targets,
                                                  help="Also build kernels for the FPGA.")
+        cls.buildFettKernels = cls.add_bool_option("build-fett-kernels", show_help=True, _allow_unknown_targets=True,
```
arichardson, line 1182: I've been trying to gradually clean up the inconsistent naming to move to underscores since that's the python default. I'm happy to change this as part of the bulk rename commits so no urgent need to update this pr.

jrtc27: Alternatively it'd be nice if cheribsd-riscv64-purecap extended from cheribsd-riscv64 (and cheribsd-riscv64-nocheri would be where you put things specifically for _that_).... :D

pycheribuild/projects/cross/cheribsd.py
jrtc27: Alternatively it'd be nice if cheribsd-riscv64-purecap extended from cheribsd-riscv64 (and cheribsd-riscv64-nocheri would be where you put things specifically for _that_).... :D

arichardson: Yes that would be nice. I think it should be easy to do that once I finally fix way config options are inherited.

brooksdavis: Good call.  Fixed now.

PR #134, opened 2020-06-24
CheriOS qemu

arichardson:
The CheriOS changes have been merge to QEMU, so we no longer need a
separate target.

Not merging directly since I don't have a setup to test CheriOS stuff right now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,7 @@ usage: cheribuild.py [-h] [--config-file FILE] [--help-all] [--pretend] [--build
                      [--build-and-test] [--list-targets] [--print-chosen-targets] [--dump-configuration]
                      [--print-targets-only] [--clang-path CLANG_PATH] [--clang++-path CLANG++_PATH]
                      [--clang-cpp-path CLANG_CPP_PATH] [--pass-k-to-make] [--with-libstatcounters] [--skip-buildworld]
-                     [--freebsd-subdir SUBDIRS] [--install-subdir-to-sysroot] [--buildenv] [--libcompat-buildenv]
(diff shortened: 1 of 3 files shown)


nwf:
This seems fine, but it means that the qemu binary is now in `${outroot}/sdk/bin` not `${outroot}/cherios-sdk/bin`.

LawrenceEsswood:
Seems fine to me, when we get LLVM merged then cherios-sdk should die completely.

LawrenceEsswood:
I should note only my branch of CheriOS will work with this, which won't become master until LLVM is merged. Maybe hold off committing this until then? Otherwise we might annoy people on master of CheriOS.Sent from my Samsung device

-------- Original message --------

jrtc27:
@LawrenceEsswood What's the status of things? Does CheriOS master work with stock CHERI-QEMU these days?

LawrenceEsswood:
LLVM merge went stale =/ Need to resurrect that and merge it first.

jrtc27:
I don't understand why supporting QEMU is dependent on getting your LLVM merged into the main CHERI-LLVM. Can you not just fix your own LLVM to do whatever is needed for newer QEMU (not that I understand why that's even needed)?

LawrenceEsswood:
I think I was holding off pointing CheriOS to a new LLVM /(and updating it to my branch) until it could be pointed at CHERI-LLVM. I think its achievable if this branch needs killing off urgently, but it is not currently on the list.

LawrenceEsswood:
Actually, now I'm just massively confused on what the working state was. I really should go back and get CheriOS sorted.

jrtc27:
It's not urgent, I was just having a look through old issues and PRs for cheribuild trying to clean things up.

arichardson:
Superseded by #253.

PR #141, opened 2020-07-05
Properly inherit the purecap_kernel option for MFS root target.

qwattash:
Inherit the purecap_kernel option from the cheribsd kernel target instead
of requiring a new flag to be passed.


pycheribuild/projects/cross/cheribsd.py
```diff
@@ -1347,18 +1347,22 @@ def process(self):
...
+        self._build_and_install_kernel_binaries(
+            self.build_cheribsd_instance, kernconfs=kernel_configs, image=self.image)
 
     @property
     def fpga_kernconf(self):
         if self.compiling_for_mips(include_purecap=True):
             if self.crosscompile_target.is_hybrid_or_purecap_cheri():
                 prefix = "CHERI_" if self.config.mips_cheri_bits == 128 else "CHERI"
-                purecap = "PURECAP_" if self.purecap_kernel else ""
+                if self.build_cheribsd_instance.purecap_kernel:
```
arichardson, line 1362: Could keep the ternary expression here but I don't really mind

PR #146, opened 2020-07-21
Don't force a default cap table ABI.

bsdjhb:
The ABI is now only forced if --cap-table-abi is given explicitly.

While here, remove support for the legacy ABI.

### DIFF
--- a/pycheribuild/config/chericonfig.py
+++ b/pycheribuild/config/chericonfig.py
@@ -406,12 +406,6 @@ class CheriConfig(object):
         if "CLICOLOR" in os.environ:
             del os.environ["CLICOLOR"]
 
-    def _initialize_derived_paths(self):
-        # Set CHERI_BITS variable to allow e.g. { cheribsd": { "install-directory": "~/rootfs${CHERI_BITS}" } }
-        os.environ["CHERI_BITS"] = self.mips_cheri_bits_str
-        if self.cheri_cap_table_abi:
-            os.environ["CHERI_CAPTABLE_ABI"] = self.cheri_cap_table_abi
-
     @property
     def dollar_path_with_other_tools(self) -> str:
         return str(self.other_tools_dir / "bin") + ":" + os.getenv("PATH")
--- a/pycheribuild/config/chericonfig.py
+++ b/pycheribuild/config/chericonfig.py
@@ -158,8 +158,8 @@ class CheriConfig(object):
             "libcompat-buildenv", "-libcheri-buildenv", group=loader.freebsd_group,
             help="Open a shell with the right environment for building compat libraries.")
 
-        self.cheri_cap_table_abi = loader.add_option("cap-table-abi", help_hidden=True, default="pcrel",
-                                                     choices=("pcrel", "plt", "legacy", "fn-desc"),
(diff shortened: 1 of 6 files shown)


bsdjhb:
I've only tested this with cheribsd so far.

pycheribuild/config/jenkinsconfig.py
```diff
@@ -176,9 +176,9 @@ def sdk_archive_path(self):
 
     @property
     def cheri_sdk_isa_name(self):
-        guessed_abi_suffix = "cap-table-" + self.cheri_cap_table_abi
-        if self.cheri_cap_table_abi == "legacy":
-            guessed_abi_suffix = "legacy"
+        guessed_abi_suffix=""
```
jrtc27, line 179:
```suggestion
        guessed_abi_suffix = ""
```

pycheribuild/projects/project.py
```diff
@@ -1711,7 +1711,7 @@ def default_compiler_flags(self):
                            # "-Xclang", "-cheri-bounds=everywhere-unsafe"])
                            ])
         # Add mxcaptable for projects that need it
-        if self.compiling_for_cheri() and self.config.cheri_cap_table_abi != "legacy":
+        if self.compiling_for_cheri():
```
jrtc27, line 1714: This really was to be guarded by MIPS... are we not getting `clang-10: warning: argument unused during compilation: '-mxcaptable' [-Wunused-command-line-argument]` left right and centre on RISC-V? Maybe because we don't set the `force_*_linkage` very often?

arichardson: Yeah I think it's only WebKit and PostgreSQL when statically linked.

pycheribuild/config/target_info.py
```diff
@@ -491,7 +490,7 @@ def cheri_config_suffix(self, config: "CheriConfig"):
             # MIPS supports 128/256 -> include that in the configuration
             result += config.mips_cheri_bits_str
         if self.is_hybrid_or_purecap_cheri():
-            if config.cheri_cap_table_abi != self.DEFAULT_CAP_TABLE_ABI:
```
arichardson, line 494: I guess it would be good to not use a separate if it matches the default but since this is an option for very few people I think "don't pass the wrong argument" is a reasonable solution.

pycheribuild/config/chericonfig.py
```diff
@@ -409,7 +409,8 @@ def load(self):
     def _initialize_derived_paths(self):
         # Set CHERI_BITS variable to allow e.g. { cheribsd": { "install-directory": "~/rootfs${CHERI_BITS}" } }
         os.environ["CHERI_BITS"] = self.mips_cheri_bits_str
-        os.environ["CHERI_CAPTABLE_ABI"] = self.cheri_cap_table_abi
+        if self.cheri_cap_table_abi:
```
arichardson, line 412: This can probably go completely. I believe I only added it to allow different install/build dirs in the config file using 
`"install-directory": "/a/b/foo-$CHERI_CAPTABLE_ABI"`. But that's done automatically now.

arichardson:
If you'd like (almost) the equivalent of the CI to run before pushing, you can run
`ln -s ../../githooks/pre-push.sh .git/hooks/pre-push`

PR #149, opened 2020-07-28
Allow compiling kernel modules on RISC-V.

bsdjhb:
While here, move a comment about the malo module failure to the right
place.

### DIFF
--- a/pycheribuild/projects/cross/cheribsd.py
+++ b/pycheribuild/projects/cross/cheribsd.py
@@ -438,7 +438,8 @@ class BuildFreeBSD(BuildFreeBSDBase):
             self.make_args.set(LOCAL_XTOOL_DIRS="lib/libnetbsd usr.sbin/makefs usr.bin/mkimg")
             # Don't build ZFS for CHERI hybrid/purecap kernels. It's marked as broken in the kernel build so no
             # point building it for userspace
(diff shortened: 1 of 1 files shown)


bsdjhb:
Individual kernel configs can still opt out of modules via makeoptions.

jrtc27:
Out of interest, have you tried letting it build malo? It's possible this bug has inadvertently been fixed or no longer gets triggered after the two most recent upstream LLVM merges.

bsdjhb:
I can try that easily enough

bsdjhb:
I haven't tried building malo with the master branch btw, do we want to wait to push that until we update LLVM master?  (When are we next planning on updating CheriBSD and LLVM master anyway?)

arichardson:
I'd be happy to merge LLVM any time but it needs a cheribsd merge first.

brooksdavis:
I'm working on a CheriBSD merge now (currently verifying that LLVM master builds CheriBSD dev)

bsdjhb:
I think brooks merged CheriBSD so that leaves an llvm merge to 'master' as a dependency of this I think?

arichardson:
I've just merged LLVM dev to master, so will merge this on Monday to give people time to rebuild LLVM.

bsdjhb:
Ok to merge this now?

PR #155, opened 2020-08-20
Add --sysroot-install-dir configuration option

nwf:
cheribuild currently installs the sysroots under the sdk dir, which is derived
from the tools_root, which, in turn, defaults to the output_dir.  When
attempting to share tools directories between long-lived branches of cheribsd
(for example), this derivation chain implies that the sysroots end up in the
nominally shared tools directory, which is not desirable.  This change allows
cheribuild to still use a shared compiler and qemu (and so on) and yet use
different sysroots.

### DIFF
--- a/pycheribuild/config/defaultconfig.py
+++ b/pycheribuild/config/defaultconfig.py
@@ -123,6 +123,9 @@ class DefaultCheriConfig(CheriConfig):
                                                  default=lambda p, cls: p.output_root, group=loader.path_group,
                                                  help="The directory to find sdk and bootstrap tools (default: "
                                                       "'<OUTPUT_ROOT>')")
+        self.sysroot_pfx = loader.add_path_option("sysroot-install-dir",
+                                                  default=lambda p, cls: p.tools_root, group=loader.path_group,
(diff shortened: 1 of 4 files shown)


pycheribuild/config/defaultconfig.py
```diff
@@ -123,13 +123,17 @@ def __init__(self, loader: ConfigLoaderBase, available_targets: list):
                                                  default=lambda p, cls: p.output_root, group=loader.path_group,
                                                  help="The directory to find sdk and bootstrap tools (default: "
                                                       "'<OUTPUT_ROOT>')")
+        self.sysroot_pfx = loader.add_path_option("sysroot-root",
```
arichardson, line 126: sysroot-root looks a bit odd, maybe sysroot-install-dir?

arichardson (changes requested):
This also needs to be added to JenkinsConfig (and maybe the MockConfig for the unit tests)

arichardson (approved):

arichardson (approved):
Just noticed that the commit message should be updated. Once that is done we can merge this.

PR #156, opened 2020-08-25
Add KGDB support

brooksdavis:


arichardson (approved):
LGTM if it works for you. Would be nice if we didn't have to build two separate GDB projects.

jrtc27:
I think the KGDB one is a superset of the GDB one, so in theory that could replace it. The only issue is it requires building on FreeBSD currently. But yes, that's where we'd like to end up ultimately.

PR #157, opened 2020-08-27
FETT and caprevoke kernel support

brooksdavis:


pycheribuild/projects/cross/cheribsd.py
```diff
@@ -350,8 +350,14 @@ def default_kernel_config(self):
             # TODO: purecap/hybrid kernel
             if xtarget.is_hybrid_or_purecap_cheri():
                 assert isinstance(self, BuildCHERIBSD)
+                if self.caprevoke_kernel:
+                    if self.build_fett_kernels:
```
jrtc27, line 354: This changes the meaning of build-fett-kernels; previously it behaved like the FPGA kernels and would just add extra kernels to build but now it changes which kernel is in use. I think this new meaning should be a separate option?

brooksdavis: I see your point, but due to time constraints am planning to leave it as is now (I also need to follow up with some purecap changes).

PR #160, opened 2020-09-29
Various fixes for docker build

jedevc:
Fix a number of issues that prevented running cheribuild in the docker configuration. I'm new to this, so hopefully these are all reasonable fixes!

- Install missing samba dependency in Dockerfile
- Install missing texinfo dependency in Dockerfile (required for building GDB)
- Don't attempt to remove schg flag on non-BSD systems, the chflags command doesn't exist on Linux

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -8,7 +8,8 @@ RUN apt-get update && apt-get install -y \
   git \
   python3-minimal \
   lsb-release \
-  wget
+  wget \
+  samba
 
 # RUN git config --global http.sslVerify false
 # RUN cd /tmp && git clone https://github.com/arichardson/bmake && cd bmake \
--- a/pycheribuild/__main__.py
+++ b/pycheribuild/__main__.py
@@ -168,7 +168,7 @@ def real_main():
             docker_dir_mappings = [
                 # map cheribuild and the sources read-only into the container
                 "-v", cheribuild_dir + ":/cheribuild:ro",
-                "-v", str(cheri_config.source_root.absolute()) + ":/source:ro",
+                "-v", str(cheri_config.source_root.absolute()) + ":/source",
                 # build and output are read-write:
                 "-v", str(cheri_config.build_root.absolute()) + ":/build",
                 "-v", str(cheri_config.output_root.absolute()) + ":/output",
--- a/pycheribuild/projects/cross/cheribsd.py
+++ b/pycheribuild/projects/cross/cheribsd.py
@@ -1235,10 +1235,11 @@ class BuildCHERIBSD(BuildFreeBSD):
                 self.warning("Unsupported architecture for FETT kernels")
 
     def _remove_schg_flag(self, *paths: "typing.Iterable[str]"):
-        for i in paths:
-            file = self.install_dir / i
-            if file.exists():
-                self.run_cmd("chflags", "noschg", str(file))
+        if shutil.which("chflags"):
+            for i in paths:
+                file = self.install_dir / i
+                if file.exists():
+                    self.run_cmd("chflags", "noschg", str(file))
 
     def _remove_old_rootfs(self):
         if not self.config.skip_buildworld:


pycheribuild/__main__.py
```diff
@@ -168,7 +168,7 @@ def real_main():
             docker_dir_mappings = [
                 # map cheribuild and the sources read-only into the container
                 "-v", cheribuild_dir + ":/cheribuild:ro",
-                "-v", str(cheri_config.source_root.absolute()) + ":/source:ro",
+                "-v", str(cheri_config.source_root.absolute()) + ":/source",
```
arichardson, line 171: Ideally, the source tree could remain read-only. I guess this is only required for GDB?

docker/Dockerfile
```diff
@@ -29,7 +30,7 @@ RUN echo "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-8 main" > /etc/a
 # RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 1E9377A2BA9EF27F
 RUN apt-get update && apt-get install -y clang-8 lld-8
 
-RUN apt-get update && apt-get install -y cmake
+RUN apt-get update && apt-get install -y cmake texinfo
```
arichardson, line 33: Thanks, I haven't used the docker configuration lately. I guess texinfo is only needed for GDB?

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -1242,7 +1242,7 @@ def _remove_schg_flag(self, *paths: "typing.Iterable[str]"):
 
     def _remove_old_rootfs(self):
         if not self.config.skip_buildworld:
-            if self.install_as_root:
+            if self.install_as_root and not self.crossbuild:
```
arichardson, line 1245: I'd prefer if this was done using a `if OSInfo.IS_FREEBSD` or `shutil.which("chflags")` check inside `_remove_schg_flag()`.

arichardson:
Thank you very much for your contribution. Ideally GDB wouldn't attempt to rebuild the texinfo files. Not sure what changed there lately.

pycheribuild/__main__.py
jedevc: Ah yes sorry! This is needed because the script clones git repositories into the /source directory - if it's set to read-only, the operation fails, I forgot to put this in the commit message, I'll fix that up with the other stuff.

I suppose ideally the cloning would be done *before* starting up docker, but that's currently not the way it's done, I'll take a look and see how easy modifying that would be.

docker/Dockerfile
jedevc: Yup, the texinfo is for GDB, you're right, it should be possible to see if there's a way to suppress the texinfo files.

jedevc:
Thanks for your quick comments! I'll have a look and get back to you!

arichardson:
Could you see if GDB now works without texinfo installed after https://github.com/CTSRD-CHERI/cheribuild/commit/0582fff7ad0c06177864c94e8101b329695b8de0

It appears I change the path to `false` in https://github.com/CTSRD-CHERI/cheribuild/commit/5da7da9b86fe53c7982d43400762e5e22db1e517 which fixed macOS without texinfo but probably broke Linux without texinfo.

pycheribuild/__main__.py
jedevc: Hm, definitely doesn't look trivially easy to do, as the docker wrapper is only really understood at the top level of the build system, while the cloning is done later during compilation of the various projects. It might be worth revisiting this another time, but I think it might be a bit complex for this set of fixes?

jedevc:
Yup that fixes the texinfo dependency requirement!

pycheribuild/__main__.py
arichardson: That obviously makes sense.
I only ever used the docker wrapper with already-cloned repositories, so I never hit that issue and added the RO flag to ensure that the source directory remains the same. But without major refactoring, we will have to continue cloning inside so removing the flag is the best solution for now.

arichardson (approved):
Thanks! Will merge this once CI has completed.

PR #162, opened 2020-10-08
Use rootfs-amd64 for cheribsd-amd64.

bsdjhb:


bsdjhb:
This doesn't delete the old rootfs-x86 though, so perhaps this should have that added using whatever you use for the mips64 changes.

PR #165, opened 2020-10-16
Fix generation of purecap kernel MFS root kernel config.

qwattash:
Prevent appending duplicate -NODEBUG.

### DIFF
--- a/pycheribuild/projects/cross/cheribsd.py
+++ b/pycheribuild/projects/cross/cheribsd.py
@@ -1397,12 +1397,18 @@ class BuildCheriBsdMfsKernel(SimpleProject):
             benchmark_suffix = "-NODEBUG"
         # also build the benchmark kernel:
         if benchmark_suffix:
-            kernel_configs.append(default_kernconf + benchmark_suffix)
+            if default_kernconf.endswith(benchmark_suffix):
+                kernel_configs.append(default_kernconf[0:-len(benchmark_suffix)])
+            else:
+                kernel_configs.append(default_kernconf + benchmark_suffix)
         if self.build_fpga_kernels:
             fpga_conf = self.fpga_kernconf
             kernel_configs.append(fpga_conf)
             if benchmark_suffix:
-                kernel_configs.append(fpga_conf + benchmark_suffix)
+                if fpga_conf.endswith(benchmark_suffix):
+                    kernel_configs.append(fpga_conf[0:-len(benchmark_suffix)])
+                else:
+                    kernel_configs.append(fpga_conf + benchmark_suffix)
         if self.config.clean:
             for kernconf in kernel_configs:
                 kernel_dir = self.build_cheribsd_instance.kernel_objdir(kernconf)


jrtc27:
Why? I deliberately fixed it so -NODEBUG was preserved.

jrtc27:
Ok, the problem is BuildCheriBsdMfsKernel.process unconditionally adds `benchmark_suffix`. It should only do so if not already present.

qwattash:
@jrtc27 Latest purecap jenkins build fails because there is a duplicate -NODEBUG appended. 
https://ctsrd-build.cl.cam.ac.uk/blue/organizations/jenkins/CheriBSD-pipeline/detail/PR-614/22/pipeline/790/
I think this should generate the correct set of kernels, I tried locally a pretend build.

qwattash:
Let me change it so that it appends conditionally, probably more robust

jrtc27:
Maybe we want it to instead _strip_ it in that case, if the objective is to build all the variants?

qwattash:
Yes, I think the current change in this PR should stay there, as the default kernel should be the neutral variant, and then we append suffixes to it, if needed.

jrtc27:
No I meant that `BuildCheriBsdMfsKernel.process` should be:

```python3
        # also build the benchmark/normal kernel:
        if benchmark_suffix:
            if default_kernconf.endswith(benchmark_suffix):
                kernel_configs.append(default_kernconf[0:-len(benchmark_suffix)])
            else:
                kernel_configs.append(default_kernconf + benchmark_suffix)
```

jrtc27:
And similarly for fpga_conf later

jrtc27:
Your solution breaks if `benchmark_suffix` is `None`, no?

jrtc27:
Also, `strip`'s string argument is a set of characters to strip, so it will remove anything matching `[-NODEBUG]`, which is not what you want.

qwattash:
`strip(None)` behaves as `strip()` with no argument, unless some older python version does something weird. So it should be fine, if a bit contrived.

jrtc27:
.. which then strips whitespace?

qwattash:
There should be no whitespace at the beginning or end of the string for a file name, I would hope. But I can go with the more explicit solution if it reads better.

jrtc27:
My point is, `strip` is entirely the wrong interface. It's broken if you have a non-None `benchmark_suffix` due to being a character set not a suffix (and even if it were what you think it is it'd match *prefixes* too), and if it's None then it tries to do a bunch of work that ends up being a no-op. Python 3.9 finally adds nice methods to remove prefixes and suffixes, but that's very new.

qwattash:
Ok, I agree that I was doing something silly. Should be correct now.

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -319,7 +319,7 @@ def default_kernel_config(self):
                         return "CHERI-CAPREVOKE-QEMU-FETT"
                     return "CHERI-CAPREVOKE-QEMU"
                 if self.purecap_kernel:
-                    return "CHERI-PURECAP-QEMU-NODEBUG"
+                    return "CHERI-PURECAP-QEMU"
```
jrtc27, line 322: Please revert this; we don't want everyone to be waiting twice as long for QEMU to boot...

jrtc27:
Looks fine now from my perspective but please update the commit message to reflect the new approach.

arichardson (approved):

bsdjhb:
Wasn't the slow boot with DEBUG made better when we added WITNESS_SKIPSPIN to riscv's GENERIC kernel?

jrtc27:
I don't know if anyone measured the difference for the purecap kernel.

arichardson:
> Wasn't the slow boot with DEBUG made better when we added WITNESS_SKIPSPIN to riscv's GENERIC kernel?

It "feels" faster after WITNESS_SKIPSPIN but I didn't do any measurements.

PR #166, opened 2020-10-22
Move TARGET_ARCH computation to common code

jrtc27:
This is motivated by perl, where I want to include the FreeBSD TARGET_ARCH in its targetarch variable that gets printed in perl --version (and used in some directory names).

### DIFF
--- a/pycheribuild/config/compilation_targets.py
+++ b/pycheribuild/config/compilation_targets.py
@@ -256,6 +256,18 @@ class FreeBSDTargetInfo(_ClangBasedTargetInfo):
         # TODO: do we need any special cases here?
         return target.cpu_architecture.value + common_suffix
 
+    @property
+    def freebsd_target_arch(self):
+        mapping = {
+            CPUArchitecture.AARCH64: "aarch64",
+            CPUArchitecture.ARM32: "armv7",
+            CPUArchitecture.I386: "i386",
+            CPUArchitecture.MIPS64: self.config.mips_float_abi.freebsd_target_arch(),
+            CPUArchitecture.RISCV64: "riscv64",
+            CPUArchitecture.X86_64: "amd64",
+            }
+        return mapping[self.target.cpu_architecture]
+
     @classmethod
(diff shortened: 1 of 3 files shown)


pycheribuild/projects/cross/cheribsd.py
```diff
@@ -355,26 +355,22 @@ def _stdout_filter(self, line: bytes):
 
     @property
     def arch_build_flags(self):
+        result = {
+            "TARGET_ARCH": self.target_info.target_arch,
+            }
         if self.compiling_for_mips(include_purecap=True):
-            target_arch = self.config.mips_float_abi.freebsd_target_arch()
-            if self.crosscompile_target.is_cheri_purecap():
-                target_arch += "c" + self.config.mips_cheri_bits_str  # build purecap
```
jrtc27, line 361: Hm, I guess I need to inherit this special case :( the legacy of CHERI-256 lives on...

arichardson:
Generally looks good to me, although I'd prefer the name `freebsd_target_arch` since that seems more self-explanatory.

jrtc27:
> Generally looks good to me, although I'd prefer the name `freebsd_target_arch` since that seems more self-explanatory.

Done, and moved to just FreeBSDTargetInfo as a result, which is a lot nicer than all the ValueError stubs.

pycheribuild/config/compilation_targets.py
```diff
@@ -256,6 +256,18 @@ def triple_for_target(cls, target: "CrossCompileTarget", config: "CheriConfig",
         # TODO: do we need any special cases here?
         return target.cpu_architecture.value + common_suffix
 
+    @property
+    def freebsd_target_arch(self):
+        mapping = {
+            CPUArchitecture.AARCH64: "aarch64",
+            CPUArchitecture.ARM32: "armv7",
```
jrtc27, line 263: Wasn't sure whether to add this or not, but I thought might as well give it a sensible value.

arichardson: LGTM

arichardson (approved):

PR #167, opened 2020-10-23
Firmware fun

arichardson:
😢

### DIFF
--- a/pycheribuild/config/compilation_targets.py
+++ b/pycheribuild/config/compilation_targets.py
@@ -861,7 +861,7 @@ class ArmNoneEabiGccTargetInfo(TargetInfo):
 
     @property
     def linker(self) -> Path:
-        return self.bindir / (self.binary_prefix + "ld")
+        return self.bindir / (self.binary_prefix + "ld.bfd")
 
     @property
     def ar(self) -> Path:
--- a/pycheribuild/config/compilation_targets.py
+++ b/pycheribuild/config/compilation_targets.py
@@ -164,7 +164,11 @@ class _ClangBasedTargetInfo(TargetInfo, metaclass=ABCMeta):
             if self.is_baremetal() or self.is_rtems():
                 # Both RTEMS and baremetal FreeRTOS are linked above 0x80000000
                 result.append("-mcmodel=medium")
-
+        elif self.target.is_aarch64(include_purecap=True):
+            if self.target.is_cheri_hybrid():
+                result += ["-march=morello", "-mabi=aapcs"]
+            elif self.target.is_cheri_purecap():
+                result += ["-march=morello+c64", "-mabi=purecap"]
         else:
             self.project.warning("Compiler flags might be wong, only native + MIPS checked so far")
         return result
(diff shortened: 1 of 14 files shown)


pycheribuild/projects/fvp_firmware.py
```diff
@@ -0,0 +1,264 @@
+#
```
jrtc27, line 1: Should this not be a morello_firware.py given the FVP and Morello board are supposed to be very similar? (In theory I assume they're meant to be able to run basically the same firmware, but in practice I'm sure that's not actually the case...)

pycheribuild/config/compilation_targets.py
```diff
@@ -755,6 +755,135 @@ def _get_rootfs_project(self, xtarget: CrossCompileTarget) -> "Project":
         return BuildNewlib.get_instance(self.project, cross_target=xtarget)
 
 
+class MorelloBaremetalTargetInfo(_ClangBasedTargetInfo):
+    shortname = "Morello-Baremetal"
+
+    @property
+    def cmake_system_name(self) -> str:
+        return "Generic"  # Unknown platform, CMake expects the value to be set to Generic
```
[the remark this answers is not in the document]
arichardson: "Generic" is the value that CMake expects for baremetal targets. Not that we are using this toolchain for any CMake targets.

pycheribuild/projects/cross/gdb.py
```diff
@@ -117,6 +118,10 @@ def setup(self):
             "--with-guile=no",
             ])
 
+        if self.compiling_for_host() and self.target_info.is_macos():
+            # Allow building ld.bfd by building for FreeBSD
+            self.configure_args.append("--target=x86_64-unknown-freebsd13")
```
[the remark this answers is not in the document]
arichardson: Well it doesn't matter what target I use since we enable all of them. This is just to force the binutils-gdb autoconf to not disable ld.

pycheribuild/projects/fvp_firmware.py
arichardson: Yeah they should be the same but until I know for sure I'll keep the name FVP. I believe some of the configurations are a bit different for FVP.

pycheribuild/projects/run_fvp.py
```diff
@@ -114,17 +120,19 @@ def execute_fvp(self, args: list, disk_image_path: Path = None, firmware_path: P
...
-            return 0
+        if self._fvp_revision is None:
+            revpath = "sw/ARM_Fast_Models_FVP_Morello/rev"
+            try:
+                if self.use_docker_container:
+                    rev = self.run_cmd(["docker", "run", "-it", "--rm", self.container_name, "cat",
+                                        "/opt/FVP_Morello/" + revpath], capture_output=True,
+                                       run_in_pretend_mode=True).stdout
+                    self._fvp_revision = int(rev.strip())
+                self._fvp_revision = int(self.read_file(self.install_dir / revpath))
```
arichardson, line 131: This needs an else now that there is no longer an early return.

Much better than returning zero!

jrtc27: Oops yes

PR #168, opened 2020-11-09
FreeRTOS/Baremetal RISC-V32 Support

heshamelmatary:


pycheribuild/config/compilation_targets.py
```diff
@@ -165,7 +165,8 @@ def essential_compiler_and_linker_flags_impl(cls, ti: "_ClangBasedTargetInfo", *
                     result.append("-cheri=" + config.mips_cheri_bits_str)
                     result.append("-mcpu=beri")
         elif target.is_riscv(include_purecap=True):
-            assert target.cpu_architecture == CPUArchitecture.RISCV64
+            assert target.cpu_architecture == CPUArchitecture.RISCV64 or \
+                   target.cpu_architecture == CPUArchitecture.RISCV32
```
jrtc27, line 169: This can now go.

pycheribuild/config/compilation_targets.py
```diff
@@ -199,9 +200,11 @@ def riscv_arch_string(self):
         # Use the insane RISC-V arch string to enable CHERI
         if self.is_baremetal():
             # Baremetal/FreeRTOS only supports softfloat
-            arch_string = "rv64imac"
+            arch_string = "rv64imac" if self.target.cpu_architecture == CPUArchitecture.RISCV64 else \
+                          "rv32imac"
         else:
-            arch_string = "rv64imafdc"
+            arch_string = "rv64imafdc" if self.target.cpu_architecture == CPUArchitecture.RISCV64 else \
+                          "rv32imacf"
```
jrtc27, line 205: This is not in the right order. But why is it using F? Our P1s have no FPU.

pycheribuild/config/compilation_targets.py
```diff
@@ -199,9 +200,11 @@ def riscv_arch_string(self):
         # Use the insane RISC-V arch string to enable CHERI
         if self.is_baremetal():
             # Baremetal/FreeRTOS only supports softfloat
-            arch_string = "rv64imac"
+            arch_string = "rv64imac" if self.target.cpu_architecture == CPUArchitecture.RISCV64 else \
+                          "rv32imac"
```
jrtc27, line 202:
```suggestion
            if self.target.cpu_architecture == CPUArchitecture.RISCV64:
                arch_string = "rv64imac"
            else:
                arch_string = "rv32imac"
```
looks better I think given the long condition (repeated for all the similar diffs below).

pycheribuild/config/compilation_targets.py
```diff
@@ -211,17 +214,25 @@ def riscv_arch_string(self):
     def riscv_abi(self):
         assert self.target.is_riscv(include_purecap=True)
         if self.target.is_cheri_purecap():
-            return "l64pc128d"  # 64-bit double-precision hard-float + purecap
+            # 64-bit double-precision hard-float + purecap
+            return "l64pc128d" if self.target.cpu_architecture == CPUArchitecture.RISCV64 else \
+                   "il32pc64f"  # 32-bit single-precision soft-float + purecap
```
jrtc27, line 217: Soft-float means no f.

pycheribuild/config/compilation_targets.py
```diff
@@ -211,17 +214,25 @@ def riscv_arch_string(self):
...
         else:
-            return "lp64d"  # 64-bit double-precision hard-float
+            # 64-bit double-precision hard-float
+            return "lp64d" if self.target.cpu_architecture == CPUArchitecture.RISCV64 else \
+                   "ilp32f"  # 32-bit single-precision soft-float
```
jrtc27, line 221: Ditto

pycheribuild/config/target_info.py
```diff
@@ -274,6 +275,8 @@ def capability_size(self):
             return self.config.mips_cheri_bits // 8
         elif self.target.is_hybrid_or_purecap_cheri([CPUArchitecture.RISCV64]):
             return 16  # RISCV64 uses 128-bit capabilities
+        elif self.target.is_hybrid_or_purecap_cheri([CPUArchitecture.RISCV32]):
+            return 8  # RISCV32 uses 64-bit capabilities
```
jrtc27, line 279:
```suggestion
        elif self.target.is_hybrid_or_purecap_cheri([CPUArchitecture.RISCV32]):
            return 8  # RISCV32 uses 64-bit capabilities
        elif self.target.is_hybrid_or_purecap_cheri([CPUArchitecture.RISCV64]):
            return 16  # RISCV64 uses 128-bit capabilities
```

pycheribuild/config/target_info.py
```diff
@@ -543,7 +546,8 @@ def is_mips(self, include_purecap: bool = None):
         return self._check_arch(CPUArchitecture.MIPS64, include_purecap)
 
     def is_riscv(self, include_purecap: bool = None):
-        return self._check_arch(CPUArchitecture.RISCV64, include_purecap)
+        return self._check_arch(CPUArchitecture.RISCV64, include_purecap) or \
+            self._check_arch(CPUArchitecture.RISCV32, include_purecap)
```
jrtc27, line 550: Might be nicer to make `_check_arch` take a list instead.

pycheribuild/projects/cross/freertos.py
```diff
@@ -47,17 +47,35 @@ class BuildFreeRTOS(CrossCompileAutotoolsProject):
     needs_sysroot = False  # We don't need a complete sysroot
     supported_architectures = [
         CompilationTargets.BAREMETAL_NEWLIB_RISCV64_PURECAP,
-        CompilationTargets.BAREMETAL_NEWLIB_RISCV64]
+        CompilationTargets.BAREMETAL_NEWLIB_RISCV64,
+        CompilationTargets.BAREMETAL_NEWLIB_RISCV32_PURECAP,
+        CompilationTargets.BAREMETAL_NEWLIB_RISCV32]
```
jrtc27, line 52: Reorder this to be 32, 32_PURECAP, 64, 64_PURECAP, i.e. alphabetical.

pycheribuild/projects/cross/freertos.py
```diff
@@ -47,17 +47,35 @@ class BuildFreeRTOS(CrossCompileAutotoolsProject):
...
     default_install_dir = DefaultInstallDir.SYSROOT
 
     # FreeRTOS Demos to build
     supported_freertos_demos = [
         # Generic/simple (CHERI-)RISC-V Demo that runs main_blinky on simulators
         # and simple SoCs
-        "RISC-V-Generic"]
+        "RISC-V-Generic",
+        # CHERI-RISC-V sophisticated Demo with more advanced device drivers
+        # and currently only runs on FPGA-GFE, purecap
```
jrtc27, line 61: FPGA-GFE is a strange expression; just use "the GFE".

pycheribuild/projects/cross/freertos.py
```diff
@@ -47,17 +47,35 @@ class BuildFreeRTOS(CrossCompileAutotoolsProject):
...
+        "RISC-V_Galois_P1"]
 
     # Map Demos and the FreeRTOS apps we support building/running for
-    supported_demo_apps = {"RISC-V-Generic": ["main_blinky"]}
+    supported_demo_apps = {"RISC-V-Generic": ["main_blinky"],
+                           "RISC-V_Galois_P1":
+                           ["main_blinky", "main_netboot",
```
jrtc27, line 67: This doesn't look like valid formatting, but even if flake8 is happy it could be formatted more nicely.

pycheribuild/projects/cross/freertos.py
```diff
@@ -47,17 +47,35 @@ class BuildFreeRTOS(CrossCompileAutotoolsProject):
...
+                               # The followings haven't been tested for purecap
+                               "main_testgen",
```
jrtc27, line 69: This is what gets used for the CWE tests, so it has been tested, but it's meaningless on its own.

pycheribuild/projects/cross/freertos.py
```diff
@@ -47,17 +47,35 @@ class BuildFreeRTOS(CrossCompileAutotoolsProject):
...
+                               "main_uart_malware",
+                               "main_rtc",
+                               "main_sd",
+                               "main_udp",
+                               "main_peekpoke",
+                               "main_tcp",
```
jrtc27, line 75: These got tested ages ago when we were working on the smart ballot box (though I forget whether peekpoke was made to derive capabilities or not, that might just have been the separate smart ballot box application that embedded an extended copy of it).

pycheribuild/projects/cross/freertos.py
```diff
@@ -47,17 +47,35 @@ class BuildFreeRTOS(CrossCompileAutotoolsProject):
...
+                               "main_gpio",
+                               "main_iic",
+                               "main_full"]}
```
jrtc27, line 78: Also tested way back when.

pycheribuild/projects/cross/freertos.py
```diff
@@ -76,22 +94,23 @@ def __init__(self, config: CheriConfig):
...
-            self.supported_freertos_demos.append("RISC-V_Galois_P1")
-            self.supported_demo_apps["RISC-V_Galois_P1"] = ["main_blinky", "main_netboot"]
+        if self.demo == "RISC-V_Galois_P1":
+            if "rv32" in self.demo_bsp:
+                self.make_args.set(XARCH="32")
+            else:
+                self.make_args.set(XARCH="64")
 
+        else:  # RISC-V-Generic
+            # Only build 64-bit FreeRTOS for RISC-V-Generic
```
jrtc27, line 110: Do one of:

* assert
* leave it and let the target fail (as you do for non-CHERI RISC-V_Galois_P1 demos)
* cheribuild magic so the incompatible combinations don't exist as targets

pycheribuild/projects/cross/freertos.py
```diff
@@ -109,7 +128,7 @@ def setup_config_options(cls, **kwargs):
             help="The FreeRTOS program to build.")  # type: str
 
         cls.demo_bsp = cls.add_config_option(
-            "bsp", metavar="BSP", show_help=True,
+            "demo_bsp", metavar="BSP", show_help=True,
```
jrtc27, line 131: That's not the naming style. You should perhaps put generic in there.

pycheribuild/projects/cross/freertos.py
```diff
@@ -47,17 +47,35 @@ class BuildFreeRTOS(CrossCompileAutotoolsProject):
...
     supported_architectures = [
         CompilationTargets.BAREMETAL_NEWLIB_RISCV64_PURECAP,
-        CompilationTargets.BAREMETAL_NEWLIB_RISCV64]
+        CompilationTargets.BAREMETAL_NEWLIB_RISCV64,
+        CompilationTargets.BAREMETAL_NEWLIB_RISCV32_PURECAP,
+        CompilationTargets.BAREMETAL_NEWLIB_RISCV32]
     default_install_dir = DefaultInstallDir.SYSROOT
 
     # FreeRTOS Demos to build
     supported_freertos_demos = [
```
jrtc27, line 56: I do feel each demo should probably be a separate Project, deriving from a common base class, and probably also each app should be a separate Project deriving from the demo's. Having to use options to specify which you want isn't a great interface.

jrtc27 (changes requested):

pycheribuild/config/target_info.py
arichardson: I think this is fine since there are no other cases that need a list.

pycheribuild/projects/cross/freertos.py
arichardson: The order used to matter since the first one was used when you omit the suffix, but now that the suffix is required alphabetical is better.

pycheribuild/config/compilation_targets.py
arichardson: Shouldn't this include f? I.e. the comment is wrong not the ABI? Or do we really want to assume that softfloat for all rv32 targets?

pycheribuild/config/compilation_targets.py
```diff
@@ -211,17 +212,25 @@ def riscv_arch_string(self):
...
 
     @property
     def riscv_softfloat_abi(self):
         assert self.target.is_riscv(include_purecap=True)
         if self.target.is_cheri_purecap():
-            return "l64pc128"  # 64-bit soft-float purecap
+            # 64-bit soft-float purecap
+            return "l64pc128" if self.target.cpu_architecture == CPUArchitecture.RISCV64 else \
```
arichardson, line 228: I prefer if/else statements if you need a line continuation for the ternary.

pycheribuild/config/compilation_targets.py
```diff
@@ -748,7 +757,8 @@ def triple_for_target(cls, target, config, include_version: bool) -> str:
                 return "mips64c{}-qemu-elf-purecap".format(config.mips_cheri_bits)
             return "mips64-qemu-elf"
         if target.is_riscv(include_purecap=True):
-            return "riscv64-unknown-elf"
+            return "riscv64-unknown-elf" if target.cpu_architecture == CPUArchitecture.RISCV64 else \
+                   "riscv32-unknown-elf"
```
arichardson, line 761:
```suggestion
            return target.cpu_architecture.value + "-unknown-elf"
```

pycheribuild/projects/cross/freertos.py
arichardson: That's probably nicer. However, each added projects incurs startup overheads due to all the command line flags that need to be added to argparse. If this is rarely used, I'd prefer making it a single option instead of a separate project.

pycheribuild/config/compilation_targets.py
jrtc27: Well it needs to be soft-float if the arch string doesn't have F. In theory we could have an FD hard-float RV32 implementation but we don't currently and it's not something that will be used in SSITH.

pycheribuild/config/compilation_targets.py
heshamelmatary: That was rather intended to be generic (in case we do have simulators or targets that support F in the future) if soft-float ABI isn't used. But right now, FreeRTOS and all baremetal projects only use the `riscv_softfloat_abi` and thus `rv32imacf` is never used yet

pycheribuild/config/compilation_targets.py
arichardson: yeah, but this is the generic case. IMO there should just be a supports_hardfloat option that the newlib targetinfo sets to false and then the ABI and arch_string can match again.

pycheribuild/config/compilation_targets.py
jrtc27: `rv32imafc` (NB: *not* `rv32imacf`) is a strange beast. Even for RV32 the Unix variant is intended to be `rv32imafdc` with `ilp32d` as the ABI.

pycheribuild/config/compilation_targets.py
jrtc27: Well actually newlib supports it just fine (I did setjmp/longjmp properly), it's just that the FreeRTOS context switch code doesn't save/restore floating point registers.

pycheribuild/config/compilation_targets.py
heshamelmatary: Right, I changed that to `rv32imafdc`

pycheribuild/config/compilation_targets.py
heshamelmatary: Yeah, also RTEMS/newlib RISC-V32 support hard-float. I fixed the comment and made RISC-V32 hard-float default to `il32pc64d` for ports that may (and choose to) support it (eg, RTEMS on spike/qemu)

pycheribuild/projects/cross/freertos.py
heshamelmatary: better now?

pycheribuild/projects/cross/freertos.py
heshamelmatary: OK I removed it based on that comment (and my impression it needs NIX build which we don't support)

pycheribuild/projects/cross/freertos.py
heshamelmatary: I don't remember peekpoke needs work for CHERI when I made it run on `RISC-V-Generic`

pycheribuild/projects/cross/freertos.py
heshamelmatary: Not sure I get that?

pycheribuild/config/compilation_targets.py
```diff
@@ -211,17 +216,37 @@ def riscv_arch_string(self):
...
     @property
     def riscv_softfloat_abi(self):
         assert self.target.is_riscv(include_purecap=True)
         if self.target.is_cheri_purecap():
-            return "l64pc128"  # 64-bit soft-float purecap
+            if self.target.cpu_architecture == CPUArchitecture.RISCV32:
+                # 32-bit soft-float + purecap
+                return "ilp32"
+            else:
+                # 64-bit soft-float + purecap
```
arichardson, line 241: Those are the non-purecap ABIs, you need to swap the cases.

pycheribuild/projects/cross/freertos.py
arichardson: None of the other command line flags use an underscore, it should probably be:
```suggestion
            "demo-bsp", metavar="BSP", show_help=True,
```

pycheribuild/projects/cross/freertos.py
```diff
@@ -76,22 +94,24 @@ def __init__(self, config: CheriConfig):
...
         # Add compiler-rt location to the search path
         # self.make_args.set(LDFLAGS="-L"+str(self.compiler_resource / "lib"))
 
-        if self.target_info.target.is_cheri_purecap():
-            # CHERI-RISC-V sophisticated Demo with more advanced device drivers
-            # and currently only runs on FPGA-GFE, purecap
-            self.supported_freertos_demos.append("RISC-V_Galois_P1")
-            self.supported_demo_apps["RISC-V_Galois_P1"] = ["main_blinky", "main_netboot"]
+        if self.demo == "RISC-V_Galois_P1":
+            if "rv32" in self.demo_bsp:
```
arichardson, line 104: It seems to me like XARCH should only depend on whether the `crosscompile_target` is rv32 or rv64. Requesting one that doesn't match should just cause a `fatal_error`.

arichardson (changes requested):
Mostly looks fine, just a few issues.

pycheribuild/config/compilation_targets.py
```diff
@@ -211,17 +216,37 @@ def riscv_arch_string(self):
     def riscv_abi(self):
         assert self.target.is_riscv(include_purecap=True)
         if self.target.is_cheri_purecap():
```
arichardson, line 218:
```suggestion
        if self.is_baremetal():
            return self.riscv_softfloat_abi()  # Baremetal/FreeRTOS only supports softfloat
        if self.target.is_cheri_purecap():
```
I think you need something like this to make the arch string and the ABI match up.

pycheribuild/config/compilation_targets.py
```diff
@@ -210,18 +215,42 @@ def riscv_arch_string(self):
...
+            else:
+                # 64-bit double-precision hard-float
+                return "lp64d"
 
     @property
     def riscv_softfloat_abi(self):
         assert self.target.is_riscv(include_purecap=True)
         if self.target.is_cheri_purecap():
-            return "l64pc128"  # 64-bit soft-float purecap
+            if self.target.cpu_architecture == CPUArchitecture.RISCV32:
```
jrtc27, line 241: Purecap and hybrid/vanilla are now round the wrong way, and the comments have got worse.

pycheribuild/config/compilation_targets.py
```diff
@@ -199,9 +198,15 @@ def riscv_arch_string(self):
         # Use the insane RISC-V arch string to enable CHERI
         if self.is_baremetal():
             # Baremetal/FreeRTOS only supports softfloat
-            arch_string = "rv64imac"
+            if self.target.cpu_architecture == CPUArchitecture.RISCV32:
+                arch_string = "rv32imac"
```
jrtc27, line 202: Better to pull out the rv32i/rv64i prefix as a base_arch that then gets imac or imafdc appended.

pycheribuild/config/compilation_targets.py
```diff
@@ -983,7 +1021,8 @@ class CompilationTargets(BasicCompilationTargets):
                                     CHERIBSD_MIPS_HYBRID, CHERIBSD_MIPS_NO_CHERI, CHERIBSD_MIPS_PURECAP]
 
     ALL_SUPPORTED_BAREMETAL_TARGETS = [BAREMETAL_NEWLIB_MIPS64, BAREMETAL_NEWLIB_MIPS64_PURECAP,
-                                       BAREMETAL_NEWLIB_RISCV64, BAREMETAL_NEWLIB_RISCV64_PURECAP]
+                                       BAREMETAL_NEWLIB_RISCV64, BAREMETAL_NEWLIB_RISCV64_PURECAP,
+                                       BAREMETAL_NEWLIB_RISCV32, BAREMETAL_NEWLIB_RISCV32_PURECAP]
```
jrtc27, line 1025: Sort

pycheribuild/config/target_info.py
```diff
@@ -48,6 +48,7 @@ class CPUArchitecture(Enum):
     I386 = "i386"
     MIPS64 = "mips64"
     RISCV64 = "riscv64"
+    RISCV32 = "riscv32"
```
jrtc27, line 51: Sort

pycheribuild/projects/cross/freertos.py
```diff
@@ -76,22 +94,27 @@ def __init__(self, config: CheriConfig):
...
         self.make_args.set(NIX_ENV="no")
 
-        # Only build 64-bit FreeRTOS as cheribuild currently only supports building
-        # for RV64
-        self.make_args.set(RISCV_XLEN="64")
-
         # Set sysroot Makefile arg to pick up libc
         self.make_args.set(SYSROOT=str(self.sdk_sysroot))
 
+        # Is purecap?
```
jrtc27, line 100: Pointless comment

pycheribuild/projects/cross/freertos.py
```diff
@@ -76,22 +94,27 @@ def __init__(self, config: CheriConfig):
...
         # Add compiler-rt location to the search path
         # self.make_args.set(LDFLAGS="-L"+str(self.compiler_resource / "lib"))
 
-        if self.target_info.target.is_cheri_purecap():
-            # CHERI-RISC-V sophisticated Demo with more advanced device drivers
-            # and currently only runs on FPGA-GFE, purecap
-            self.supported_freertos_demos.append("RISC-V_Galois_P1")
-            self.supported_demo_apps["RISC-V_Galois_P1"] = ["main_blinky", "main_netboot"]
+        if self.demo == "RISC-V_Galois_P1":
+            if self.crosscompile_target.is_riscv32(include_purecap=self.is_purecap):
```
jrtc27, line 107: Setting include_purecap based on the current target info makes no sense, and it's also really weird to have XARCH=64 in an else branch that could be hit for riscv32-purecap if self.is_purecap is False (which can't happen, because it's all redundantly defined, but is just another reason why this is nonsense).

pycheribuild/projects/cross/freertos.py
```diff
@@ -76,22 +94,27 @@ def __init__(self, config: CheriConfig):
...
 
-        # Only build 64-bit FreeRTOS as cheribuild currently only supports building
-        # for RV64
-        self.make_args.set(RISCV_XLEN="64")
-
         # Set sysroot Makefile arg to pick up libc
         self.make_args.set(SYSROOT=str(self.sdk_sysroot))
 
+        # Is purecap?
+        self.is_purecap = self.target_info.target.is_cheri_purecap()
```
jrtc27, line 101: Why does this need to be a member variable?

pycheribuild/projects/cross/freertos.py
```diff
@@ -76,22 +94,27 @@ def __init__(self, config: CheriConfig):
...
+            else:
+                self.make_args.set(XARCH="64")
 
+        else:  # RISC-V-Generic
+            # Only build 64-bit FreeRTOS for RISC-V-Generic
+            assert(self.demo == "RISC-V-Generic")
+            self.make_args.set(RISCV_XLEN="64")
+
+        if self.target_info.target.is_cheri_purecap():
             self.make_args.set(EXTENSION="cheri")
```
jrtc27, line 118: This does not exist for RISC-V_Galois_P1.

pycheribuild/projects/cross/freertos.py
jrtc27: Besides, don't you want to give an error if someone's trying to build this demo as non-purecap given that's not supported by our fork?

jrtc27 (changes requested):

heshamelmatary:
Sorry, I only managed to have a look at this now.  Galois seems (as of last week) to have moved to the new official FreeRTOS release rather than the FreeRTOS-mirror one used here, so I'm closing this PR. I have another branch now that keeps track of that and builds their demo, including 32-bit support. I'll aim so submit a PR to merge that at some point. Thanks for all your feedback.

PR #169, opened 2020-11-10
gdb: add -lmd when compiling for target

nwf:
After 6edd51bc7326 (FreeBSD) / 6c4dac25e39d (CheriBSD), liblzma depends on
libmd, but this isn't automagically picked up by configure, so hold the build
system's hand.


arichardson (approved):
LGTM, I guess that's the problem with trying to build a static binary.

arichardson:
I guess we could save some space on the minimal images by linking libc+libmd dynamically (since they will exist even on the minimal images), but that would require some more invasive changes.

PR #172, opened 2020-11-30
add new webkit target

brettferdosi:
draft until the webkit repository is made public


jrtc27:
Some of those hunks look like they should be pulled out and can be merged straight away?

pycheribuild/projects/cross/crosscompileproject.py
```diff
@@ -50,7 +50,8 @@
 # avoids repeating this for every target than can be cross-built
 class CrossCompileMixin(object):
     do_not_add_to_targets = True
-    supported_architectures = CompilationTargets.ALL_SUPPORTED_CHERIBSD_AND_HOST_TARGETS
+    supported_architectures = (CompilationTargets.ALL_SUPPORTED_CHERIBSD_AND_HOST_TARGETS +
```
arichardson, line 53: This should be a separate commit, since it adds all the -morello targets.

arichardson: Also you should probably update ALL_SUPPORTED_CHERIBSD_AND_HOST_TARGETS instead.

pycheribuild/config/target_info.py
```diff
@@ -275,6 +275,8 @@ def capability_size(self):
             return self.config.mips_cheri_bits // 8
         elif self.target.is_hybrid_or_purecap_cheri([CPUArchitecture.RISCV64]):
             return 16  # RISCV64 uses 128-bit capabilities
+        elif self.target.is_hybrid_or_purecap_cheri([CPUArchitecture.AARCH64]):
```
arichardson, line 278: Can you split this into a separate commit?

pycheribuild/projects/cross/webkit.py
```diff
@@ -0,0 +1,128 @@
+#
```
arichardson, line 1: This looks pretty much identical to the other version that I wrote, maybe add a 
Co-authored-by: Alex Richardson <Alexander.Richardson@cl.cam.ac.uk>` line to the commit message.

arichardson (commented):
Generally LGTM, but we should probably split the non-webkit changes into separate commits.

pycheribuild/projects/cross/webkit.py
jrtc27: In that case can we make them share code?

arichardson:
This will need a few smaller changes to get `./cheribuild.py -p __run_everything__` to pass. Alternatively you could install the pre-push commit hook to run the tests locally before you push:
`ln -s ../../githooks/pre-push.sh .git/hooks/pre-push`

pycheribuild/projects/cross/webkit.py
arichardson: Sorry, I don't mean the existing the qtwebkit implementation, I meant the externally-managed private version that existed prior to open-sourcing the new webkit repo.

pycheribuild/config/compilation_targets.py
```diff
@@ -611,7 +611,7 @@ def essential_compiler_and_linker_flags(self) -> typing.List[str]:
 
     @property
     def must_link_statically(self):
-        return True  # dynamic linking is still experimental
+        return False
```
arichardson, line 618: This is already the default in the parent class, so you could just remove the override.

arichardson (approved):
1 minor comment otherwise LGTM.

pycheribuild/config/compilation_targets.py
```diff
@@ -973,7 +977,8 @@ class CompilationTargets(BasicCompilationTargets):
                                            CHERIBSD_MIPS_PURECAP, CHERIBSD_MIPS_HYBRID, CHERIBSD_MIPS_NO_CHERI]
     ALL_CHERIBSD_NON_MORELLO_TARGETS = ALL_CHERIBSD_MIPS_AND_RISCV_TARGETS + [CHERIBSD_AARCH64, CHERIBSD_X86_64]
     ALL_CHERIBSD_MORELLO_TARGETS = [CHERIBSD_MORELLO_PURECAP, CHERIBSD_MORELLO_HYBRID]
-    ALL_SUPPORTED_CHERIBSD_AND_HOST_TARGETS = ALL_CHERIBSD_NON_MORELLO_TARGETS + [BasicCompilationTargets.NATIVE]
+    ALL_SUPPORTED_CHERIBSD_AND_HOST_TARGETS = (ALL_CHERIBSD_NON_MORELLO_TARGETS + ALL_CHERIBSD_MORELLO_TARGETS +
```
jrtc27, line 980: No need for the parentheses

pycheribuild/projects/cross/qt5.py
```diff
@@ -433,6 +433,7 @@ def __init__(self, config):
             # self.configure_args.append("--disable-extras")  # can't add this to host build, it will fail otherwise
             # We have modified the ICU data Makefile so that ICU builds a big endian data archive
             self.configure_args.append("--with-data-packaging=archive")
+            self.COMMON_FLAGS.append("-fPIC")  # needed for morello hybrid webkit not to complain about relocations
```
jrtc27, line 436: Probably wants to be -fPIE not -fPIC, no?

jrtc27: (although they should fix their toolchain to support PDEs properly... and an XXX: and maybe only do this for aarch64?)

pycheribuild/projects/cross/webkit.py
```diff
@@ -0,0 +1,128 @@
...
+            ENABLE_SAMPLING_PROFILER=False,
+            CHERI_PURE_CAPABILITY=self.crosscompile_target.is_cheri_purecap(),
+            )
+
+        if self.crosscompile_target.is_cheri_purecap():
+            # TODO: we can get this from the pre-processor instead
+            self.add_cmake_options(CHERI_CAPABILITY_SIZE=self.target_info.capability_size_in_bits)
+            self.add_cmake_options(ENABLE_JSHEAP_CHERI_OFFSET_REFS=self.jsheapoffsets)
+
+        # To allow benchmark comparability, we use system malloc and the contiguous arena for
```
jrtc27, line 110:
```suggestion
        # To allow benchmark comparability, we use system malloc and the continuous arena for
```

pycheribuild/projects/cross/qt5.py
arichardson: I'm not sure if the build system produces any, but the flag also applies to shared libraries, so probably -fPIC is safer.

arichardson:
> draft until the webkit repository is made public

We can merge this before the repo is public, we also have other targets that need non-public resources (e.g. spec2006).

PR #174, opened 2021-01-05
Build against rootfs instead of sysroot

arichardson:
This should make it easier to install 3rd-party software
since we no longer have to install to both the sysroot
and the rootfs.

This patch keeps the existing SYSROOT/ROOTFS install
directories: the latter installs to /opt (and any libraries installed
are not part of the compiler search path), the former installs
to /usr/local. It probably makes sense to remove this distinction
in a follow-up change, but for now this should be sufficient to
build against the rootfs instead of the sysroot.

This PR also adds support for building GDB against the FreeBSD
sysroot and can be used to get a debugger in the disk image
without building LLDB or installing packages.

I've built a few things with these changes (e.g. gdb-freebsd-amd64),
but it would be good if someone could verify that I haven't broken
anything before I merge this.
@brettferdosi could you check whether webkit still builds with these
changes (you'll need at least --reconfigure, maybe even --clean for the dependencies).

### DIFF
--- a/githooks/pre-push.sh
+++ b/githooks/pre-push.sh
@@ -48,7 +48,10 @@ check_bad_commit_msg() {
     range=$1
     pattern=$2
     msg=$3
-    bad_commit=$(git rev-list --grep "$pattern" "$range" | grep -v fbb3f6dad35497ed5dcea7e96d0a228b3e383741 | grep -v 4063bcdfe4fcd33b9dd43bf39bd8be3771274ced)
+    # Don't scan commits already in origin/master when creating new branches
+    # for pull requests: At least fbb3f6dad35497ed5dcea7e96d0a228b3e383741 and
+    # 4063bcdfe4fcd33b9dd43bf39bd8be3771274ced have a bad commit message.
+    bad_commit=$(git rev-list --grep "$pattern" "$range" ^origin/master)
     if [ -n "$bad_commit" ]
     then
         echo >&2 "Found $msg commit $bad_commit, not pushing"
--- a/githooks/pre-push.sh
+++ b/githooks/pre-push.sh
@@ -48,7 +48,7 @@ check_bad_commit_msg() {
     range=$1
     pattern=$2
     msg=$3
-    bad_commit=$(git rev-list -n 1 --grep "$pattern" "$range")
+    bad_commit=$(git rev-list --grep "$pattern" "$range" | grep -v fbb3f6dad35497ed5dcea7e96d0a228b3e383741 | grep -v 4063bcdfe4fcd33b9dd43bf39bd8be3771274ced)
     if [ -n "$bad_commit" ]
     then
         echo >&2 "Found $msg commit $bad_commit, not pushing"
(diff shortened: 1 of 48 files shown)


githooks/pre-push.sh
```diff
@@ -48,7 +48,7 @@ check_bad_commit_msg() {
     range=$1
     pattern=$2
     msg=$3
-    bad_commit=$(git rev-list -n 1 --grep "$pattern" "$range")
+    bad_commit=$(git rev-list --grep "$pattern" "$range" | grep -v fbb3f6dad35497ed5dcea7e96d0a228b3e383741 | grep -v 4063bcdfe4fcd33b9dd43bf39bd8be3771274ced)
```
jrtc27, line 51:
```suggestion
    bad_commit=$(git rev-list --grep "$pattern" "$range" ^origin/master)
```
avoids needing to hard-code hashes and will automatically ignore anything that's already been pushed to master (I assume the problem is that, when creating a new branch, range starts at the zero hash and so includes everything; I think I even ran into that once).

pycheribuild/projects/project.py
```diff
@@ -1612,10 +1612,9 @@ def update(self, current_project: "Project", *, src_dir: Path, base_project_sour
 class DefaultInstallDir(Enum):
     DO_NOT_INSTALL = "Should not be installed"
     IN_BUILD_DIRECTORY = "$BUILD_DIR/test-install-prefix"
-    ROOTFS = "The rootfs for this target"
+    ROOTFS = "The rootfs for this target (<rootfs>/opt/<arch>/<program> by default)"
     COMPILER_RESOURCE_DIR = "The compiler resource directory"
-    SYSROOT = "The sysroot for this target"
-    SYSROOT_AND_ROOTFS = "The sysroot for this target and the sysroot"
+    SYSROOT = "The sysroot for this target (<rootfs>/usr/local/<arch> by default)"
```
jrtc27, line 1617: Maybe we should call these OPTBASE+LOCALBASE instead or similar? Having SYSROOT and ROOTFS both exist, with both going into the rootfs by default, just different places, is a bit confusing.

jrtc27: (and I don't think there's a case when you ever want to install something to the sysroot but _not_ the corresponding directory in the rootfs when the two are different; you don't always _need_ to if you're statically linking/it's a header-only library and you don't want to support native compilation, but you still _should_)

brettferdosi: I also think this or a similar renaming is a good idea so that the `path_in_rootfs` attribute doesn't mistakenly get interpreted as relative to the `/opt/...` directory.

brettferdosi (commented):
I successfully built webkit and its dependencies on a clean system.

githooks/pre-push.sh
arichardson: Thanks, that's a better solution. Yes, the problem is that for new branches this check is run for all existing commits.

PR #179, opened 2021-02-23
Allow us to build minimal images for Morello

zxombie:


jrtc27:
Should include aarch64 for completeness

jrtc27:
Or maybe at this point we just support the same set of architectures as CheriBSD rather than repeating the list?...

zxombie:
I tried adding `MORELLO_NO_CHERI`, but got:
```
LookupError: Could not find 'cheribsd' target for CheriBSDMorelloTargetInfo(AARCH64), caller was <pycheribuild.projects.disk_image.BuildMinimalCheriBSDDiskImage_
CheriBSD-Morello-morello-purecap object at 0x8017b6fd0>
```

jrtc27:
That's because it's called AArch64.

arichardson:
> > Or maybe at this point we just support the same set of architectures as CheriBSD rather than repeating the list?...
> 
> @arichardson thoughts?

Yes that would make sense. I think the only reason that the lists are different is that cheribsdbox was only supported for MIPS initially.

arichardson (approved):
This looks fine to me, we can merge the lists in a follow-up commit.

PR #180, opened 2021-02-23
Fix sysroot archive creation

arichardson:
We will also need to update the cheribsd dev Jenkinsfile.


jrtc27:
Can we drop the second commit now? :)

jrtc27:
(well, keep the XZ part, drop the ./ part and rewrite the message)

arichardson:
Should be fixed now.

jrtc27 (approved):

PR #181, opened 2021-02-26
spec2006-new: force u+w perms after unpacking

nwf:
Is there a better way to do this?


arichardson (approved):

arichardson:
I assume you need this to apply patches?

nwf:
It actually trips up the `rm` from a `-c` build, even without wanting to apply patches.

PR #183, opened 2021-03-04
Multicore vcu118

PeterRugg:


vcu118-run.py
```diff
@@ -92,7 +92,8 @@
 exit 0
 """
 
-OPENOCD_SCIPRT = b"""
+def generate_openocd_script (n_cores: int):
```
arichardson, line 95:
```suggestion
def generate_openocd_script(n_cores: int):
```

vcu118-run.py
```diff
@@ -371,6 +392,7 @@ def main():
     parser.add_argument("--gdb", default=shutil.which("gdb") or "gdb", help="Path to GDB binary", type=Path)
     parser.add_argument("--openocd", default=shutil.which("openocd") or "openocd", help="Path to openocd binary",
                         type=abspath_arg)
+    parser.add_argument("--n-cores", type=int, default=1, help="Number of harts on bitstream")
```
arichardson, line 395:
```suggestion
    parser.add_argument("--cores", type=int, default=1, help="Number of harts on bitstream")
```
The -n prefix seems a bit odd, I'd go for `--cores` or `--num-cores`.

arichardson (approved):
Minor suggestions, otherwise LGTM. I haven't reviewed the openocd script changes since I don't know enough about it, but assuming single-core still works this should be fine.

vcu118-run.py
```diff
@@ -109,9 +110,22 @@
...
+        openocd_script += "target create $_TARGETNAME_{:d} riscv -chain-position $_CHIPNAME.cpu -coreid {:d}".format(core, core)
+        if (core == 0):
+            openocd_script += " -rtos hwthread"
+        openocd_script += "\n"
+
+    if (n_cores > 0):
+        openocd_script += "\ntarget smp"
+        for core in range(n_cores):
+            openocd_script += " $_TARGETNAME_{:d}".format(core)
+        openocd_script += "\n"
```
jrtc27, line 126:
```suggestion
    for core in range(n_cores):
        openocd_script += "\nset _TARGETNAME_{0:d} $_CHIPNAME.cpu{0:d}".format(core)
        openocd_script += "target create $_TARGETNAME_{0:d} riscv -chain-position $_CHIPNAME.cpu -coreid {0:d}".format(core)
        if core == 0:
            openocd_script += " -rtos hwthread"
        openocd_script += "\n"

    if n_cores > 0:
        openocd_script += "\ntarget smp"
        for core in range(n_cores):
            openocd_script += " $_TARGETNAME_{:d}".format(core)
        openocd_script += "\n"
```

vcu118-run.py
```diff
@@ -286,6 +301,11 @@ def load_and_start_kernel(*, gdb_cmd: Path, openocd_cmd: Path, bios_image: Path,
     args += ["-ex", "set style enabled off"]  # disable colours since they break the matcher strings
     args += ["-ex", "monitor reset init"]  # reset and go back to boot room
     args += ["-ex", "si 5"]  # we need to run the first few instructions to get a valid DTB
+    for core in range(1, n_cores):
+        args += ["-ex", "thread {:d}".format(core + 1)] # switch to thread (core + 1) (GDB counts from 1)
+        args += ["-ex", "si 5"] # execute bootrom on every core
+        args += ["-ex", "set $pc=0xc0000000"] # set every core to start of cached memory XXX Assumes this the start of bios
```
jrtc27, line 307: If you want to do this properly, you could do:
```
set $entry_point=$pc
thread 2
set $pc=$entry_point
thread 3
set $pc=$entry_point
...
```
(as `-ex` commands) after the `load` below instead, I believe.

jrtc27 (commented):
Calling the variable `num_cores` everywhere would be better style for this I think?

PeterRugg:
Thanks both for your prompt comments! I've implemented them all, so hopefully good to merge?

arichardson (approved):

vcu118-run.py
```diff
@@ -333,13 +353,17 @@ def load_and_start_kernel(*, gdb_cmd: Path, openocd_cmd: Path, bios_image: Path,
     load_end_time = datetime.datetime.utcnow()
     success("Finished loading bootloader image in ", load_end_time - load_start_time)
     gdb_finish_time = load_end_time
+    for core in range(1, num_cores):
+        gdb.expect_exact(["0x0000000044000000"])
+    if (num_cores > 1):
```
jrtc27, line 358:
```suggestion
    if num_cores > 1:
```

vcu118-run.py
```diff
@@ -301,6 +315,12 @@ def load_and_start_kernel(*, gdb_cmd: Path, openocd_cmd: Path, bios_image: Path,
         args += ["-ex", "symbol-file " + shlex.quote(str(kernel_debug_file.absolute()))]
         args += ["-ex", "load " + shlex.quote(str(kernel_image.absolute()))]
     args += ["-ex", "load " + shlex.quote(str(Path(bios_image).absolute()))]
+    args += ["-ex", "set $entry_point = $pc"] # Record the entry point to the bios
+    for core in range(1, num_cores):
+        args += ["-ex", "thread {:d}".format(core + 1)] # switch to thread (core + 1) (GDB counts from 1)
+        args += ["-ex", "si 5"] # execute bootrom on every core
+        args += ["-ex", "set $pc=$entry_point"] # set every core to the start of the bios
```
jrtc27, line 322:
```suggestion
        args += ["-ex", "si 5"] # execute bootrom on every other core
        args += ["-ex", "set $pc=$entry_point"] # set every other core to the start of the bios
```

vcu118-run.py
```diff
@@ -301,6 +315,12 @@ def load_and_start_kernel(*, gdb_cmd: Path, openocd_cmd: Path, bios_image: Path,
...
+    args += ["-ex", "thread 1"] # switch back to core 0
```
jrtc27, line 323: Could guard this whole thing with `if num_cores > 1:`, not sure if I like that idea or not

jrtc27:
Oh, actually, I hadn't noticed the over-indented success line. I think it makes sense to put the other loop underneath the if then too, just not when it was (incorrectly) _only_ a for loop.

PeterRugg:
> Oh, actually, I hadn't noticed the over-indented success line. I think it makes sense to put the other loop underneath the if then too, just not when it was (incorrectly) _only_ a for loop.

Done :)

jrtc27:
Ok and please fold commit 3 into commit 1, then I think this is ready to merge

PR #184, opened 2021-03-05
Make the minimal AArch64/Morello images bootable

zxombie:
I'm working on a change to boot the minimal arm64 images. Fix two issues I found while working on it:
 * Replace boot1.efi with one of the loader.efi variants. the former loads the latter, so follow FreeBSD and use the latter directly.
 * Include the kernel in the minimal image on arm64 so loader.efi can find it.

### DIFF
--- a/pycheribuild/projects/disk_image.py
+++ b/pycheribuild/projects/disk_image.py
@@ -875,6 +875,9 @@ class BuildMinimalCheriBSDDiskImage(_BuildDiskImageBase):
                         include_local_file("files/minimal-image/etc.files")]
         if self._have_cplusplus_support(["lib", "usr/lib"]):
             files_to_add.append(include_local_file("files/minimal-image/need-cplusplus.files"))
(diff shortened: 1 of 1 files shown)


jrtc27:
Kernels don't belong in minimal images, those are meant to be for embedding in MFS_ROOT kernels.

pycheribuild/projects/disk_image.py
```diff
@@ -178,6 +178,7 @@ def __init__(self, config, source_class: "typing.Type[BuildFreeBSD]"):
         self.hostname = os.path.expandvars(self.hostname)  # Expand env vars in hostname to allow $CHERI_BITS
         # MIPS needs big-endian disk images
         self.big_endian = self.compiling_for_mips(include_purecap=True)
+        self.efi_loader = "loader_lua.efi" # Which loader to use when booting via UEFI
```
jrtc27, line 181: Just use loader.efi and let the build system decide what the default is?

PR #185, opened 2021-03-05
Deconflate minimal and MFS_ROOT images

jrtc27:
Currently we regard minimal images as being the images embedded MFS_ROOT kernels, but they have another use case as just being a much smaller disk image. This separates things out everywhere so both exist, with minimal images now being bootable, and MFS_ROOT images only being a rootfs.

### DIFF
--- a/pycheribuild/projects/cross/cheribsd.py
+++ b/pycheribuild/projects/cross/cheribsd.py
@@ -1353,8 +1353,8 @@ class BuildCheriBSDFett(BuildCHERIBSD):
 # FIXME: this should inherit from BuildCheriBSD to avoid subtle problems
 class BuildCheriBsdMfsKernel(SimpleProject):
     project_name = "cheribsd-mfs-root-kernel"
-    dependencies = ["disk-image-minimal"]
-    # TODO: also support building a non-CHERI kernel... But that needs a plain MIPS disk-image-minimal first...
+    dependencies = ["disk-image-mfs-root"]
+    # TODO: also support building a non-CHERI kernel... But that needs a plain MIPS disk-image-mfs-root first...
     _always_add_suffixed_targets = True
 
     @classproperty
@@ -1369,12 +1369,12 @@ class BuildCheriBsdMfsKernel(SimpleProject):
 
     def __init__(self, config: CheriConfig):
         super().__init__(config)
-        from ..disk_image import BuildMinimalCheriBSDDiskImage
-        self.minimal_image_instance = BuildMinimalCheriBSDDiskImage.get_instance(self)
+        from ..disk_image import BuildMfsRootCheriBSDDiskImage
+        self.mfs_root_image_instance = BuildMfsRootCheriBSDDiskImage.get_instance(self)
         # Re-use the same build directory as the CheriBSD target that was used for the disk image
         # This ensure that the kernel build tools can be found in the build directory
-        self.image = self.minimal_image_instance.disk_image_path
-        self.build_cheribsd_instance = self.minimal_image_instance.cheribsd_class.get_instance(self)
+        self.image = self.mfs_root_image_instance.disk_image_path
+        self.build_cheribsd_instance = self.mfs_root_image_instance.cheribsd_class.get_instance(self)
 
     def process(self):
         default_kernconf = self._get_kernconf_to_build(self.build_cheribsd_instance)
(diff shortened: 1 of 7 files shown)


jrtc27:
This should supersede #184

pycheribuild/projects/disk_image.py
```diff
@@ -458,13 +476,33 @@ def add_all_files_in_dir(self, root_dir: Path):
...
+    @property
+    def include_efi_partition(self):
+        # TODO: Make this unconditional once all branches support EFI
+        if self.crosscompile_target.is_riscv(include_purecap=True):
+            return (self.rootfs_dir / "boot/loader.efi").exists()
+        return False
+
+    @property
+    def include_swap_partition(self):
+        return self.crosscompile_target.is_riscv(include_purecap=True)
```
nwf, line 495: Even in the minimal images?  And why only RISC-V?

pycheribuild/projects/disk_image.py
```diff
@@ -458,13 +476,33 @@ def add_all_files_in_dir(self, root_dir: Path):
...
+
+    @property
+    def rootfs_only(self):
```
nwf, line 498: This function also seems to have a strangely-shaped predicate.  At least some docs would probably help, even if it has to stay strangely shaped and can't be pushed into child classes like `BuildMinimalCheriBSDDiskImage`?

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -1353,8 +1353,8 @@ def setup_config_options(cls, **kwargs):
 # FIXME: this should inherit from BuildCheriBSD to avoid subtle problems
 class BuildCheriBsdMfsKernel(SimpleProject):
     project_name = "cheribsd-mfs-root-kernel"
-    dependencies = ["disk-image-minimal"]
-    # TODO: also support building a non-CHERI kernel... But that needs a plain MIPS disk-image-minimal first...
+    dependencies = ["disk-image-mfs-root"]
+    # TODO: also support building a non-CHERI kernel... But that needs a plain MIPS disk-image-mfs-root first...
```
arichardson, line 1357: I think this has long been resolved? Since you're changing this line might as well remove it?

nwf (commented):
A good plan, I think.  Two minor nits.

pycheribuild/projects/disk_image.py
```diff
@@ -483,6 +483,28 @@ def run_mkimg(self, cmd: list, **kwargs):
             self.fatal("Missing mkimg command! Should be found in FreeBSD build dir (or set $MKIMG_CMD)")
         self.run_cmd([self.mkimg_cmd] + cmd, **kwargs)
 
+    @property
+    def include_efi_partition(self):
+        if self.crosscompile_target.is_mips(include_purecap=True):
+            return False
+        # TODO: Make this unconditional once all branches support EFI
+        if self.crosscompile_target.is_riscv(include_purecap=True):
```
arichardson, line 491: Why is this only for RISC-V? make_aarch64_disk_image was using an EFI partition too?

arichardson (commented):
Much more consistent now, just a few questions on the last commit.

pycheribuild/projects/disk_image.py
jrtc27: I screwed up the original patch, should be fixed already (defaults to True, with the MIPS early bail-out)

pycheribuild/projects/disk_image.py
jrtc27: Why only RISC-V: because that's what was already being done (added for FETT it not the nicest way)

Why minimal images: oops yes good point

pycheribuild/projects/disk_image.py
jrtc27: _sigh_ the RISC-V predicate is meant to be negated

pycheribuild/projects/cross/cheribsd.py
jrtc27: Oh, hm, I had interpreted that as meaning FreeBSD not CheriBSD-without-CHERI, but yes, this can go

pycheribuild/projects/disk_image.py
arichardson: Would it make sense to just add swap for all images? It makes the raw images a bit bigger but the compressed size should stay the same?

arichardson (approved):

pycheribuild/projects/disk_image.py
zxombie: If it's added to all images I'd like a way to disable that as I have image size constraints and swap doesn't help when the disk image is loaded into memory before booting.

jrtc27: I'm not convinced it should be there by default for RISC-V to be honest, IMO it should've only been there for FETT disk images.

nwf (approved):

PR #187, opened 2021-03-17
Use jsonc to support comment

SchrodingerZhu:
Just to remove those ugly syntax errors


arichardson:
Nice, I did not know about that syntax highlightinng feature.

jrtc27:
Do those comments get parsed correctly by the JSON library uses by cheribuild?

arichardson:
Yes I have code to strip both lines starting with `#` and `//`

PR #190, opened 2021-04-09
More morello FVP options for plugins.

qwattash:


pycheribuild/projects/run_fvp.py
```diff
@@ -608,6 +618,49 @@ def add_hostbridge_params(*params):
                 # "-C", "css.trustedBootROMloader.fname=" + str(trusted_fw),
                 "-C", "css.pl011_uart_ap.unbuffered_output=1",
             ]
+
+            if self.fvp_cadiserver:
+                fvp_args += ["-S"]
```
jrtc27, line 623: How do you connect to it? I guess the client requires Linux so forwarding that for Docker doesn't really make much sense?

pycheribuild/projects/run_fvp.py
```diff
@@ -608,6 +618,49 @@ def add_hostbridge_params(*params):
...
+
+            if self.fvp_gdbserver:
+                fvp_args += [
+                    "--allow-debug-plugin",
+                    "--plugin", self.fvp_project.plugin_dir / "GDBRemoteConnection.so",
+                    "-C", "REMOTE_CONNECTION.GDBRemoteConnection.listen_address=127.0.0.1",
+                    "-C", "REMOTE_CONNECTION.GDBRemoteConnection.port={}".format(self.fvp_gdbport)]
```
jrtc27, line 630: Needs forwarding for Docker like SSH

pycheribuild/projects/run_fvp.py
qwattash: I'm still unsure how the whole CADI debugging stack works end-to-end but there is no option to specify the port, I only added it to test whether I could get a connection locally, but I guess that for Docker does not make much sense. Not sure if we want to ditch it or disable it?

pycheribuild/projects/run_fvp.py
```diff
@@ -475,6 +484,14 @@ def setup_config_options(cls, **kwargs):
         cls.smp = cls.add_bool_option("smp", help="Simulate multiple CPU cores in the FVP", default=True)
         cls.force_headless = cls.add_bool_option("force-headless", default=False,
                                                  help="Force headless use of the FVP")
+        cls.fvp_gdbserver = cls.add_bool_option("gdbserver", default=False,
+                                                help="Enable FVP GDB debugging plugin and wait for GDB to connect")
+        cls.fvp_gdbport = cls.add_config_option("gdb-port", default=cls.default_gdb_port(), kind=int)
```
arichardson, line 489: Should we default to the same port as QEMU for consistency? I.e. 1234 by default but use `find_free_port()` in the same way as run_qemu.py?

qwattash: Good point, fixed.

pycheribuild/projects/run_fvp.py
```diff
@@ -475,6 +481,14 @@ def setup_config_options(cls, **kwargs):
         cls.smp = cls.add_bool_option("smp", help="Simulate multiple CPU cores in the FVP", default=True)
         cls.force_headless = cls.add_bool_option("force-headless", default=False,
                                                  help="Force headless use of the FVP")
+        cls.fvp_gdbserver = cls.add_bool_option("gdbserver", default=False,
```
arichardson, line 484: Maybe this should just use the global `--wait-for-debugger` option (or use that as the default value)?

arichardson (approved):
Otherwise looks good to me.

pycheribuild/projects/run_fvp.py
qwattash: Dropped

PR #192, opened 2021-04-19
Apache

bsdjhb:
This isn't quite ready as I will need to update the repository URLs and branches for apr and apache.

### DIFF
--- a/pycheribuild/projects/cross/apache.py
+++ b/pycheribuild/projects/cross/apache.py
@@ -1,0 +1,141 @@
+#
+# Copyright (c) 2021 SRI International
+# All rights reserved.
+#
+# This software was developed by SRI International and the University of
+# Cambridge Computer Laboratory (Department of Computer Science and
+# Technology) under DARPA contract HR0011-18-C-0016 ("ECATS"), as part of the
+# DARPA SSITH research programme.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+# OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+# OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+# SUCH DAMAGE.
+#
+from .crosscompileproject import (CrossCompileAutotoolsProject, DefaultInstallDir, GitRepository, SubversionRepository)
+from .expat import BuildExpat
+
+
+class BuildPcre(CrossCompileAutotoolsProject):
+    target = "pcre"
+    project_name = "pcre"
+
+    repository = SubversionRepository("svn://vcs.pcre.org/pcre",
+                                      default_branch="code/trunk")
+
+    native_install_dir = DefaultInstallDir.BOOTSTRAP_TOOLS
+    cross_install_dir = DefaultInstallDir.ROOTFS_LOCALBASE
+
+    def configure(self, **kwargs):
+        self.run_cmd("autoreconf", "-i", cwd=self.source_dir)
+        super().configure(**kwargs)
+
+
+class BuildApr(CrossCompileAutotoolsProject):
+    target = "apr"
+    project_name = "apr"
+
+    repository = GitRepository("https://github.com/CTSRD-CHERI/apr.git",
+                               default_branch="cheri")
+
+    dependencies = ["libexpat"]
+
+    native_install_dir = DefaultInstallDir.BOOTSTRAP_TOOLS
+    cross_install_dir = DefaultInstallDir.ROOTFS_LOCALBASE
+
+    def setup(self):
+        super().setup()
+        self.configure_args.extend([
+            "--enable-threads",
+            "--enable-posix-shm",
+            "--with-devrandom",
+            "--with-expat=" + str(BuildExpat.get_install_dir(self))
+            ])
+        if self.build_type.is_debug:
+            self.configure_args.append("--enable-debug")
+
+        # configure is too broken to add the -L path needed
+        # for -lexpat to the generated Makefile though it
+        # does use it while testing if libexpat works
+        self.LDFLAGS.append("-L" + str(self.install_dir / "lib"))
+
+        if not self.compiling_for_host():
+            # Can't determine these when cross-compiling
+            self.configure_environment.update(ac_cv_file__dev_zero="yes",
+                                              ac_cv_mmap__dev_zero="yes",
+                                              # XXX: This might be yes on Linux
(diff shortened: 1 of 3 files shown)


pycheribuild/projects/cross/apache.py
```diff
@@ -0,0 +1,144 @@
...
+                                              # XXX: This might be yes on Linux
+                                              apr_cv_tcp_nodelay_with_cork="no",
+                                              )
+
+        self.run_cmd("./buildconf", cwd=self.source_dir)
+        super().configure(**kwargs)
+
+        # gen_test_char rules in server/ assume a native build
+        if not self.compiling_for_host():
+            self.run_cmd(str(self.host_CC), "-DCROSS_COMPILE", "-c",
```
jrtc27, line 137: CFLAGS too? Both for things like optimisation, but also because host_CC may be a clang that doesn't target the host by default.

arichardson: I think we generally assume that `host_CC` can build files for the current system without any target/abi/etc. flags.

bsdjhb: Hmm, do we have a suitable host CFLAGS?

pycheribuild/projects/project.py
```diff
@@ -1645,6 +1645,44 @@ def update(self, current_project: "Project", *, src_dir: Path, base_project_sour
...
+    def ensure_cloned(self, current_project: "Project", src_dir: Path, **kwargs):
+        if (src_dir / ".svn").is_dir():
+            return
+
+        if current_project.config.skip_clone:
+            current_project.fatal("Sources for", str(src_dir), " missing!")
+            return
+
+        assert isinstance(self.url, str), self.url
+        assert not self.url.startswith("<"), "Invalid URL " + self.url
```
arichardson, line 1662: I couldn't remember why I added that assert in 210264525a61c9e75b13a0d7d26782c9c021113e. I believe the problem was that I had accidentally used `str(<instance of GitRepository>)`. And the default python str() is `<$CLASS object at 0x....>`. Not sure this assert is required.

pycheribuild/projects/cross/apache.py
arichardson: I don't think so. I guess this is just a simple build tool? In that case shouldn't it be fine to just use the host compiler defaults?

pycheribuild/projects/project.py
bsdjhb: I had guessed that was your reason which is why I kept it when copying from GitRepository.  I don't mind dropping it though if you'd rather me do so.

bsdjhb:
Updated URLs and branches for local repos, and fixed the flake8 warnings from the last run.

pycheribuild/projects/cross/apache.py
bsdjhb: It just uses <ctype.h> to print out tables of 'isprint()', etc.  Technically using the host version of the tables probably isn't correct, but is probably sufficient for cheribuild.

pycheribuild/projects/cross/apache.py
```diff
@@ -0,0 +1,148 @@
...
+                               default_branch="cheri")
+
+    dependencies = ["libexpat"]
+
+    native_install_dir = DefaultInstallDir.BOOTSTRAP_TOOLS
+    cross_install_dir = DefaultInstallDir.ROOTFS_LOCALBASE
+
+    def configure(self, **kwargs):
+        self.configure_args.extend([
+            "--enable-debug",
```
bsdjhb, line 63: I added this (and a similar one for apache) to get debug symbols.

It doesn't look like there's a generic "enable debug mode" knob in cheribuild projects on which I would perhaps make this flag conditional?

bsdjhb: Ah, used 'build_type.is-debug'.

pycheribuild/projects/cross/apache.py
```diff
@@ -0,0 +1,150 @@
...
+
+    native_install_dir = DefaultInstallDir.BOOTSTRAP_TOOLS
+    cross_install_dir = DefaultInstallDir.ROOTFS_LOCALBASE
+
+    def configure(self, **kwargs):
+        self.configure_args.extend([
+            "--enable-threads",
+            "--enable-posix-shm",
+            "--with-devrandom",
+            "--with-expat=" + str(self.install_dir)
```
arichardson, line 66:
```suggestion
            "--with-expat=" + str(BuildExpat.get_install_dir())
```

pycheribuild/projects/cross/apache.py
```diff
@@ -0,0 +1,150 @@
...
+
+    repository = GitRepository("https://github.com/CTSRD-CHERI/apr.git",
+                               default_branch="cheri")
+
+    dependencies = ["libexpat"]
+
+    native_install_dir = DefaultInstallDir.BOOTSTRAP_TOOLS
+    cross_install_dir = DefaultInstallDir.ROOTFS_LOCALBASE
+
+    def configure(self, **kwargs):
```
arichardson, line 61:
```suggestion
    def setup(self):
        super().setup()
```

arichardson: I've generally been trying to move changes to instance variables to `setup()` and not modify `self.configure_environment`, etc. inside ` configure()`

pycheribuild/projects/cross/apache.py
```diff
@@ -0,0 +1,150 @@
...
+                                              # XXX: This might be yes on Linux
+                                              ac_cv_func_setpgrp_void="no",
+                                              ac_cv_struct_rlimit="yes",
+                                              ac_cv_func_sem_open="yes",
+                                              apr_cv_process_shared_works="yes",
+                                              apr_cv_mutex_robust_shared="yes",
+                                              # XXX: This might be yes on Linux
+                                              apr_cv_tcp_nodelay_with_cork="no",
+                                              )
+
```
arichardson, line 89:
```suggestion

    def configure(self, **kwargs):
```

pycheribuild/projects/cross/apache.py
```diff
@@ -0,0 +1,150 @@
...
```
arichardson, line 139: Maybe add the setup()/configure() split here as well?

pycheribuild/projects/project.py
arichardson: Seems fine to keep it

pycheribuild/projects/cross/apache.py
```diff
@@ -0,0 +1,150 @@
...
+    native_install_dir = DefaultInstallDir.IN_BUILD_DIRECTORY
+    cross_install_dir = DefaultInstallDir.ROOTFS_LOCALBASE
+
+    def configure(self, **kwargs):
+        self.configure_args.extend([
+            "--enable-layout=FreeBSD",
+            "--enable-http",
+            "--enable-mod-ssl",
+            "--with-expat=" + str(self.install_dir),
```
arichardson, line 111:
```suggestion
            "--with-expat=" + str(BuildExpat.get_install_dir()),
```

pycheribuild/projects/cross/apache.py
```diff
@@ -0,0 +1,150 @@
...
+# ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+# OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+# OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+# SUCH DAMAGE.
+#
+from .crosscompileproject import (CrossCompileAutotoolsProject, DefaultInstallDir, GitRepository, SubversionRepository)
```
arichardson, line 31:
```suggestion
from .crosscompileproject import (CrossCompileAutotoolsProject, DefaultInstallDir, GitRepository, SubversionRepository)
from .expat import BuildExpat
```
I'll push a change shortly to move expat to a separate .py file.

arichardson (approved):
Minor suggestions, otherwise looks good.

pycheribuild/projects/cross/apache.py
bsdjhb: Ah, ok, thanks.  That is why I hadn't used it yet.

pycheribuild/projects/cross/apache.py
```diff
@@ -0,0 +1,150 @@
...
+
+        # The cross-compile always assumes this is true (so one
+        # can't normally cross-compile Apache), and it's a fatal
+        # error, so just lie always.
+        self.configure_environment.update(ap_cv_void_ptr_lt_long="no")
+
+        # mod_ssl doesn't use -L from --with-expat
+        self.LDFLAGS.append("-L" + str(self.install_dir / "lib"))
+
+        if not self.compiling_for_host() and False:
```
bsdjhb, line 129: Oops, this was copy/pasta from Apr in case I needed it.  I will axe it.

pycheribuild/projects/cross/apache.py
bsdjhb: Will do.

arichardson (approved):

PR #196, opened 2021-05-12
CheriBsdMfsKernel: [no-]build-bench-kernels option

nwf:
Sometimes the goal is to build one particular kernel configuration.


jrtc27:
Maybe default=True though? Especially -NODEBUG kernels are pretty common to want.

qwattash:
This is now replaced by #199

PR #199, opened 2021-05-18
Try to cleanup the handling of cheribsd kernel configuration files

qwattash:
This attempts to have a separate class to fetch (and possibly generate) kernel configuration variants to build and run in cheribuild targets.

### DIFF
--- a/pycheribuild/config/compilation_targets.py
+++ b/pycheribuild/config/compilation_targets.py
@@ -277,6 +277,23 @@ class FreeBSDTargetInfo(_ClangBasedTargetInfo):
         return target.cpu_architecture.value + common_suffix
 
     @property
+    def freebsd_target_cputype(self):
+        """
+        Return the name of the target CPU type, which is also the name of
+        the machine-dependent code directory in the source tree.
+        (e.g. <arch> in sys/<arch>)
+        """
+        mapping = {
+            CPUArchitecture.AARCH64: "arm64",
+            CPUArchitecture.ARM32: "arm",
+            CPUArchitecture.I386: "i386",
+            CPUArchitecture.MIPS64: "mips",
+            CPUArchitecture.RISCV64: "riscv",
+            CPUArchitecture.X86_64: "amd64",
+        }
+        return mapping[self.target.cpu_architecture]
+
+    @property
     def freebsd_target_arch(self):
         mapping = {
             CPUArchitecture.AARCH64: "aarch64",
--- a/pycheribuild/config/compilation_targets.py
+++ b/pycheribuild/config/compilation_targets.py
@@ -329,13 +329,17 @@ class CheriBSDTargetInfo(FreeBSDTargetInfo):
     def is_cheribsd(cls):
         return True
 
-    def _get_mfs_root_kernel(self, use_benchmark_kernel: bool) -> Path:
+    def _get_mfs_root_kernel(self, platform, use_benchmark_kernel: bool) -> Path:
         assert self.is_cheribsd(), "Other cases not handled yet"
         from ..projects.cross.cheribsd import BuildCheriBsdMfsKernel
-        if use_benchmark_kernel:
-            return BuildCheriBsdMfsKernel.get_installed_benchmark_kernel_path(self.project)
-        else:
-            return BuildCheriBsdMfsKernel.get_installed_kernel_path(self.project)
+        xtarget = self._get_mfs_kernel_xtarget()
+        if xtarget not in BuildCheriBsdMfsKernel.supported_architectures:
+            self.project.fatal("No MFS kernel for target", xtarget)
+            return None
+        mfs_kernel = BuildCheriBsdMfsKernel.get_instance_for_cross_target(
+            xtarget, self.config, caller=self.project)
+        kernconf = mfs_kernel.default_kernel_config(platform, benchmark=use_benchmark_kernel)
+        return mfs_kernel.get_kernel_install_path(kernconf)
 
     def _get_mfs_kernel_xtarget(self):
         kernel_xtarget = self.target
@@ -385,6 +389,7 @@ class CheriBSDTargetInfo(FreeBSDTargetInfo):
                 from ..projects.disk_image import BuildMinimalCheriBSDDiskImage
                 disk_image_path = BuildMinimalCheriBSDDiskImage.get_instance(self.project).disk_image_path
         elif kernel_path is None and "--kernel" not in self.config.test_extra_args:
+            from ..projects.cross.cheribsd import ConfigPlatform
             # Use the benchmark kernel by default if the parameter is set and the user didn't pass
             # --no-use-minimal-benchmark-kernel on the command line or in the config JSON
             use_benchmark_kernel_value = self.config.use_minimal_benchmark_kernel  # Load the value first to ensure
@@ -393,8 +398,8 @@ class CheriBSDTargetInfo(FreeBSDTargetInfo):
             assert isinstance(use_benchmark_config_option, ConfigOptionBase)
             want_benchmark_kernel = use_benchmark_kernel_value or (
                     use_benchmark_kernel_by_default and use_benchmark_config_option.is_default_value)
-            kernel_path = self._get_mfs_root_kernel(use_benchmark_kernel=want_benchmark_kernel)
-            if not kernel_path.exists() and is_jenkins_build():
+            kernel_path = self._get_mfs_root_kernel(ConfigPlatform.QEMU, want_benchmark_kernel)
+            if (kernel_path is None or not kernel_path.exists()) and is_jenkins_build():
                 jenkins_kernel_path = self.config.cheribsd_image_root / "kernel.xz"
                 if jenkins_kernel_path.exists():
                     kernel_path = jenkins_kernel_path
@@ -517,10 +522,10 @@ class CheriBSDTargetInfo(FreeBSDTargetInfo):
         if self.config.tests_interact:
             runbench_args.append("--interact")
 
-        from ..projects.cross.cheribsd import BuildCheriBsdMfsKernel
+        from ..projects.cross.cheribsd import BuildCheriBsdMfsKernel, ConfigPlatform
         if self.config.benchmark_with_qemu:
             # When benchmarking with QEMU we always spawn a new instance
-            kernel_image = self._get_mfs_root_kernel(use_benchmark_kernel=not self.config.benchmark_with_debug_kernel)
+            kernel_image = self._get_mfs_root_kernel(ConfigPlatform.QEMU, not self.config.benchmark_with_debug_kernel)
             basic_args.append("--kernel-img=" + str(kernel_image))
         elif self.config.benchmark_clean_boot:
             # use a bitfile from jenkins. TODO: add option for overriding
@@ -528,11 +533,9 @@ class CheriBSDTargetInfo(FreeBSDTargetInfo):
             basic_args.append("--jenkins-bitfile=cheri128")
             mfs_kernel = BuildCheriBsdMfsKernel.get_instance_for_cross_target(self._get_mfs_kernel_xtarget(),
                                                                               self.config, caller=self.project)
-            kernel_configs = mfs_kernel.get_kernel_configs(kernABI=mfs_kernel.get_default_kernel_abi(),
-                                                           platform=ConfigPlatform.BERI,
-                                                           benchmark=(not self.config.benchmark_with_debug_kernel))
-            assert len(kernel_configs), "No matching kernel configurations"
-            kernel_image = mfs_kernel.get_kernel_install_path(kernel_configs[0].kernconf)
+            kernel_config = mfs_kernel.default_kernel_config(ConfigPlatform.BERI,
+                                                             benchmark=not self.config.benchmark_with_debug_kernel)
+            kernel_image = mfs_kernel.get_kernel_install_path(kernel_config)
             basic_args.append("--kernel-img=" + str(kernel_image))
         else:
             runbench_args.append("--skip-boot")
--- a/pycheribuild/config/compilation_targets.py
+++ b/pycheribuild/config/compilation_targets.py
@@ -528,12 +528,12 @@ class CheriBSDTargetInfo(FreeBSDTargetInfo):
             basic_args.append("--jenkins-bitfile=cheri128")
             mfs_kernel = BuildCheriBsdMfsKernel.get_instance_for_cross_target(self._get_mfs_kernel_xtarget(),
                                                                               self.config, caller=self.project)
-            if self.config.benchmark_with_debug_kernel:
-                kernel_config = mfs_kernel.fpga_kernconf
-            else:
-                kernel_config = mfs_kernel.fpga_kernconf + "_BENCHMARK"
-            basic_args.append(
-                "--kernel-img=" + str(mfs_kernel.installed_kernel_for_config(self.project, kernel_config)))
+            kernel_configs = mfs_kernel.get_kernel_configs(kernABI=mfs_kernel.get_default_kernel_abi(),
+                                                           platform=ConfigPlatform.BERI,
+                                                           benchmark=(not self.config.benchmark_with_debug_kernel))
+            assert len(kernel_configs), "No matching kernel configurations"
+            kernel_image = mfs_kernel.get_kernel_install_path(kernel_configs[0].kernconf)
+            basic_args.append("--kernel-img=" + str(kernel_image))
         else:
             runbench_args.append("--skip-boot")
         if benchmark_script:
@@ -1016,6 +1016,7 @@ class CompilationTargets(BasicCompilationTargets):
     ALL_CHERIBSD_NON_CHERI_TARGETS = [CHERIBSD_MIPS_NO_CHERI, CHERIBSD_RISCV_NO_CHERI, CHERIBSD_AARCH64,
                                       CHERIBSD_X86_64]
     ALL_FREEBSD_AND_CHERIBSD_TARGETS = ALL_CHERIBSD_TARGETS + ALL_SUPPORTED_FREEBSD_TARGETS
+    ALL_CHERIBSD_CHERI_TARGETS = (set(ALL_CHERIBSD_TARGETS) - set(ALL_CHERIBSD_NON_CHERI_TARGETS))
 
     # Same as above, but the default is purecap RISC-V
     FETT_DEFAULT_ARCHITECTURE = CHERIBSD_RISCV_PURECAP
--- a/pycheribuild/projects/cross/cheribsd.py
+++ b/pycheribuild/projects/cross/cheribsd.py
@@ -27,6 +27,7 @@
 # OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 # SUCH DAMAGE.
 #
+import itertools
 import os
 import shutil
 import subprocess
@@ -307,7 +308,7 @@ class AArch64KernelConfigFactory(KernelConfigFactory):
         return configs
 
 
-def filter_kernel_configs(configs: list, **filter_kwargs):
+def filter_kernel_configs(configs: list, **filter_kwargs) -> "typing.Sequence[CheriBSDConfig]":
     """
     Helper function to filter kernel configuration lists.
     Keyword filter arguments are mapped to CheriBSDConfig properties.
@@ -1606,6 +1607,11 @@ class BuildCHERIBSD(BuildFreeBSD):
             only_add_for_targets=(CompilationTargets.ALL_CHERIBSD_MIPS_AND_RISCV_TARGETS +
                                   CompilationTargets.ALL_CHERIBSD_MORELLO_TARGETS),
             help="Also build kernels with non-default ABI (purecap or hybrid)")
+
+        cls.build_bench_kernels = cls.add_bool_option("build-bench-kernels", show_help=True,
+                                                      _allow_unknown_targets=True,
+                                                      help="Also build benchmark kernels")
+
         cls.caprevoke_kernel = cls.add_bool_option("caprevoke-kernel", show_help=True, _allow_unknown_targets=True,
                                                    only_add_for_targets=[CompilationTargets.CHERIBSD_MIPS_PURECAP,
                                                                          CompilationTargets.CHERIBSD_MIPS_HYBRID,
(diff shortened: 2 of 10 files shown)


pycheribuild/projects/cross/cheribsd.py
```diff
@@ -67,6 +71,277 @@ def _clear_dangerous_make_env_vars():
...
+    Maintain lists of kernel configurations for each target we support.
+    The following requirements need to be enforced in order to avoid missing
+    default configurations:
+    - For each supported platform there should be at least a configuration marked as default.
+      There may be multiple defaults with debug/mfsroot/benchmark flags.
+    - The platforms are used to select configurations available to run jobs.
+    - Default configurations must select an unique (platform, kernel, flags) set, non-default
+      configurations may select multiple kernels.
+    """
+    # XXX-AM: For now hardcode the lists, ideally these may be generated
```
jrtc27, line 157: This is really nasty, this shouldn't be too hard to generate

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -67,6 +71,277 @@ def _clear_dangerous_make_env_vars():
...
+        self.fuzzing = kwargs.get("fuzzing", False)
+        self.fett = kwargs.get("fett", False)
+
+    def __repr__(self):
+        flags = []
+        for key,val in self.__dict__.items():
+            if key == "kernconf" or key == "platform" or key == "kABI":
+                continue
+            if val:
+                flags.append(key)
```
jrtc27, line 122: Surely there's a better way to do this kind of thing?

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -67,6 +71,277 @@ def _clear_dangerous_make_env_vars():
...
+    def __str__(self):
+        platforms = []
+        for flag in ConfigPlatform:
+            if self & flag:
+                platforms.append(flag.name)
+        return "|".join(platforms)
+
+    @classmethod
+    def fpga_platforms(cls):
+        return (cls.BERI | cls.GFE)
```
jrtc27, line 96: Just make it a normal enum and use sets

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -67,6 +71,277 @@ def _clear_dangerous_make_env_vars():
             os.unsetenv(k)
             del os.environ[k]
 
+# XXX-AM: Replace this with CompilationTarget?
+class KernelABI(Enum):
+    NATIVE = "native"
```
jrtc27, line 76: What does native even mean for CHERI when hybrid and purecap exist too?

pycheribuild/config/compilation_targets.py
```diff
@@ -37,6 +37,7 @@
 from .loader import ConfigOptionBase
 from .target_info import (AutoVarInit, BasicCompilationTargets, CPUArchitecture, CrossCompileTarget, MipsFloatAbi,
                           TargetInfo)
+
```
jrtc27, line 40:
```suggestion
```

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -1259,45 +1534,65 @@ def __init__(self, config: CheriConfig):
...
+        return config.kernconf
+
+    def extra_kernel_configs(self):
+        xtarget = self.crosscompile_target
+        kernABI = self.get_default_kernel_abi()
+        platform = self.get_default_kernel_platform()
+        configs = []
+        if self.build_purecap_kernels and kernABI != KernelABI.PURECAP:
+            # The default config has already been added if the default is purecap
+            # XXX-AM: should be build the hybrid kernel when the default is purecap?
```
jrtc27, line 1570: Yes

pycheribuild/projects/run_qemu.py
```diff
@@ -413,28 +419,59 @@ def setup_config_options(cls, **kwargs):
...
     def _copy_kernel_image_from_remote_host(self):
         scp_path = os.path.expandvars(self.remote_kernel_path)
         self.info("Copying kernel image from build machine:", scp_path)
         self.makedirs(self.current_kernel.parent)
         self.copy_remote_file(scp_path, self.current_kernel)
 
     def process(self):
+        if self.list_kernels:
+            self._list_kernel_configs()
+            sys.exit(1)
```
jrtc27, line 474: 1 is wrong, it ran successfully. But this is very inflexible; what happens if I want, say, `--run/list-kernels run-mips64-purecap run-riscv64-purecap`? IMO this should be a `return` not a `sys.exit`.

pycheribuild/projects/cross/cheribsd.py
qwattash: Probably `NOCHERI` is better here, I wanted to convey the fact that it is the target without CHERI extensions.

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -1301,6 +1280,89 @@ def __init__(self, config: CheriConfig):
...
+        """
+        Return a list of pure capability kernels to build, the first one is used as the default
+        when the purecap kernel is selected as the default kernel
+        """
+        purecap_kernels = []
+        assert self.crosscompile_target.is_hybrid_or_purecap_cheri()
+        if self.compiling_for_mips(include_purecap=True):
+            purecap_kernels.append("CHERI_PURECAP_MALTA64")
+        elif self.compiling_for_riscv(include_purecap=True):
+            purecap_kernels.append("CHERI-PURECAP-QEMU-NODEBUG")
```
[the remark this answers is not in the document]
qwattash: I think I copied whatever was being done in the original function, but I do agree that we probably don't want the benchmark kernel there.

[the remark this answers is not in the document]
qwattash: I think with the current way of generating configuration entries this is fixed.

pycheribuild/projects/run_qemu.py
```diff
@@ -413,28 +418,71 @@ def setup_config_options(cls, **kwargs):
...
+        if self.qemu_options.can_boot_kernel_directly:
+            self._project_specific_options += ["-append", "kern.module_path={}".format(self.current_kernel.parent)]
         if hasattr(source_class, "get_rootfs_dir"):
             # noinspection PyCallingNonCallable
             self.rootfs_path = source_class.get_rootfs_dir(self, config=config)
         if needs_disk_image:
             self.disk_image = disk_image_class.get_instance(self).disk_image_path
 
+    def _valid_kernel_configs(self):
+        return self.source_project.get_kernel_configs(platform=ConfigPlatform.QEMU)
```
arichardson, line 465: Should we just list all files `sys/<arch>/conf/` that don't contain a "." to get the full list?

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -68,6 +71,333 @@ def _clear_dangerous_make_env_vars():
...
+        self.kernconf = kernconf
+        self.platform = {platform} if type(platform) != set else platform
+        self.kABI = kwargs.get("kABI", KernelABI.NOCHERI)
+        self.default = kwargs.get("default", False)
+        self.caprevoke = kwargs.get("caprevoke", False)
+        self.mfsroot = kwargs.get("mfsroot", False)
+        self.debug = kwargs.get("debug", False)
+        self.benchmark = kwargs.get("benchmark", False)
+        self.fuzzing = kwargs.get("fuzzing", False)
+        self.fett = kwargs.get("fett", False)
```
arichardson, line 106:
```suggestion
    def __init__(self, kernconf: str, platform: ConfigPlatform, kABI=KernelABI.NOCHERI, default=False, ...):
        self.kernconf = kernconf
        self.platform = {platform} if type(platform) != set else platform
        self.kABI = kABI
        self.default = default
        ...
```
Is there a reason we can't use default arguments here?

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -68,6 +71,333 @@ def _clear_dangerous_make_env_vars():
...
+        ConfigPlatform.QEMU: "MALTA64",
+        ConfigPlatform.BERI: "DE4"
+    }
+
+    def get_kabi_name(self, platform, kABI):
+        if platform == ConfigPlatform.BERI and kABI == KernelABI.NOCHERI:
+            return "BERI"
+        return super().get_kabi_name(platform, kABI)
+
+    def get_flag_names(self, **kwargs):
```
arichardson, line 172: default arguments?

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -68,6 +71,333 @@ def _clear_dangerous_make_env_vars():
...
+        ConfigPlatform.GFE: "GFE"
+    }
+
+    def __init__(self):
+        # Note: the FETT FPGA kernels do not have the platform name in
+        # the name, so we drop the platform from the kerconf components
+        self.fett_gfe_kernconf_components = OrderedDict(self.kernconf_components)
+        del self.fett_gfe_kernconf_components["platform_name"]
+
+    def get_flag_names(self, **kwargs):
```
arichardson, line 224: Same here?

pycheribuild/projects/run_qemu.py
```diff
@@ -99,6 +99,8 @@ def setup_config_options(cls, default_ssh_port: int = None, **kwargs):
                                                             help="The port on localhost to forward to the QEMU ssh "
                                                                  "port. You can then use `ssh root@localhost -p $PORT` "
                                                                  "to connect to the VM")
+        cls.snapshot = cls.add_bool_option("snapshot", show_help=True,
+                                           help="Run in snapshot mode, changes to the disk image are non-persistent")
```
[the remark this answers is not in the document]
qwattash: Makes sense, I used the same name as the qemu flag without much thought, but as Jess noted it is already confusing on the qemu side as there is also a separate snapshot functionality.

pycheribuild/projects/run_qemu.py
qwattash: There are configs in `sys/<arch>/conf` that we do not use or support in cheribuild and we never build. This would be sensible if we generated the set of kernel configurations by reading files in `sys/<arch>/conf` in the first place, which is possible, but I just went the other way around in the generator to avoid having a blacklist of files. I can try to have the generator read config files in `sys/<arch>` but would leave this call as is, so that it does not care about how we get configuration names

pycheribuild/projects/cross/cheribsd.py
qwattash: Not really, was easier to add new flags on the fly but perhaps less readable.

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -1476,56 +1776,39 @@ def _build_and_install_kernel_binaries(self, build_cheribsd: BuildCHERIBSD, kern
...
+        default_kABI = self.get_default_kernel_abi()
+        other_kABI = KernelABI.PURECAP if default_kABI != KernelABI.PURECAP else KernelABI.HYBRID
+        platforms = self.get_default_kernel_platform()
+        if self.build_fpga_kernels:
+            platforms |= ConfigPlatform.fpga_platforms()
+        configs = CheriBSDConfigTable.get_configs(xtarget, platforms, default_kABI,
+                                                  mfsroot=True)
+        configs += CheriBSDConfigTable.get_configs(xtarget, platforms, default_kABI,
+                                                   mfsroot=True, benchmark=True)
+        if self.build_purecap_kernels:
```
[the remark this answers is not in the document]
qwattash: This is because when the default kernel is already purecap, the build-purecap-kernels flag adds the hybrid kernels to the build instead. This was changed originally after commenting with @jrtc27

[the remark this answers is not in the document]
arichardson: Ah I missed that discussion. Maybe we should rename the field to something like "build_alternative_abi_kernels"?

pycheribuild/projects/run_qemu.py
arichardson: Sorry, I should have been a bit clearer in my comment: I meant printing the files in `sys/<arch>/conf` only for the --list-kernels option (possibly indicating which ones are automatically used by cheribuild), so that you can see the full list of strings that you can pass as the --kernel-config override. Does that make more sense?

qwattash: Ah yes that can be done, however I think that in the cheribsd target it should list the kernels that can be given as default kernel overriede in --cheribsd/kernel-config, instead in qemu it should list the kernels that have been built and can be run (yet with the --qemu/kernel-config override, possibly changing the name of the qemu option to avoid confusion)

pycheribuild/config/compilation_targets.py
```diff
@@ -517,23 +539,21 @@ def run_fpga_benchmark(self, benchmarks_dir: Path, *, output_file: str = None, b
...
             mfs_kernel = BuildCheriBsdMfsKernel.get_instance_for_cross_target(self._get_mfs_kernel_xtarget(),
                                                                               self.config, caller=self.project)
-            if self.config.benchmark_with_debug_kernel:
-                kernel_config = mfs_kernel.fpga_kernconf
-            else:
-                kernel_config = mfs_kernel.fpga_kernconf + "_BENCHMARK"
-            basic_args.append(
-                "--kernel-img=" + str(mfs_kernel.installed_kernel_for_config(self.project, kernel_config)))
+            kernel_config = mfs_kernel.default_kernel_config(ConfigPlatform.BERI,
+                                                             benchmark=(not self.config.benchmark_with_debug_kernel))
```
arichardson, line 554:
```suggestion
                                                             benchmark=not self.config.benchmark_with_debug_kernel)
```

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -679,6 +995,18 @@ def clean(self) -> ThreadJoiner:
         else:
             return self.async_clean_directory(builddir)
 
+    def _list_kernel_configs(self):
+        """Emit a list of valid kernel configurations that can be given as --kernel-config overrides"""
+        conf_dir = self.source_dir / "sys" / self.target_info.freebsd_target_arch_dir / "conf"
+        configs = conf_dir.glob("*")
+        blacklist = ["NOTES", "LINT", "DEFAULTS"]
+        self.info("Valid kernel configuration files for --freebsd/kernel-config:")
```
arichardson, line 1003:
```suggestion
        self.info("Valid kernel configuration files for --" +self.target + "/kernel-config:")
```

arichardson: Or even better since it ensures the second part won't become stale:
```suggestion
        self.info("Valid kernel configuration files for", self.get_config_option_name("kernel_config"))
```

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -1239,15 +1593,18 @@ def setup_config_options(cls, install_directory_help=None, **kwargs):
                                                  help="Path to an MFS root image to be embedded in the kernel for "
                                                       "booting")
 
+        cls.default_kernel = cls.add_config_option("default-kernel", show_help=True, _allow_unknown_targets=True,
```
arichardson, line 1596: Maybe this should be "default-kernel-abi" instead? Might make the --help confusing otherwise.

arichardson: And I guess it also wants `only_add_for_targets`?

arichardson (approved):
A few minor points, but I think this is ready to land otherwise.

pycheribuild/config/compilation_targets.py
jrtc27: still there

pycheribuild/config/compilation_targets.py
```diff
@@ -276,6 +277,22 @@ def triple_for_target(cls, target: "CrossCompileTarget", config: "CheriConfig",
         # TODO: do we need any special cases here?
         return target.cpu_architecture.value + common_suffix
 
+    @property
+    def freebsd_target_arch_dir(self):
+        """
+        Return the name of the target architecture machine-dependent code directory
+        in the source tree. (e.g. <arch> in sys/<arch>)
+        """
```
jrtc27, line 285: Isn't this TARGET_CPUTYPE, i.e. we should call this function freebsd_target_cputype?

jrtc27: Uh, TARGET_CPUARCH even (TARGET_CPUTYPE is like -mcpu).

jrtc27: This is what cheribsd.py's arch_build_flags does, too, so if you fix this to be `freebsd_target` then you can remove most of that `if` and leave behind just the CHERI handling.

qwattash:
Now should replace  #196 via the `--build-bench-kernels` option available for both cheribsd and MFS root targets.

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -1630,50 +1636,63 @@ def get_default_kernel_abi(self):
...
+    def _get_all_kernel_configs(self):
+        kernABIs = self._get_kABIs_to_build()
+        platform = self.get_default_kernel_platform()
+        combinations = []
+        if self.build_bench_kernels:
+            combinations.append("benchmark")
+        if self.caprevoke_kernel:
+            combinations.append("caprevoke")
+        if self.build_fpga_kernels:
+            combinations.append("mfsroot")
```
arichardson: This does not seem right. Should it be like this below?
```python3
        if self.build_fpga_kernels:
            configs += self._get_config_variants([ConfigPlatform.fpga_platforms()], kernABIs, combinations + ["mfsroot"])
```

arichardson (approved):
Still looks fine. Would be great to have a unit test that checks the kernel config list for a few configurations, but not essential right now.

pycheribuild/projects/cross/cheribsd.py
qwattash: I think you are right

qwattash:
> Still looks fine. Would be great to have a unit test that checks the kernel config list for a few configurations, but not essential right now.

Added some basic tests and changed the one in master to drop the `_BENCHMARK` variant unless `--build-bench-kernels` is given.

pycheribuild/config/compilation_targets.py
jrtc27: This *still* hasn't been deleted

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -35,6 +36,9 @@
 import typing
 from pathlib import Path
 
+from collections import OrderedDict
+from enum import Enum
```
jrtc27, line 40: Why are these in a separate block?

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -68,6 +72,345 @@ def _clear_dangerous_make_env_vars():
...
+            flags.append("NFSROOT")
+        flags += super().get_flag_names(mfsroot=mfsroot, fuzzing=fuzzing)
+        if benchmark:
+            flags.append("BENCHMARK")
+        return flags
+
+    def make_config(self, platform, kABI, base_context=None, **kwargs):
+        kernconf_ctx = self._prepare_kernconf_context(platform, kABI, base_context=base_context, **kwargs)
+
+        valid_ctx_items = (v for v in kernconf_ctx.values() if v is not None and v != "")
```
jrtc27, line 194: If we return None for NOCHERI's kABI name can this lose the `!= ""`? Logically `""` would mean `-GFE` not `GFE`, for example.

pycheribuild/projects/cross/cheribsd.py
qwattash: No reason, stray newline.

pycheribuild/projects/cross/cheribsd.py
qwattash: Yes, fixed now.

tests/test_argument_parsing.py
```diff
@@ -617,9 +617,32 @@ def test_disk_image_path(target, expected_name):
...
-                 "CHERI_MALTA64", ["CHERI_DE4_USBROOT", "CHERI_DE4_USBROOT_BENCHMARK", "CHERI_DE4_NFSROOT"]),
+                 "CHERI_MALTA64", ["CHERI_DE4_USBROOT", "CHERI_DE4_NFSROOT"]),
+    pytest.param("cheribsd-mips64-purecap", ["--cheribsd/build-fpga-kernels", "--cheribsd/build-bench-kernels"],
+                 "CHERI_MALTA64", ["CHERI_DE4_USBROOT_BENCHMARK", "CHERI_DE4_USBROOT", "CHERI_DE4_NFSROOT"]),
+    pytest.param("cheribsd-mips64-purecap", ["--cheribsd/build-alternate-abi-kernels"],
+                 "CHERI_MALTA64", ["CHERI_PURECAP_MALTA64"]),
+    pytest.param("cheribsd-mips64-purecap",
+                 ["--cheribsd/build-alternate-abi-kernels",
+                  "--cheribsd/default-kernel-abi", "purecap"],
+                 "CHERI_PURECAP_MALTA64", ["CHERI_MALTA64"]),
```
arichardson, line 645: Could you add tests for morello too?

tests/test_argument_parsing.py
```diff
@@ -617,9 +617,32 @@ def test_disk_image_path(target, expected_name):
...
+    pytest.param("cheribsd-riscv64-purecap",
+                 ["--cheribsd/build-alternate-abi-kernels",
+                  "--cheribsd/default-kernel-abi", "purecap"],
+                 "CHERI-PURECAP-QEMU", ["CHERI-QEMU"]),
+    pytest.param("cheribsd-riscv64-purecap", ["--cheribsd/build-fett-kernels"],
+                 "CHERI-QEMU-FETT", ["CHERI-QEMU"]),
+    pytest.param("cheribsd-riscv64-purecap",
+                 ["--cheribsd/build-fett-kernels",
+                  "--cheribsd/build-alternate-abi-kernels"],
+                 "CHERI-QEMU-FETT", ["CHERI-QEMU", "CHERI-PURECAP-QEMU"]),
```
arichardson, line 634: Maybe also add one RISC-V case to show that benchmark builds `CHERI-QEMU-NODEBUG`?

qwattash: Turns out this was missing in the generator. Now fixed.

arichardson (approved):
LGTM once @jrtc27 is happy too.

jrtc27:
The unnecessary whitespace diff lives on 20 days later...

pycheribuild/config/compilation_targets.py
qwattash: I appear to have been very good at missing this :(

arichardson (approved):

PR #206, opened 2021-06-01
Add support for creating semi-minimal kernels for BESSPIN

brooksdavis:


brooksdavis:
oops, this branch has a couple local changes that I'd committed and never pushed. Should be fairly harmless.

pycheribuild/mtree.py
```diff
@@ -255,6 +256,29 @@ def __contains__(self, item):
         mtree_path = self._ensure_mtree_path_fmt(str(item))
         return mtree_path in self._mtree
 
+    def filter(self, globs, exceptions=[], print_status=False):
```
jrtc27, line 259:
```suggestion
    def filter_out(self, globs, exceptions=[], print_status=False):
```
? This to me implies positive not negative filtering.

pycheribuild/projects/cross/openssl.py
```diff
@@ -51,16 +50,23 @@ def setup(self):
         self.configure_args.append(str(self.source_dir / "Configure"))
         self.configure_args.append("BSD-generic64")
         self.configure_args.append("-shared")
-        self.configure_args.append("--install-prefix=" + str(self.destdir))
+	# XXX: should be based on version
```
jrtc27, line 53: Spaces

pycheribuild/projects/cross/openssl.py
```diff
@@ -51,16 +50,23 @@ def setup(self):
...
+        if (self.project_name != "fett-openssl"):
+            self.configure_args.append("--prefix=" + str(self.destdir) + str(self._install_prefix))
```
jrtc27, line 55: No parens, and looking at `self.project_name` is ugly

pycheribuild/projects/disk_image.py
```diff
@@ -1069,6 +1069,236 @@ def cheribsd_class(self):
...
+    include_boot = True
+
+    @classmethod
+    def setup_config_options(cls, **kwargs):
+        super().setup_config_options(default_hostname=_default_disk_image_hostname("cheribsd-besspin"),
+                                     extra_files_suffix="-besspin", **kwargs)
+        cls.strip_binaries = cls.add_bool_option(
+            "strip", default=True, help="strip ELF files to reduce size of generated image")
+        cls.use_cheribsd_purecap_rootfs = cls.add_bool_option(
+            "use-cheribsd-purecap-rootfs", default=True, help="Use the rootfs built by cheribsd-purecap instead")
```
jrtc27, line 1085: Do we really need this? It's a silly target to build, just leave it as hybrid and thus uninteresting?

pycheribuild/projects/cross/openssl.py
brooksdavis: I'm hoping to purge the FETT bits in October...

pycheribuild/projects/disk_image.py
brooksdavis: I'll remove this. It's leftover from foking the minimal target.

jrtc27: ... we should probably remove that for the same reason (though at least that one is default=False)

pycheribuild/mtree.py
brooksdavis: Thinking this morning, I wonder if `exclude_matching` would be better (it had been `filter_matching` initially in my tree)

pycheribuild/projects/disk_image.py
```diff
@@ -1069,6 +1069,223 @@ def cheribsd_class(self):
...
+            "libasn1.so.11",
+            "libypclnt.so.4",
+            "libcom_err.so.5",
+            "libhx509.so.11",
+            "libwind.so.11",
+            "libheimbase.so.11",
+            "libprivateheimipcc.so.11",
+            "libssl.so.111"
+        ]
+        self.mtree.exclude_matching("./usr/lib/*", list(map(lambda p: "./usr/lib/" + p, usr_lib_exceptions)))
```
arichardson, line 1244: Maybe better to just keep the `pam_*` exclusion and add the libraries explicitly afterwards? That way we should get a build error on a version bump/library rename.

pycheribuild/projects/disk_image.py
```diff
@@ -1069,6 +1069,223 @@ def cheribsd_class(self):
...
+        self.mtree.exclude_matching("./usr/libexec/*", ["*/getty"])
+
+        usr_sbin_exceptions = [
+            "service",
+            "sshd",
+            "newsyslog",
+            "utx",
+            "syslogd"
+        ]
+        self.mtree.exclude_matching("./usr/sbin/*", list(map(lambda p: "./usr/sbin/" + p, usr_sbin_exceptions)))
```
arichardson, line 1255: Much less of a problem here, but maybe removing all and then explicitly adding the 5 binaries is makes this code easier to understand? Some of the arrays are positive, others are negative.

arichardson (approved):
Could we use this to build minimal FreeBSD images (since we can't use cheribsdbox there)?

PR #211, opened 2021-06-24
Add option to build qemu with profiler enabled.

qwattash:
Originally from @LawrenceEsswood

### DIFF
--- a/pycheribuild/projects/build_qemu.py
+++ b/pycheribuild/projects/build_qemu.py
@@ -72,6 +72,8 @@ class BuildQEMUBase(AutotoolsProject):
 
         cls.gui = cls.add_bool_option("gui", show_help=False, default=False,
                                       help="Build a the graphical UI bits for QEMU (SDL,VNC)")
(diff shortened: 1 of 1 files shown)


pycheribuild/projects/build_qemu.py
```diff
@@ -72,6 +72,8 @@ def setup_config_options(cls, **kwargs):
 
         cls.gui = cls.add_bool_option("gui", show_help=False, default=False,
                                       help="Build a the graphical UI bits for QEMU (SDL,VNC)")
+        cls.profile = cls.add_bool_option("profile", show_help=False, default=False,
```
arichardson, line 75:
```suggestion
        cls.profile = cls.add_bool_option("build-profiler", show_help=False, default=False,
```
maybe buid-profiler makes more sense? `--qemu/profile` sounds to me like the build will be profiled.

arichardson (approved):

pycheribuild/projects/build_qemu.py
qwattash: Good point

PR #212, opened 2021-06-30
Update netperf job to build for RISC-V and link statcounters libraries.

qwattash:


pycheribuild/projects/cross/benchmarks.py
```diff
@@ -786,35 +786,40 @@ def install(self, **kwargs):
...
         self.configure_args.append("--enable-unixdomain")
+        if self.hw_counters:
+            self.configure_args.append("--enable-pmc={}".format(self.hw_counters))
         self.add_configure_vars(ac_cv_func_setpgrp_void="yes")
         super().configure(**kwargs)
 
+    def process(self):
+        if (self.compiling_for_riscv(include_purecap=True) and
+                self.hw_counters == "pmc"):
+            self.fatal("hwpmc not supported on riscv")
```
arichardson, line 820: Is this to support running on Morello? Should we add Morello to the supported architectures?

arichardson (approved):
Looks reasonable to me.

pycheribuild/projects/cross/benchmarks.py
qwattash: yes and no. Statcounters are not supported but hwpmc should be. I left it out as I have not tested any of it yet, and I have to add the code to netperf to enable and read hwpmc for morello.

PR #215, opened 2021-07-12
Improve logic for FETT install dirs

arichardson:


brooksdavis (approved):
Seems like a significant improvement.  I'll test further tomorrow and update the docs once it/s committed.

PR #216, opened 2021-07-16
Extra support for running CheriBSD tests on QEMU Morello

qwattash:


pycheribuild/boot_cheribsd/__init__.py
```diff
@@ -815,15 +823,23 @@ def boot_and_login(child: CheriBSDSpawnMixin, *, starttime, kernel_init_only=Fal
...
         i = child.expect(boot_messages, timeout=5 * 60, timeout_msg="timeout before /sbin/init")
         # Skip 10s wait from x86 loader if we see the "Hit [Enter] to boot" message
         if i == 1:  # Hit Enter
+            if boot_alternate_kernel:
+                # Expected loader(8) to skip autoboot, fail
+                failure("expected autoboot disabled but got autoboot prompt")
             success("Got '", child.match.string, "' from loader")
             child.sendline("")
             i = child.expect(boot_messages, timeout=5 * 60, timeout_msg="timeout before /sbin/init")
+        if i == 6:  # loader(8) prompt
```
arichardson, line 839: There are so many messages now, I think we should consider using .index() like below to avoid future bugs.

qwattash: There is also another issue here, the alternate kernel is None if we are booting the default kernel, so I need to just send the "boot" command in that case.

PR #217, opened 2021-07-16
llvm: disable the new pass manager

brooksdavis:
Linking libunwind.cpp on CheriBSD/RISC-V is broken with the new pass
manager so disable it until that can be addressed.


brooksdavis:
This is already in master, something was causing my system not to update.

PR #219, opened 2021-07-20
Drop support for Ubuntu 16.04 and Python < 3.6

arichardson:
Ubuntu 16.04 has been EOL for a while and based on the python versions
available on most Linux distributions requiring 3.6 seems like a
reasonable baseline. While I would like to use 3.7 for dataclasses,
Ubuntu 18.04 ships 3.6 and that's probably the distribution that is most
commonly used when building CHERI projects.
Bumping the Python version to 3.6 allows using features such as f-strings
and type annotations on variables (https://docs.python.org/3/whatsnew/3.6.html).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,25 +7,7 @@ pipeline {
   stages {
   stage('Test') {
   parallel {
-    stage('Test Python 3.5.0') {
-      agent {
-        dockerfile {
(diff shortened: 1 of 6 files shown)


arichardson:
Looks like Ubuntu 18.04 still has 3.6.9 so we can't use 3.7 :/

jrtc27:
Yeah, 3.6 only for another year or two

PR #221, opened 2021-07-23
Add the ability to filter the built dependencies with regexes

arichardson:
When building e.g. KDE programs, I want to build the KDE framework
dependencies but skip targets that take a long time to build such as QtBase
and all the underlying dependencies. This patch lets me do the following:

`cheribuild.py <target> -d --skip-sdk "--skip-dependency-filter=qt.*" "--skip-dependency-filter=libx.*"`

This avoids lots of build time (especially when combined with --clean or
--reconfigure since we also need to build a native qtbase+qtdeclarative
for the bootstrap tools in KConfig and other frameworks.

### DIFF
--- a/pycheribuild/config/chericonfig.py
+++ b/pycheribuild/config/chericonfig.py
@@ -32,6 +32,7 @@ import getpass
 import grp
 import json
 import os
+import re
 import shutil
 import typing
 from collections import OrderedDict
@@ -98,6 +99,11 @@ def _default_arm_none_eabi_prefix(c: "CheriConfig", _):
             return str(Path(in_path).parent / "arm-none-eabi-")
         # Otherwise suggest the non-existent local installation
         return str(default_path / "bin/arm-none-eabi-")
+
+
+def _skip_dependency_filter_arg(values: "list[str]") -> "list[re.Pattern]":
+    result = [re.compile(item) for item in values]
+    return result
 
 
 class CheriConfig(ConfigBase):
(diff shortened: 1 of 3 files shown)


arichardson:
Not sure about the flag name, maybe it should be shorter (e.g. `skip-deps=`) or maybe we want that as an alias?

pycheribuild/config/chericonfig.py
```diff
@@ -242,6 +248,11 @@ def __init__(self, loader, action_class):
             help="When building with --include-dependencies ignore the SDK dependencies. Saves a lot of time "
                  "when building libc++, etc. with dependencies but the sdk is already up-to-date. "
                  "This is like --no-include-toolchain-depedencies but also skips the target that builds the sysroot.")
+        self.skip_dependency_filters = loader.add_option(
+            "skip-dependency-filter", group=loader.dependencies_group, action="append", default=[],
+            type=_skip_dependency_filter_arg, metavar="REGEX",
+            help="A regular expression to match against to target names that should be skipped when using"
+                 "--include-dependency. Can be passed multiple times to add more patters.")  # type: list[re.Pattern]
```
jrtc27, line 255: s/patters/patterns/

pycheribuild/config/chericonfig.py
```diff
@@ -444,6 +455,10 @@ def load(self):
         if "CLICOLOR" in os.environ:
             del os.environ["CLICOLOR"]
 
+        # Check that the skip_dependency_filters arguments are all valid regular expressions (do it now since otherwise
+        # the validation is delayed until the first time the object is used.
```
jrtc27, line 459: Missing )

PR #222, opened 2021-07-23
Don't add -hybrid compilation targets for most projects

arichardson:
For almost all targets compiling hybrid does not change anything (other
than some minor code-generation differences due being able to use
capability registers for memory accesses) since the projects don't have
any __capability annotations. This should speed up running cheribuild a
bit since it doesn't have to parse as many command line flags and also
prevents users from building targets that don't make much sense.

Before: `cheribuild.py --list-targets | wc -l` -> 2381
After:  `cheribuild.py --list-targets | wc -l` -> 1871


arichardson:
@brooksdavis I'm not sure about the FETT targets, does it make sense to keep hybrid?

jrtc27:
This seems fine to me. I don't think hybrid FETT targets make any sense.

I would personally like to go further and have all hybrid targets (other than hybrid-for-purecap-rootfs ones) hidden/disabled by default as users should not be trying to use them unless they really know what they're doing. Like a `cheribuild --enable-hybrid-targets` or something (and those that really want it will likely put it in their cheribuild.json).

arichardson:
> This seems fine to me. I don't think hybrid FETT targets make any sense.
> 
> I would personally like to go further and have all hybrid targets (other than hybrid-for-purecap-rootfs ones) be hidden/disabled by default as users should not be trying to use them unless they really know what they're doing. Like a `cheribuild --enable-hybrid-targets` or something (and those that really want it will likely put it in their cheribuild.json).

That sounds like a great idea to me. The only problem I see is that the current ways that arguments are parsed might make this a bit awkward but I'll have a look if it can be done easily.

brooksdavis:
Removing the FETT targets was fine. I might have used them for debugging something once, but they should not be needed.

PR #223, opened 2021-07-23
Add rules to build sslproc and an sslproc-enabled Apache.

bsdjhb:
This probably has some rough edges that need to be fixed.

### DIFF
--- a/pycheribuild/projects/cross/apache.py
+++ b/pycheribuild/projects/cross/apache.py
@@ -28,7 +28,9 @@
 # OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 # SUCH DAMAGE.
 #
-from .crosscompileproject import (CrossCompileAutotoolsProject, DefaultInstallDir, GitRepository, SubversionRepository)
+from .crosscompileproject import CrossCompileAutotoolsProject, CrossCompileCMakeProject, DefaultInstallDir
+from .crosscompileproject import GitRepository, SubversionRepository
+from ..project import ReuseOtherProjectRepository
 from .expat import BuildExpat
 
 
@@ -129,3 +131,31 @@ class BuildApache(CrossCompileAutotoolsProject):
                          cwd=self.build_dir / "server")
             self.run_cmd(str(self.host_CC), "gen_test_char.lo", "-o",
                          "gen_test_char", cwd=self.build_dir / "server")
+
+
+class BuildSSLProc(CrossCompileCMakeProject):
+    target = "sslproc"
+
+    repository = GitRepository("https://github.com/CTSRD-CHERI/sslproc.git")
+
+    has_optional_tests = True
+    default_build_tests = False
+    show_optional_tests_in_help = True
+
+    def setup(self):
+        super().setup()
+        self.add_cmake_options(BUILD_TESTS=self.build_tests)
+
+
+class BuildSSLProcApache(BuildApache):
+    target = "apache-sslproc"
+
+    repository = ReuseOtherProjectRepository(BuildApache, do_update=True)
+
+    dependencies = BuildApache.dependencies + ["sslproc"]
+
+    def setup(self):
+        super().setup()
+        self.configure_args.append(
+            "--with-sslproc=" + str(BuildSSLProc.get_install_dir(self))
+            )


pycheribuild/projects/cross/apache.py
```diff
@@ -129,3 +129,36 @@ def configure(self, **kwargs):
...
             self.run_cmd(str(self.host_CC), "gen_test_char.lo", "-o",
                          "gen_test_char", cwd=self.build_dir / "server")
+
+class BuildSSLProc(CrossCompileCMakeProject):
+    target = "sslproc"
+
+    repository = GitRepository("https://github.com/CTSRD-CHERI/sslproc.git")
+
+    native_install_dir = DefaultInstallDir.IN_BUILD_DIRECTORY
+    cross_install_dir = DefaultInstallDir.ROOTFS_LOCALBASE
```
jrtc27, line 139: This should be the default?

pycheribuild/projects/cross/apache.py
```diff
@@ -129,3 +129,36 @@ def configure(self, **kwargs):
...
+    @classmethod
+    def setup_config_options(cls, **kwargs):
+        super().setup_config_options(**kwargs)
+        cls.build_tests = cls.add_bool_option("build-tests", help="Build tests",
+                                              show_help=True)
+
+    def setup(self):
+        super().setup()
+        if self.build_tests:
+            self.configure_args.append("-DBUILD_TESTS=ON")
```
jrtc27, line 150:
```suggestion
            self.add_cmake_options(BUILD_TESTS=True)
```

pycheribuild/projects/cross/apache.py
```diff
@@ -129,3 +129,36 @@ def configure(self, **kwargs):
...
+
+class BuildSSLProcApache(BuildApache):
+    target = "apache-sslproc"
+    default_directory_basename = "apache"
+    build_dir_suffix = "-sslproc"
+
+    # XXX: Can this extend the list from BuildApache?
+    dependencies = ["apr", "pcre", "sslproc"]
```
jrtc27, line 158:
```suggestion
    dependencies = BuildApache.dependencies + ["sslproc"]
```
should work?

pycheribuild/projects/cross/apache.py
```diff
@@ -129,3 +129,36 @@ def configure(self, **kwargs):
...
+        super().setup()
+        if self.build_tests:
+            self.configure_args.append("-DBUILD_TESTS=ON")
+
+class BuildSSLProcApache(BuildApache):
+    target = "apache-sslproc"
+    default_directory_basename = "apache"
+    build_dir_suffix = "-sslproc"
+
+    # XXX: Can this extend the list from BuildApache?
```
bsdjhb, line 157: Not sure if there's a better way to do this.  I'd really like to be able to say something like:

```
    dependencies.append("sslproc")
```

assuming the super() constructor had run first, but maybe that doesn't work?  Would I need to do something like:

```
    dependencies = BuildApache.dependencies + ["sslproc"]
```

pycheribuild/projects/cross/apache.py
```diff
@@ -129,3 +129,36 @@ def configure(self, **kwargs):
...
+    dependencies = ["apr", "pcre", "sslproc"]
+
+    def setup(self):
+        super().setup()
+        self.configure_args.extend([
+            "--with-sslproc=" + str(BuildSSLProc.get_install_dir(self)),
+            ])
```
jrtc27, line 164: Use append if it's just the one option

pycheribuild/projects/cross/apache.py
jrtc27: We raced here. The latter works. The former doesn't at that scope, but we do let dependencies be a method instead, which would mean writing it as:
```
    @classmethod
    def dependencies(cls, config):
        return super().dependencies + ["sslproc"]
```
(yay dynamic typing?..)

bsdjhb: Umm, I'm going to go with your suggestion (option 2 above).  The other feels a bit too gross. :)

pycheribuild/projects/cross/apache.py
```diff
@@ -129,3 +129,36 @@ def configure(self, **kwargs):
...
+
+    repository = GitRepository("https://github.com/CTSRD-CHERI/sslproc.git")
+
+    native_install_dir = DefaultInstallDir.IN_BUILD_DIRECTORY
+    cross_install_dir = DefaultInstallDir.ROOTFS_LOCALBASE
+
+    @classmethod
+    def setup_config_options(cls, **kwargs):
+        super().setup_config_options(**kwargs)
+        cls.build_tests = cls.add_bool_option("build-tests", help="Build tests",
```
arichardson, line 144: This appears to be rather common, I wonder if we should allow something like `has_optional_tests = True` to avoid all this duplication.

arichardson (approved):
flake8 points out a few coding style issues
```
pycheribuild/projects/cross/apache.py:31:121: E501 line too long (145 > 120 characters)
from .crosscompileproject import (CrossCompileAutotoolsProject, CrossCompileCMakeProject, DefaultInstallDir, GitRepository, SubversionRepository)
                                                                                                                        ^
pycheribuild/projects/cross/apache.py:133:1: E302 expected 2 blank lines, found 1
class BuildSSLProc(CrossCompileCMakeProject):
^
pycheribuild/projects/cross/apache.py:152:1: E302 expected 2 blank lines, found 1
class BuildSSLProcApache(BuildApache):
^
2     E302 expected 2 blank lines, found 1
1     E501 line too long (145 > 120 characters)
3
Error: Process completed with exit code 1.
```

LGTM once those issues and @jrtc27's comments have been addressed

pycheribuild/projects/cross/apache.py
bsdjhb: For both or just cross_install_dir?  (I had read this as only being for cross_install_dir at first and only removed that one, but on re-reading I see you might have meant both?)

jrtc27: I don't recall what I meant, but it does apply to both

pycheribuild/projects/cross/apache.py
```diff
@@ -129,3 +129,34 @@ def configure(self, **kwargs):
...
+                                              show_help=True)
+
+    def setup(self):
+        super().setup()
+        if self.build_tests:
+            self.add_cmake_options(BUILD_TESTS=True)
+
+class BuildSSLProcApache(BuildApache):
+    target = "apache-sslproc"
+    default_directory_basename = "apache"
```
jrtc27, line 153: Should this not be `repository = ReuseOtherProjectRepository(BuildApache, do_update=True)` instead?

bsdjhb: Whichever is preferred?  (I don't care either way)

bsdjhb: Using this does let me remove `build_suffix` so that does seem cleaner.

pycheribuild/projects/cross/apache.py
arichardson: I've added support for this now so you can do the following instead:
```
    has_optional_tests = True
    default_build_tests = False
    show_optional_tests_in_help = True
```

pycheribuild/projects/cross/apache.py
```diff
@@ -129,3 +131,36 @@ def configure(self, **kwargs):
...
+    @classmethod
+    def setup_config_options(cls, **kwargs):
+        super().setup_config_options(**kwargs)
+        cls.build_tests = cls.add_bool_option("build-tests", help="Build tests",
+                                              show_help=True)
+
+    def setup(self):
+        super().setup()
+        if self.build_tests:
+            self.add_cmake_options(BUILD_TESTS=True)
```
arichardson, line 152:
```suggestion
        self.add_cmake_options(BUILD_TESTS=self.build_tests)
```

arichardson (approved):

PR #225, opened 2021-07-27
Rather awful approach to add a $ORIGIN RUNPATH to X11 libs

arichardson:
Some binaries (e.g. xkbcomp) can't find the required libraries otherwise.
Dealing with the multiple levels of quoting in autotools projects is
rather awkward and it took me a while to find something that works:
`'$$ORIGIN/../lib'` (in single quotes).

### DIFF
--- a/pycheribuild/projects/cross/x11.py
+++ b/pycheribuild/projects/cross/x11.py
@@ -33,7 +33,7 @@ from .freetype import BuildFreeType2
 from ..project import DefaultInstallDir, GitRepository
 from ...config.chericonfig import BuildType
 from ...config.compilation_targets import CompilationTargets
-from ...processutils import set_env
+from ...processutils import DoNoQuoteStr, set_env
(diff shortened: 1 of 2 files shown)


pycheribuild/projects/cross/x11.py
```diff
@@ -79,6 +79,14 @@ def setup(self):
             # Needed for many of the projects but not all of them:
             self.configure_args.append("--enable-malloc0returnsnull")
 
+    @property
+    def default_ldflags(self):
+        result = super().default_ldflags
+        if not self.compiling_for_host():
+            result.append(DoNoQuoteStr("-Wl,--enable-new-dtags,-rpath,'$$ORIGIN/../lib'"))
+            result.append("-Wl,-z,origin")
```
arichardson, line 87: Probably shouldn't have copy-pasted that flag, since it means rtld won't look in the normal search dirs.

PR #226, opened 2021-07-29
Automatically start xvnc if it's installed

arichardson:
This commit installs a etc/rc.d/xvnc as part of xvnc-server-<arch> and
modifies the rc.conf template installed by disk-image to enable the
service by default.


arichardson:
I guess we could also default to starting it with password auth so that the macOS builtin VNC viewer works?

PR #228, opened 2021-08-04
Use FAT-32 for EFI filesystem images made via makefs.

bsdjhb:
makefs defaults to FAT-12.


bsdjhb:
I've only tested this on RISC-V and not Morello FVP.  For upstream FreeBSD I think we use FAT-16 as FAT-12 is not supported in all EFI ROMs (or only supported when dealing emulated floppies).  I didn't dig into how to do this for mtools either which is still using whatever default it uses.

arichardson:
> I've only tested this on RISC-V and not Morello FVP. For upstream FreeBSD I think we use FAT-16 as FAT-12 is not supported in all EFI ROMs (or only supported when dealing emulated floppies). I didn't dig into how to do this for mtools either which is still using whatever default it uses.

I wouldn't worry about mtools, that's just a fallback I added back when makefs FAT support was broken.

arichardson (approved):

PR #229, opened 2021-08-04
Lib64c

bsdjhb:
The first commit is probably a prerequisite for merging the CheriBSD lib64c change.  The second commit probably needs to wait a while until we deprecate the libcheri name and assume lib64c is present everywhere instead.


arichardson:
I think this is now ready to be merged?

jrtc27:
Have the caprevoke and cocall branches caught up with the change? Should wait on that before merging.

arichardson:
> Have the caprevoke and cocall branches caught up with the change? Should wait on that before merging.

Ah yes, forgot about them.

bsdjhb:
I also don't know if we don't want to wait for morello-llvm to pull in the lib64c change as well?

jrtc27:
Given we create the symlinks I don't think that matters?

bsdjhb:
Probably, yes.

arichardson (approved):

PR #231, opened 2021-08-12
 Update Dockerfile and add UID/GID support

gvnn3:
Update the Dockerfile to reference more recent versions of Ubuntu and tools.

Add support for docker-uid/docker-gid config values to be used when running the docker image.


docker/Dockerfile
```diff
@@ -20,18 +24,20 @@ COPY cheribuild.json /root/.config/cheribuild.json
...
 RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
 RUN echo "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-8 main" > /etc/apt/sources.list.d/llvm.list
 # RUN echo "deb http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu trusty main" > /etc/apt/sources.list.d/r-toolchain.list
 # RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 1E9377A2BA9EF27F
-RUN apt-get update && apt-get install -y clang-8 lld-8
+RUN apt-get update && apt-get install -y clang-10 lld-10
 
 RUN apt-get update && apt-get install -y cmake
 
+RUN groupadd -g 12345 -r cheri && useradd -u 12345 --no-log-init -r -g cheri cheri
```
jrtc27, line 39: No home directory? Also, adduser is preferred and can create both the group and user in one command

pycheribuild/__main__.py
```diff
@@ -198,8 +198,12 @@ def real_main():
                 subprocess.check_call(start_cmd)
                 docker_run_cmd = ["docker", "exec", cheri_config.docker_container] + cheribuild_args
             else:
-                docker_run_cmd = ["docker", "run", "--user", str(os.getuid()) + ":" + str(os.getgid()),
-                                  "--rm", "--interactive", "--tty"] + docker_dir_mappings
+                if cheri_config.docker_uid != "":
+                        docker_run_cmd = ["docker", "run", "--user", cheri_config.docker_uid + ":" + cheri_config.docker_gid,
+                                          "--rm", "--interactive", "--tty"] + docker_dir_mappings
+                else:
+                        docker_run_cmd = ["docker", "run", "--user", str(os.getuid()) + ":" + str(os.getgid()),
```
jrtc27, line 205: Use variables to avoid the duplication

pycheribuild/__main__.py
```diff
@@ -198,8 +198,12 @@ def real_main():
                 subprocess.check_call(start_cmd)
                 docker_run_cmd = ["docker", "exec", cheri_config.docker_container] + cheribuild_args
             else:
-                docker_run_cmd = ["docker", "run", "--user", str(os.getuid()) + ":" + str(os.getgid()),
-                                  "--rm", "--interactive", "--tty"] + docker_dir_mappings
+                if cheri_config.docker_uid != "":
```
jrtc27, line 201: What if I specify a UID but not a GID?

pycheribuild/config/chericonfig.py
```diff
@@ -278,6 +278,8 @@ def __init__(self, loader, action_class):
                                                              help="Attach to the same container again (note: "
                                                                   "docker-container option must be an id rather than "
                                                                   "a container name")
+        self.docker_uid = loader.add_option("docker-uid", help="UID to under Docker", default="", group=loader.docker_group)
```
jrtc27, line 281: Missing a word? Also why "" rather than None? And could you not just default this to os.get[ug]id() so you don't need to check whether it's set?

pycheribuild/projects/release.py
```diff
@@ -0,0 +1,162 @@
...
+        Path(source_root, "cheribuild/cheribuild.json").write_text("""{
+            "source-root": "../../sources",
+            "build-root": "../../build",
+            "output-root": "../../output",
+            "skip-update": true
+} 
+            """)
+
+
+class MorelloRelease(Release):
```
jrtc27, line 98: Would these not be better as a single "CrossCompile" project so you have a single class that does multiple architectures?

pycheribuild/projects/release.py
```diff
@@ -0,0 +1,162 @@
...
+    target = "tag"
+    
+    @classmethod
+    def setup_config_options(cls, **kwargs):
+        super().setup_config_options(**kwargs)
+
+        cls.gittag = cls.add_config_option("gittag", help="Tag to apply")
+
+    def __init__(self, config: CheriConfig):
+        super().__init__(config)
```
jrtc27, line 55: These aren't needed if you're not doing anything in the overridden constructor

pycheribuild/projects/release.py
```diff
@@ -0,0 +1,162 @@
...
+    project_name = "release"
+    do_not_add_to_targets = True
+
+    def __init__(self, config: CheriConfig):
+        super().__init__(config)
+
+    def process(self):
+        source_root = self.config.source_root
+
+        Path(source_root, "cheribuild/cheribuild.json").write_text("""{
```
jrtc27, line 89: These will have side-effects in --pretend mode I believe, and thus during the "pretend build everything" test

pycheribuild/projects/release.py
```diff
@@ -0,0 +1,162 @@
...
+            "source-root": "../../sources",
```
jrtc27, line 90: Is this evaluated relative to cheribuild.py or the current working directory?

Also, that would result in a non-standard layout; the standard layout is that build-root is source-root/build and output-root is source-root/output.

pycheribuild/projects/release.py
```diff
@@ -0,0 +1,162 @@
...
+
+        run_command("bsdtar", "-cavf", output_root / "release-morello.tar.xz", 
+            "-C", root,
+            "--options=xz:threads=" + str(default_make_jobs_count()),
+            "--options=compression-level=9",  # reduce size a bit more
+            "--exclude=*.git",
+            "output/morello-sdk/firmware",
+            "output/cheribsd-morello-purecap.img",
+            "output/sdk",
+            "sources/cheribuild",
```
jrtc27, line 132: Only because you've configured your cheribuild to have a non-standard layout. If I were to run this it wouldn't work. If you want a specific layout you need to assemble that yourself in a staging directory.

pycheribuild/projects/release.py
```diff
@@ -0,0 +1,162 @@
...
+        install_script.chmod(0o755)
+
+        run_command("bsdtar", "-cavf", output_root / "release-morello.tar.xz", 
+            "-C", root,
+            "--options=xz:threads=" + str(default_make_jobs_count()),
+            "--options=compression-level=9",  # reduce size a bit more
+            "--exclude=*.git",
+            "output/morello-sdk/firmware",
+            "output/cheribsd-morello-purecap.img",
+            "output/sdk",
```
jrtc27, line 131: Missing the rootfs

pycheribuild/projects/release.py
```diff
@@ -0,0 +1,162 @@
...
+        root = self.config.output_root.parent
+        output_root = self.config.output_root
+
+        install_script = Path(output_root, "install_and_run_fvp.sh")
+        install_script.write_text("""#!/bin/sh
+        dir=$(CDPATH= cd -- "$(dirname -- "$0")" && pwd)
+        exec "${dir}/cheribuild.py" install-morello-fvp run-fvp-morello-purecap "$@""")
+        install_script.chmod(0o755)
+
+        run_command("bsdtar", "-cavf", output_root / "release-morello.tar.xz",
```
jrtc27, line 124: Should we be putting -purecap in the names? For Morello it's perhaps unnecessary as foo-morello doesn't exist, only foo-morello-{hybrid,purecap}, so one can infer it's purecap, but for RISC-V we do have foo-riscv64 without any suffix that it might get confused with, and we should probably be consistent.

pycheribuild/projects/release.py
```diff
@@ -0,0 +1,162 @@
...
+            "source-root": "../../sources",
+            "build-root": "../../build",
+            "output-root": "../../output",
+            "skip-update": true
+} 
+            """)
+
+
+class MorelloRelease(Release):
+    target = "morello-release"
```
jrtc27, line 99: Normally we only put the architecture as a prefix when you're building a native tool, like morello-qemu (until merged into the main CHERI QEMU) or morello-llvm; when you're producing artifacts to run on that architecture it's instead a suffix.

pycheribuild/projects/release.py
jrtc27: (answer to the first part: if loaded from a config file, relative to the file, otherwise if on the command line it's relative to the current working directory)

PR #232, opened 2021-08-20
Update to latest Ubuntu and tools.  Add a docker build template.

gvnn3:
A more narrow update just for docker related files.  This includes a simple Dockerfile that can be used by Linux users to layer on more users once the base image is installed.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,8 @@
-FROM ubuntu:18.04
+FROM ubuntu:20.04
 
 LABEL maintainer="Alexander.Richardson@cl.cam.ac.uk"
+
+ARG DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && apt-get install -y \
   make ninja-build \
@@ -9,7 +11,9 @@ RUN apt-get update && apt-get install -y \
   python3-minimal \
   lsb-release \
   wget \
-  samba
+  samba \
+  texlive-base \
+  texinfo
 
 # RUN git config --global http.sslVerify false
 # RUN cd /tmp && git clone https://github.com/arichardson/bmake && cd bmake \
@@ -20,17 +24,12 @@ COPY cheribuild.json /root/.config/cheribuild.json
 
 # deps to build QEMU+elftoolchain:
 RUN apt-get update && apt-get install -y \
-  libtool pkg-config python-minimal autotools-dev automake autoconf libglib2.0-dev libpixman-1-dev \
+  libtool pkg-config python2-minimal autotools-dev automake autoconf libglib2.0-dev libpixman-1-dev \
   bison groff-base libarchive-dev flex
 
-# INSTALL clang 8.0
-RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
-RUN echo "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-8 main" > /etc/apt/sources.list.d/llvm.list
-# RUN echo "deb http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu trusty main" > /etc/apt/sources.list.d/r-toolchain.list
-# RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 1E9377A2BA9EF27F
-RUN apt-get update && apt-get install -y clang-8 lld-8
-
 RUN apt-get update && apt-get install -y cmake
+
+RUN addgroup --gid 12345 cheri && adduser --uid 12345 --ingroup cheri cheri
 
 VOLUME ["/cheribuild", "/source", "/build", "/output"]
 ENV PATH /cheribuild:$PATH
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -20,18 +24,13 @@ COPY cheribuild.json /root/.config/cheribuild.json
 
 # deps to build QEMU+elftoolchain:
 RUN apt-get update && apt-get install -y \
-  libtool pkg-config python-minimal autotools-dev automake autoconf libglib2.0-dev libpixman-1-dev \
+  libtool pkg-config python2-minimal autotools-dev automake autoconf libglib2.0-dev libpixman-1-dev \
```
jrtc27, line 27: Is python2 actually needed?

docker/Dockerfile
```diff
@@ -20,18 +24,13 @@ COPY cheribuild.json /root/.config/cheribuild.json
...
   bison groff-base libarchive-dev flex
 
-# INSTALL clang 8.0
```
jrtc27, line 26: Should still install clang+lld, will make building native things much faster

docker/docker-cheribuild.json
```diff
@@ -0,0 +1,7 @@
+# NOTE: Change the paths below to match the locations for your
+# sources, build and output directories.
+{
+  "source-root": "/home/USER/MYDIR/sources",
+  "build-root": "/home/USER/MYDIR/build",
+  "output-root": "/home/USER/MYDIR/output"
+}
```
jrtc27, line 7: None of this should be needed? The only thing you need is for (docker)-cheribuild outside of Docker to know where source-root is, which is just `{ "source-root": ".." }`, but that file needs to go in the top-level cheribuild directory to be picked up, not here. Everything else should follow automatically if you use a standard cheribuild layout rather than changing things to match your personal preferences.

docker/Dockerfile
```diff
@@ -20,18 +24,13 @@ COPY cheribuild.json /root/.config/cheribuild.json
...
-RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
-RUN echo "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-8 main" > /etc/apt/sources.list.d/llvm.list
-# RUN echo "deb http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu trusty main" > /etc/apt/sources.list.d/r-toolchain.list
-# RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 1E9377A2BA9EF27F
-RUN apt-get update && apt-get install -y clang-8 lld-8
-
 RUN apt-get update && apt-get install -y cmake
 
+RUN addgroup --gid 12345 cheri && adduser --uid 12345 --ingroup cheri cheri
```
jrtc27, line 32: And if someone has UID/GID 12345 how do they use Dockerfile.useradd without conflicts?

docker/Dockerfile.useradd
```diff
@@ -0,0 +1,7 @@
+FROM cheribuild-test
```
jrtc27, line 1: We should be using a different name to this for the release

docker/Dockerfile.useradd
```diff
@@ -0,0 +1,7 @@
...
+
+# NOTE: Change USER to your username, GROUPID and USERID to the output
+# of the id(1) command.
+ 
+RUN addgroup --gid GROUPID  USER && adduser --uid USERID --ingroup USER USER
+
```
jrtc27, line 7:
```suggestion
```

docker/Dockerfile.useradd
```diff
@@ -0,0 +1,7 @@
+FROM cheribuild-test
+
+# NOTE: Change USER to your username, GROUPID and USERID to the output
+# of the id(1) command.
+ 
+RUN addgroup --gid GROUPID  USER && adduser --uid USERID --ingroup USER USER
```
jrtc27, line 6:
```suggestion
RUN addgroup --gid GROUPID USER && adduser --uid USERID --ingroup USER USER
```

docker/Dockerfile
arichardson: It used to be required for qemu but they migrated to py3 a while ago. I don't think it's required anymore but I could be wrong.

gvnn3: I'll test removing  the python2 and push if that all "just works."

docker/Dockerfile
nwf: I recognize that this might be adding more moving parts, and so the opposite of what we want to do, but... https://github.com/containers/fuse-overlayfs has support for remapping [UG]IDs seen on the backing store(s) to different [UG]IDs as seen by the mount.  Since the in-guest vnops use the synthetic UIDs reported by the overlayfs, there's no need to synchronize UIDs inside and outside the container, and it can be made to work either with or without UID namespaces.  I think this means we could establish a mount with in-container 12345 transparently mapped to the out-of-container user on any volumes that get mounted?

Perhaps worth a look, perhaps excessive, or perhaps missing the point entirely?

PR #233, opened 2021-08-23
Create an adduser command

gvnn3:
The add user command adds the current user to the docker image by creating a special Dockerfile.adduser, executing a build, and then tagging the result back to the original name.

### DIFF
--- a/pycheribuild/projects/adduser.py
+++ b/pycheribuild/projects/adduser.py
@@ -1,0 +1,83 @@
+#
+# Copyright (c) 2021 George V. Neville-Neil
+# All rights reserved.
+#
+# This software was developed by SRI International and the University of
+# Cambridge Computer Laboratory under DARPA/AFRL contract FA8750-10-C-0237
+# ("CTSRD"), as part of the DARPA CRASH research programme.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+# OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+# OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+# SUCH DAMAGE.
+#
+
+import os
+import getpass
+import subprocess
+import sys
+
+from pathlib import Path
+from .project import CheriConfig, SimpleProject, DefaultInstallDir
+from ..utils import OSInfo, status_update
+from ..processutils import (get_program_version, print_command, run_and_kill_children_on_exit, run_command)
+
+class AddUser(SimpleProject):
+    target = "adduser"
+    native_install_dir = DefaultInstallDir.BOOTSTRAP_TOOLS
+
+    def __init__(self, config: CheriConfig):
+        super().__init__(config)
+
+    def process(self):
+        file = str(self.config.output_root.absolute()) + "/Dockerfile.adduser"
+        user = getpass.getuser()
+        try:
+            fp = open(file, mode = "w")
+            contents = "FROM cheribuild-docker\n\nRUN addgroup --gid " + \
+                     str(os.getgid()) + " " + user + \
+                     " && adduser --uid " + str(os.getuid()) + " --ingroup " + \
+                     user + " " + user + " \n"
+
+            fp.write(contents)
+
+            fp.close()
+
+            try:
+                docker_run_cmd = ["docker", "build", "-f", file, "."]
+            
+                run_command(docker_run_cmd, config=self.config, give_tty_control=True)
+                os.remove(file)
+                
+            except subprocess.CalledProcessError as e:
+                # if the image is missing print a helpful error message:
+                if e.returncode == 125:
+                    status_update("It seems like the docker image", config.docker_container, "was not found.")
+                    status_update("In order to build the default docker image for cheribuild (cheribuild-test) run:")
+                    print(
+                        coloured(AnsiColour.blue, "cd", cheribuild_dir + "/docker && docker build --tag cheribuild-docker ."))
+                    sys.exit(coloured(AnsiColour.red, "Failed to start docker!"))
+                    raise
+                sys.exit()
+
+        except:
+            status_update("Could not create ", file)
+            raise
+        sys.exit()
+
--- a/pycheribuild/projects/adduser.py
+++ b/pycheribuild/projects/adduser.py
@@ -48,6 +48,9 @@ class AddUser(SimpleProject):
     def process(self):
         file = str(self.config.output_root.absolute()) + "/Dockerfile.adduser"
         user = getpass.getuser()
+        output = ""
+
+        # Create a Dockerfile that will contain this user's name, gid, uid
         try:
             fp = open(file, mode = "w")
             contents = "FROM cheribuild-docker\n\nRUN addgroup --gid " + \
@@ -59,25 +62,41 @@ class AddUser(SimpleProject):
 
             fp.close()
 
-            try:
-                docker_run_cmd = ["docker", "build", "-f", file, "."]
-            
-                run_command(docker_run_cmd, config=self.config, give_tty_control=True)
-                os.remove(file)
-                
-            except subprocess.CalledProcessError as e:
-                # if the image is missing print a helpful error message:
-                if e.returncode == 125:
-                    status_update("It seems like the docker image", config.docker_container, "was not found.")
-                    status_update("In order to build the default docker image for cheribuild (cheribuild-test) run:")
-                    print(
-                        coloured(AnsiColour.blue, "cd", cheribuild_dir + "/docker && docker build --tag cheribuild-docker ."))
-                    sys.exit(coloured(AnsiColour.red, "Failed to start docker!"))
-                    raise
-                sys.exit()
-
         except:
             status_update("Could not create ", file)
-            raise
+            sys.exit()
+
+        # Build a new image from our installed image with this user
+        try:
+            docker_run_cmd = ["docker", "build", "-f", file, "."]
+        
+            output = run_command(docker_run_cmd, config=self.config, give_tty_control=True, capture_output=True)
+
+            os.remove(file)
+            
+        except subprocess.CalledProcessError as e:
+            os.remove(file) # Clean up after ourselves
+
+            # if the image is missing print a helpful error message:
+            if e.returncode == 125:
+                status_update("It seems like the docker image", config.docker_container, "was not found.")
+                status_update("In order to build the default docker image for cheribuild (cheribuild-test) run:")
+                print(
+                    coloured(AnsiColour.blue, "cd", cheribuild_dir + "/docker && docker build --tag cheribuild-docker ."))
+                sys.exit(coloured(AnsiColour.red, "Failed to start docker!"))
+                raise
+            sys.exit()
+
+        # Take the new image and retag it to the original name
+        try:
+            result = str(output.stdout)
+            image = result[result.rfind(" ") + 1:-3] # -3 gets the \n off
+
+            docker_run_cmd = ["docker", "image", "tag", image, "cheribuild-docker"]
+            output = run_command(docker_run_cmd, config=self.config, give_tty_control=True, capture_output=True)
+
+        except:
+                sys.exit(coloured(AnsiColour.red, "Failed to retag docker!"))
+
         sys.exit()
 


pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,102 @@
...
+
+            os.remove(file)
+            
+        except subprocess.CalledProcessError as e:
+            os.remove(file) # Clean up after ourselves
+
+            # if the image is missing print a helpful error message:
+            if e.returncode == 125:
+                status_update("It seems like the docker image", config.docker_container, "was not found.")
+                status_update("In order to build the default docker image for cheribuild (cheribuild-test) run:")
```
jrtc27, line 83: Name doesn't match elsewhere

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,102 @@
...
+
+        except:
+            status_update("Could not create ", file)
+            sys.exit()
+
+        # Build a new image from our installed image with this user
+        try:
+            docker_run_cmd = ["docker", "build", "-f", file, "."]
+        
+            output = run_command(docker_run_cmd, config=self.config, give_tty_control=True, capture_output=True)
```
jrtc27, line 73: Do these actually need control of the TTY?

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,102 @@
...
+                     " && adduser --uid " + str(os.getuid()) + " --ingroup " + \
+                     user + " " + user + " \n"
+
+            fp.write(contents)
+
+            fp.close()
+
+        except:
+            status_update("Could not create ", file)
+            sys.exit()
```
jrtc27, line 67: Use self.fatal

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,102 @@
...
+            fp = open(file, mode = "w")
+            contents = "FROM cheribuild-docker\n\nRUN addgroup --gid " + \
+                     str(os.getgid()) + " " + user + \
+                     " && adduser --uid " + str(os.getuid()) + " --ingroup " + \
+                     user + " " + user + " \n"
+
+            fp.write(contents)
+
+            fp.close()
+
```
jrtc27, line 64: Why all the blank lines?

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,102 @@
...
+    target = "adduser"
+    native_install_dir = DefaultInstallDir.BOOTSTRAP_TOOLS
+
+    def __init__(self, config: CheriConfig):
+        super().__init__(config)
+
+    def process(self):
+        file = str(self.config.output_root.absolute()) + "/Dockerfile.adduser"
+        user = getpass.getuser()
+        output = ""
```
jrtc27, line 51: Use None, or waive the lint error, not sure what the preferred approach is

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,102 @@
...
+
+        # Create a Dockerfile that will contain this user's name, gid, uid
+        try:
+            fp = open(file, mode = "w")
```
jrtc27, line 55: `with open(...) as fp:` (or `f` would be more normal in Python, certainly what we tend to use in cheribuild, this isn't C)?

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,102 @@
...
+
+class AddUser(SimpleProject):
+    target = "adduser"
+    native_install_dir = DefaultInstallDir.BOOTSTRAP_TOOLS
+
+    def __init__(self, config: CheriConfig):
+        super().__init__(config)
+
+    def process(self):
+        file = str(self.config.output_root.absolute()) + "/Dockerfile.adduser"
```
jrtc27, line 49: I imagine build_root would make more sense

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,102 @@
...
+        # Build a new image from our installed image with this user
+        try:
+            docker_run_cmd = ["docker", "build", "-f", file, "."]
+        
+            output = run_command(docker_run_cmd, config=self.config, give_tty_control=True, capture_output=True)
+
+            os.remove(file)
+            
+        except subprocess.CalledProcessError as e:
+            os.remove(file) # Clean up after ourselves
```
jrtc27, line 78: Except you don't on other errors. But if you shove it in build_root it doesn't matter that it's left over, I wouldn't even bother deleting it in the success path, you're only ever going to have one as it's always the same name, not like tmp files where they accumulate.

pycheribuild/projects/adduser.py
jrtc27: Also, doesn't need to have the .adduser suffix if it's going in its own directory

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,102 @@
...
+
+            # if the image is missing print a helpful error message:
+            if e.returncode == 125:
+                status_update("It seems like the docker image", config.docker_container, "was not found.")
+                status_update("In order to build the default docker image for cheribuild (cheribuild-test) run:")
+                print(
+                    coloured(AnsiColour.blue, "cd", cheribuild_dir + "/docker && docker build --tag cheribuild-docker ."))
+                sys.exit(coloured(AnsiColour.red, "Failed to start docker!"))
```
jrtc27, line 86: This makes no sense. Use `self.fatal` with `fixit_hint=...`

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,102 @@
...
+                raise
```
jrtc27, line 87: This is dead

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,102 @@
...
+            sys.exit()
```
jrtc27, line 88: No error message. But should be using `self.fatal` anyway which does the non-zero exit for you.

pycheribuild/projects/adduser.py
jrtc27: sys.exit also defaults to exit code 0, which is wrong for error cases, but fatal will exit for you

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,102 @@
...
+
+        # Take the new image and retag it to the original name
+        try:
+            result = str(output.stdout)
+            image = result[result.rfind(" ") + 1:-3] # -3 gets the \n off
```
jrtc27, line 93: rstrip? Looking at the docs there are examples where the Successfully built and Removing intermediate container lines get swapped around, so I don't know how reliable this parsing is going to be, should at least be looking for a line that says Successfully built IMO. But why not just pass -t to the build itself to build and tag it in one go?

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,102 @@
...
+
+            docker_run_cmd = ["docker", "image", "tag", image, "cheribuild-docker"]
+            output = run_command(docker_run_cmd, config=self.config, give_tty_control=True, capture_output=True)
+
+        except:
+                sys.exit(coloured(AnsiColour.red, "Failed to retag docker!"))
```
jrtc27, line 99: This makes no sense

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,102 @@
...
+
+        sys.exit()
```
jrtc27, line 101: Absolutely not, targets should not stop other targets from running unless there is an error

jrtc27 (commented):
Having it apply to `cheribuild` and not `docker-cheribuild` is weird, I'd expect it to be specific to that as it makes no sense for `cheribuild` itself, and with a generic name like `adduser` when used with `cheribuild` it sounds like it's building something like https://salsa.debian.org/debian/adduser or messing with your local system.

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,91 @@
...
+        # Build a new image from our installed image with this user
+        try:
+            docker_run_cmd = ["docker", "build", "--tag=cheribuild-docker",
+                              "-f", file, "."]
+            run_command(docker_run_cmd, config=self.config)
+
+            os.remove(file)
+
+        except subprocess.CalledProcessError as e:
+            os.remove(file)
```
jrtc27, line 82: Old review comment still applies, just now hidden as you removed the inline code comment:
> Except you don't on other errors. But if you shove it in build_root it doesn't matter that it's left over, I wouldn't even bother deleting it in the success path, you're only ever going to have one as it's always the same name, not like tmp files where they accumulate.

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,91 @@
...
+
+            # if the image is missing print a helpful error message:
+            if e.returncode == 125:
+                cheribuild_dir = str(Path(__file__).absolute().parent.parent)
+                status_update("It seems like the docker image", self.config.docker_container, "was not found.")
+                status_update("In order to build the default docker image for cheribuild (cheribuild-docker) run:")
+                print(coloured(AnsiColour.blue, "cd", cheribuild_dir +
+                               "/docker && docker build --tag cheribuild-docker ."))
+                sys.exit(coloured(AnsiColour.red, "Failed to start docker!"))
```
jrtc27, line 91: Old comment still applies, just now hidden as you changed the whitespace:

> This makes no sense. Use `self.fatal` with `fixit_hint=...`

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,91 @@
...
+                              "-f", file, "."]
+            run_command(docker_run_cmd, config=self.config)
+
+            os.remove(file)
+
+        except subprocess.CalledProcessError as e:
+            os.remove(file)
+
+            # if the image is missing print a helpful error message:
+            if e.returncode == 125:
```
jrtc27, line 85: Still no error for if it's anything else

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def __init__(self, config: CheriConfig):
+        super().__init__(config)
+
+    def process(self):
+        file = str(self.config.output_root.absolute()) + "/Dockerfile.adduser"
+        try:
+            user = getpass.getuser()
+        except KeyError:
+            user = "nobody"
```
jrtc27, line 55: This is going to fall over if this gets hit without --pretend. This should probably use self.fatal too but then fall back on `user = "nobody"`, since fatal defaults to non-fatal when pretending.

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,91 @@
...
+
+        # Create a Dockerfile that will contain this user's name, gid, uid
+        try:
+            fp = open(file, mode="w")
```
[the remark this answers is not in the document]
jrtc27: Oh yeah and also this introduces side-effects for --pretend, that needs to not be the case (maybe we should have CI diff trees before and after running to catch that...)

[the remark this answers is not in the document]
jrtc27: Right, this whole try/except can just be `self.write_file(file, contents, overwrite=True)`, that handles --pretend automatically

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,91 @@
...
+            os.remove(file)
+
+        except subprocess.CalledProcessError as e:
+            os.remove(file)
+
+            # if the image is missing print a helpful error message:
+            if e.returncode == 125:
+                cheribuild_dir = str(Path(__file__).absolute().parent.parent)
+                status_update("It seems like the docker image", self.config.docker_container, "was not found.")
+                status_update("In order to build the default docker image for cheribuild (cheribuild-docker) run:")
```
jrtc27, line 88: Unconvinced by this; people using a release should be getting the release, not building it for themselves. I can guarantee someone will break something and hit this message and blindly follow it...

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,91 @@
...
+                print(coloured(AnsiColour.blue, "cd", cheribuild_dir +
+                               "/docker && docker build --tag cheribuild-docker ."))
```
jrtc27, line 90: Wouldn't `docker build --tag cheribuild-docker /path/to/cheribuild/docker` also work, no cd needed? But see above, not sure this should be reported to the user.

pycheribuild/projects/adduser.py
gvnn3: I gather you mean cheribuild-test vs. cheribuild-docker?  This will be cleaned up when both PRs are processed.

pycheribuild/projects/adduser.py
gvnn3: I like vertical space?

pycheribuild/projects/adduser.py
gvnn3: Funny, this is exactly the code in main that spits out the docker message.

pycheribuild/projects/adduser.py
gvnn3: These are the exact directlons from main.

pycheribuild/projects/adduser.py
gvnn3: Again this is what we already tell the user now.

jrtc27: That's because when that code was added there were no releases. This is all for a release so the context has changed.

pycheribuild/projects/adduser.py
jrtc27: As above. And copying code is a good time to review whether the original code from many years ago that was hacked together and lightly tested is the best way of doing it.

pycheribuild/projects/adduser.py
jrtc27: I stand corrected somewhat, apparently sys.exit, if given anything other than an int, will print that out and exit with code 1. Ew. But my point of self.fatal still stands; code in __main__.py is running outside the context of a target, so the interfaces, and expected behaviours, are very different (no project to call methods on, and if there's an error then it can only call sys.exit, whereas inside a target the expectation is that only the target fails).

pycheribuild/projects/adduser.py
jrtc27: Don't raise exceptions, use self.fatal

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,81 @@
...
+
+        # Build a new image from our installed image with this user
+        try:
+            docker_run_cmd = ["docker", "build", "--tag=cheribuild-docker",
+                              "-f", target_file, "."]
+            run_command(docker_run_cmd, config=self.config)
+
+        except subprocess.CalledProcessError as e:
+            # if the image is missing print a helpful error message:
+            if e.returncode == 1:
```
jrtc27, line 71: Is 1 correct? If so that seems an awfully generic code that likely means more than just this so can't say with certainty the image was missing?

gvnn3: That's the return code we get at that point if there is no image.  I tested it.

pycheribuild/projects/adduser.py
```diff
@@ -0,0 +1,81 @@
...
+            user = getpass.getuser()
+        except KeyError:
+            user = "nobody"
+
+        # Create a Dockerfile that will contain this user's name, gid, uid
+        contents = "FROM cheribuild-docker\n\nRUN addgroup --gid " + \
+                   str(os.getgid()) + " " + user + \
+                   " && adduser --uid " + str(os.getuid()) + " --ingroup " + \
+                   user + " " + user + " \n"
+        self.write_file(target_file, contents, overwrite=True)
```
arichardson, line 61: I think this should work and is IMO a bit more readable:
```suggestion
        contents = 
        self.write_file(target_file, overwrite=True, contents=f"""
FROM cheribuild-docker
RUN addgroup --gid {os.getgid()} {user} && \
    adduser --uid {os.getuid()} --ingroup {user} {user}
""")
```

arichardson:
I feel like this  should be part of the docker logic in __main__.py (https://github.com/CTSRD-CHERI/cheribuild/blob/c6d0c171acb7bfcca32ef43854a8214c88d2c6c2/pycheribuild/__main__.py#L189) if the users didn't explicitly specify a docker image.

PR #236, opened 2021-08-30
qemu_utils: probe for disk image format

nwf:
xref https://github.com/CTSRD-CHERI/cheri-docker-images/pull/4

### DIFF
--- a/pycheribuild/qemu_utils.py
+++ b/pycheribuild/qemu_utils.py
@@ -89,13 +89,20 @@ class QemuOptions:
             raise ValueError("Unknown target " + str(xtarget))
 
     def disk_image_args(self, image) -> list:
+        # Probe to see if this is qcow2 or not.  If it isn't, we assume it's raw, which isn't exercising the full range
+        # of qemu's features but we haven't, as yet, generated anything else.  We don't let qemu probe because it warns
+        # loudly if probing detects a raw image.
+        with open(image, 'rb') as imgf:
+            magic = imgf.read(4)
+            img_fmt = "qcow2" if magic == b'QFI\xfb' else "raw"
+
         if self.virtio_disk:
             # RISC-V doesn't support virtio-blk-pci, we have to use virtio-blk-device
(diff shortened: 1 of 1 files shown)


jrtc27:
This will probably upset --pretend... I do feel like we should probably have .img and .qcow2 files and have it be a property of the disk-image target

jrtc27:
Yep there you go, Jenkins has just confirmed that

nwf:
OK.  There's already `--disk-image/use-qcow2`; I'll figure out how to hook that instead.

pycheribuild/qemu_utils.py
```diff
@@ -89,13 +89,20 @@ def __init__(self, xtarget: CrossCompileTarget, want_debugger=False):
             raise ValueError("Unknown target " + str(xtarget))
 
     def disk_image_args(self, image) -> list:
+        # Probe to see if this is qcow2 or not.  If it isn't, we assume it's raw, which isn't exercising the full range
+        # of qemu's features but we haven't, as yet, generated anything else.  We don't let qemu probe because it warns
+        # loudly if probing detects a raw image.
+        with open(image, 'rb') as imgf:
```
arichardson, line 95: I think probing is still useful even with #237. --pretend can easily be fixed by checking for .exists().

arichardson: i.e. something like:
```python
if self.config.pretend and not image.exists():
     img_fmt = "qcow2" if image.name.endswith(".qcow2") else "raw"
else:
    with image.open('rb') as imgf:
        magic = imgf.read(4)
        img_fmt = "qcow2" if magic == b'QFI\xfb' else "raw"
```

nwf: Instead of the format parameter or if the parameter is `None` or to cross-check the parameter or...?

arichardson: I was thinking this could beb useful if someone explicitly overwrites the disk image path on the command line, when using the defaults the format argument should be sufficient.

Reading 4 bytes as a sanity check unconditionally should be fine for the default case if that makes the logic simpler?

PR #237, opened 2021-08-30
Better support for qcow2 disk images

nwf:
See #236 and https://github.com/CTSRD-CHERI/cheri-docker-images/pull/4 .  Many thanks to @jrtc27 for all the suggestions so far.

### DIFF
--- a/pycheribuild/projects/disk_image.py
+++ b/pycheribuild/projects/disk_image.py
@@ -706,6 +706,7 @@ class BuildDiskImageBase(SimpleProject):
             self.run_cmd(qemu_img_command, "convert",
                          "-f", "raw",  # input file is in raw format (not required as QEMU can detect it
                          "-O", "qcow2",  # convert to qcow2 format
+                         "-c", "-o", "compression_type=zstd",  # compress data inside the image
                          raw_img,  # input file
                          self.disk_image_path)  # output file
             self.delete_file(raw_img, print_verbose_only=True)
--- a/pycheribuild/projects/disk_image.py
+++ b/pycheribuild/projects/disk_image.py
@@ -100,8 +100,12 @@ class _AdditionalFileTemplates(object):
 
 def _default_disk_image_name(config: CheriConfig, directory: Path, project: "BuildDiskImageBase"):
     xtarget = project.get_crosscompile_target(config)
+    if project.use_qcow2:
+        suffix = "qcow2"
+    else:
+        suffix = "img"
     # Don't add the os_prefix to the disk image name since it should already be encoded in project.disk_image_prefix)
-    return directory / (project.disk_image_prefix + project.build_configuration_suffix(xtarget) + ".img")
+    return directory / (project.disk_image_prefix + project.build_configuration_suffix(xtarget) + "." + suffix)
 
 
 def _default_disk_image_hostname(prefix: str) -> "ComputedDefaultValue[str]":
@@ -523,9 +527,9 @@ class BuildDiskImageBase(SimpleProject):
             return True
         return False
 
-    def make_x86_disk_image(self):
+    def make_x86_disk_image(self, out_img: Path):
         assert self.is_x86
-        root_partition = self.disk_image_path.with_suffix(".root.img")
+        root_partition = out_img.with_suffix(".root.img")
         try:
             self.make_rootfs_image(root_partition)
             # See mk_nogeli_gpt_ufs_legacy in tools/boot/rootgen.sh in FreeBSD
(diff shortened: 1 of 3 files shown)


pycheribuild/projects/disk_image.py
```diff
@@ -676,24 +680,29 @@ def make_disk_image(self):
...
+            self.run_cmd(qemu_img_command, "info", raw_img)
         if self.use_qcow2:
             if not qemu_img_command.exists():
                 self.fatal("Cannot create QCOW2 image without qemu-img command!")
             # create a qcow2 version from the raw image:
-            raw_img = self.disk_image_path.with_suffix(".raw")
-            self.run_cmd("mv", "-f", self.disk_image_path, raw_img)
             self.run_cmd(qemu_img_command, "convert",
                          "-f", "raw",  # input file is in raw format (not required as QEMU can detect it
                          "-O", "qcow2",  # convert to qcow2 format
```
arichardson, line 708: Do we want to use zstd compression here?

nwf: If you'd rather that be here than in the docker goo, I'm happy to do it here.  I was contemplating `--disk-image/use-qcow2=zstd` or something, and so hadn't added it yet, but I'm also OK just always compressing qcow2 images.

arichardson: I guess disk I/O get's a bit slower but I doubt very much that that is limiting our CheriBSD on QEMU performance. Compression also appears to be very quick so I don't think there is a reason not to do it by default? If you don't want it you can always use raw images.

nwf: OK, I'm sold.

pycheribuild/qemu_utils.py
```diff
@@ -89,6 +89,21 @@ def __init__(self, xtarget: CrossCompileTarget, want_debugger=False):
...
+                image_format = "qcow2" if image.name.endswith(".qcow2") else "raw"
+        else:
+            with image.open('rb') as imgf:
+                magic = imgf.read(4)
+                is_qcow2 = magic == b'QFI\xfb'
+
+                if image_format is None:
+                    image_format = "qcow2" if is_qcow2 else "raw"
+                elif is_qcow2 and image_format != "qcow2":
+                    self.warning("Disk image looks like qcow2 but image_format is %s" % image_format)
```
nwf, line 105: Oh, hm, no, this not being a `Project` means there isn't a `self.warning` either (oh, for static typing).  Suggestions, @arichardson?

pycheribuild/qemu_utils.py
```diff
@@ -88,14 +88,29 @@ def __init__(self, xtarget: CrossCompileTarget, want_debugger=False):
...
+                image_format = "qcow2" if image.name.endswith(".qcow2") else "raw"
+        else:
+            with image.open('rb') as imgf:
+                magic = imgf.read(4)
+                is_qcow2 = magic == b'QFI\xfb'
+
+                if image_format is None:
+                    image_format = "qcow2" if is_qcow2 else "raw"
+                elif is_qcow2 and image_format != "qcow2":
+                    warning_message("Disk image looks like qcow2 but image_format is %s" % image_format)
```
arichardson, line 105:
```suggestion
                    warning_message("Disk image looks like qcow2 but image_format is", image_format)
```

arichardson (approved):

pycheribuild/qemu_utils.py
```diff
@@ -88,14 +88,29 @@ def __init__(self, xtarget: CrossCompileTarget, want_debugger=False):
...
+                image_format = "qcow2" if image.name.endswith(".qcow2") else "raw"
+        else:
+            with image.open('rb') as imgf:
+                magic = imgf.read(4)
+                is_qcow2 = magic == b'QFI\xfb'
+
+                if image_format is None:
+                    image_format = "qcow2" if is_qcow2 else "raw"
+                elif is_qcow2 and image_format != "qcow2":
+                    warning_message("Disk image looks like qcow2 but image_format is", image_format)
```
jrtc27, line 105: Should also warn for the `not is_qcow2 and image_format == "qcow2"` case, i.e. emit a warning whenever `is_qcow2 != (image_format == "qcow2")`

pycheribuild/qemu_utils.py
```diff
@@ -88,14 +88,29 @@ def __init__(self, xtarget: CrossCompileTarget, want_debugger=False):
         else:
             raise ValueError("Unknown target " + str(xtarget))
 
-    def disk_image_args(self, image) -> list:
+    def disk_image_args(self, image: Path, image_format: str) -> list:
+        # Probe the disk image format in case someone has overridden the default image path or format is unspecified
+        if not image.exists():
+            # Either we're pretending or we'll complain elsewhere.
```
jrtc27, line 94: Hm, seems we don't have a config to hand to assert that we are pretending? Should otherwise have been checked already.

nwf: Yea, `QemuOptions` isn't a subtype of `Project`.  @arichardson mentioned that other things use the `QemuOptions` type, too, and so this seemed fine?

arichardson: I made it separate so it can be reused by the pexpect test scripts. I guess we could have a config base class that just has pretend?

nwf: Is it OK to merge as written and defer refactoring? :)

arichardson (approved):

PR #239, opened 2021-09-01
Revert qemu zstd

nwf:


pycheribuild/projects/disk_image.py
```diff
@@ -706,7 +706,6 @@ def make_disk_image(self):
             self.run_cmd(qemu_img_command, "convert",
                          "-f", "raw",  # input file is in raw format (not required as QEMU can detect it
                          "-O", "qcow2",  # convert to qcow2 format
-                         "-c", "-o", "compression_type=zstd",  # compress data inside the image
```
jrtc27, line 709: Do we want to keep -c though?

nwf: I could go either way, but my thought was

> We can't rely on zstd availability, and zlib is just a little uncomfortably slow while compressing the whole image, so go back to not compressing qcow2 images.

If you'd rather we keep `-c`, I could be convinced.

jrtc27: Yeah, for local use that probably makes sense, and we're converting manually for the Docker image

PR #241, opened 2021-09-07
docker image: switch to a new non-root user when started as root

arichardson:
Instead of manually adding a matching user with the docker-adduser
cheribuild target, this adds an ENTRYPOINT script to the docker image
that automatically creates an unprivileged user. The UID/GID/name can
be passed using environment variables (-e flag to `docker run`).


docker/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
...
+# Create a non-root user with UID/GID matching the host user to ensure that
+# files written to the volumes are not owned by root.
+: "${cheribuild_uid:=1234}"
+: "${cheribuild_gid:=1234}"
+: "${cheribuild_user:=cheri}"
+addgroup --quiet --gid ${cheribuild_gid} "${cheribuild_user}"
+yes | adduser --quiet --uid ${cheribuild_uid} --disabled-password --ingroup "${cheribuild_user}" "${cheribuild_user}" > /dev/null
+
+# Run the actual command:
+export HOME="/home/${cheribuild_user}"
```
jrtc27, line 16: Need to copy root's cheribuild.json to their home directory otherwise they probably can't read it?

docker/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
...
+    echo "Already running as non-root, can't change user."
+    exec "$@"
+fi
+# Create a non-root user with UID/GID matching the host user to ensure that
+# files written to the volumes are not owned by root.
+: "${cheribuild_uid:=1234}"
+: "${cheribuild_gid:=1234}"
+: "${cheribuild_user:=cheri}"
+addgroup --quiet --gid ${cheribuild_gid} "${cheribuild_user}"
+yes | adduser --quiet --uid ${cheribuild_uid} --disabled-password --ingroup "${cheribuild_user}" "${cheribuild_user}" > /dev/null
```
jrtc27, line 13: What's the yes for?

The --disabled-password is fine, unless we want to put the user in the sudo group, at which point they'll need a password (blank generally easiest in this kind of situation...). I don't think you can set that with adduser, so you'd either need to set their password as an extra step or switch to using the lower-level useradd (and I guess groupadd for symmetry) commands.

docker/entrypoint.sh
arichardson: Oh yeah forgot about that.

docker/entrypoint.sh
arichardson: For the `Is the information correct? [Y/n]` prompt.

However it seems like setting stdin to /dev/null also works (but prints warnings):
```
docker run -it ubuntu:latest
root@3094ac014bb2:/# addgroup --quiet --gid 501 alex && adduser --uid 501 --disabled-password --ingroup alex alex < /dev/null
Adding user `alex' ...
Adding new user `alex' (501) with group `alex' ...
Creating home directory `/home/alex' ...
Copying files from `/etc/skel' ...
Changing the user information for alex
Enter the new value, or press ENTER for the default
	Full Name []: 	Room Number []: 	Work Phone []: 	Home Phone []: Other []: Use of uninitialized value $answer in chop at /usr/sbin/adduser line 621.
Use of uninitialized value $answer in pattern match (m//) at /usr/sbin/adduser line 622.
Is the information correct? [Y/n]
```

arichardson: Changed to use useradd

arichardson:
I believe this should now be correct.

jrtc27:
My only concern with merging this is that the initial release tarball (which we know some people are using) failed to check out a tag for cheribuild, so it tracks master and prompts people to update, meaning we can't actually make the cheribuild side of the necessary changes to support this because that will break with people's existing docker images. I don't know how we deal with that though because it'll need breaking at _some_ point, and other changes to cheribuild may stop things working with the release's snapshot of repos before then...

arichardson:
Yes that is unfortunate. However, these dockerfile changes should be backwards compatible since they do nothing when `--uid` is passed to docker.

I'll look into a workaround for the release when I actually make the docker commandline change.

PR #243, opened 2021-09-22
Rework the Basic usage section to be more explicit about ./

gvnn3:
An attempt to make it very clear how to use the cheribuild command based on comments from @rmn30

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,9 +41,22 @@ dnf install libtool clang-devel bison cmake ninja-build samba flex texinfo glib2
 
 # Basic usage
 
-If you want to start up a QEMU VM running CheriBSD run `cheribuild.py run-riscv64-purecap -d` (-d means build all dependencies).
-If you would like the VM to have all userspace binaries to be built as plain RISC-V binaries instead of CHERI pure-capability ones use `cheribuild.py run-riscv64-hybrid -d`.
-This will build the CHERI compiler, QEMU, CheriBSD, create a disk image and boot that in QEMU.
+The `cheribuild.py` program is normally run from within the
+`cheribuild/` directory as `./cheribuild.py` with options following
+the main command.  Please note the leading `./` before the command name.
+
+To build and run a QEMU VM running CheriBSD on top of the CHERI-RISCV
+ISA run the following command:
+
+`./cheribuild.py run-riscv64-purecap -d` (-d means build all dependencies).
+
+This command will build the CHERI compiler, QEMU, CheriBSD, create a
+disk image and boot that in QEMU.
+
+To build and run the same set of tools and binaries for the Morello
+ISA use the following command:
+
+`./cheribuild.py run-morello-purecap -d`
 
 By default `cheribuild.py` will clone all projects in `~/cheri`, use `~/cheri/build` for build directories
 and install into `~/cheri/output`. However, these directories are all configurable (see below for details).
--- a/README.md
+++ b/README.md
@@ -41,11 +41,14 @@ dnf install libtool clang-devel bison cmake ninja-build samba flex texinfo glib2
 
 # Basic usage
 
-The `cheribuild.py` program is normally run from within the
+In these examples the `cheribuild.py` program is run from within the
 `cheribuild/` directory as `./cheribuild.py` with options following
-the main command.  Please note the leading `./` before the command name.
+the main command.  Please note the leading `./` before the command
+name.  It is also possible to add the `cheribuild/` directory to your
+`PATH` or to link the binary somewhere such as in a `bin/` diretory of
+your choosing.
 
-To build and run a QEMU VM running CheriBSD on top of the CHERI-RISCV
+If you want to start up a QEMU VM running CheriBSD on top of the CHERI-RISCV
 ISA run the following command:
 
 `./cheribuild.py run-riscv64-purecap -d` (-d means build all dependencies).
@@ -53,8 +56,8 @@ ISA run the following command:
 This command will build the CHERI compiler, QEMU, CheriBSD, create a
 disk image and boot that in QEMU.
 
-To build and run the same set of tools and binaries for the Morello
-ISA use the following command:
+If you want to start up a QEMU VM running CheriBSD on top of the
+Morello ISA use the following command:
 
 `./cheribuild.py run-morello-purecap -d`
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,9 +41,22 @@ dnf install libtool clang-devel bison cmake ninja-build samba flex texinfo glib2
 
 # Basic usage
 
-If you want to start up a QEMU VM running CheriBSD run `cheribuild.py run-riscv64-purecap -d` (-d means build all dependencies).
-If you would like the VM to have all userspace binaries to be built as plain RISC-V binaries instead of CHERI pure-capability ones use `cheribuild.py run-riscv64-hybrid -d`.
-This will build the CHERI compiler, QEMU, CheriBSD, create a disk image and boot that in QEMU.
+The `cheribuild.py` program is normally run from within the
```
arichardson, line 44: Maybe we should just add something like
```
The commandlines given below assume that you have added cheribuild.py to your $PATH, if not you will have to change to the cheribuild directory and use `./cheribuild`?
```

arichardson (commented):
Not quite sure about the `./` since I didn't use it anywhere in the rest of the README. The bit changing the riscv64-hybrid example LGTM.

README.md
gvnn3: Does anyone add cheribuild to their path?  All the examples I see imply going in there and using ./ but I'm flexible on what we say.

jrtc27 (commented):
*sigh* unsubmitted comments from yesterday

README.md
jrtc27: I do, pretty sure Alex does, I think Alfredo does too. IMO it's the best way, I have ~/bin/cheribuild (and other prefixed ones that automatically pick up separate configs) symlinked to ~/cheri/cheribuild/cheribuild.py so I can just `cheribuild foo` wherever whenever without having to think.

jrtc27: I think the main reason the examples all use `./` on the website is because it removes a step and thus is a quicker onboarding experience (and gives us the "clone and run, that's it" marketing), but it's not the best experience long-term.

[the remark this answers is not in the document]
gvnn3: We can always do the PATH thing as @arichardson suggests above.  I happy to always cd into the cheribuild I want because I often have several at any given moment, but that may not be typical of our downstream users.

README.md
```diff
@@ -41,9 +41,22 @@ dnf install libtool clang-devel bison cmake ninja-build samba flex texinfo glib2
 
 # Basic usage
 
-If you want to start up a QEMU VM running CheriBSD run `cheribuild.py run-riscv64-purecap -d` (-d means build all dependencies).
```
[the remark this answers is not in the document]
gvnn3: I made this change so that the two commands given match stylistically in how they are presented from an English language point of view.  In the previous version we did not mention Morello and adding that caused me to reword this little section.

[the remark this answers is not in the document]
jrtc27: Could just word the addition to match the existing style though? "If you would like the VM to use Arm's Morello ISA instead of CHERI-RISC-V use `cheribuild.py run-morello-purecap -d`."

jrtc27:
> The bit changing the riscv64-hybrid example LGTM.

Agreed; that was both wrong (they're not plain RISC-V binaries) and encouraging people too much to use hybrid

README.md
brooksdavis: I symlink ~/cheri/cheribuild/cheribuild.py to ~/bin/cheribuild

README.md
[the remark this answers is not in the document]
gvnn3: I find it too passive, but, sure we can make it match.

README.md
arichardson: I've also got a symlink to ~/bin (and a debug-cheribuild.py one that automatically selects debug-cheribuild.json)

README.md
```diff
@@ -41,9 +41,25 @@ dnf install libtool clang-devel bison cmake ninja-build samba flex texinfo glib2
...
-If you would like the VM to have all userspace binaries to be built as plain RISC-V binaries instead of CHERI pure-capability ones use `cheribuild.py run-riscv64-hybrid -d`.
-This will build the CHERI compiler, QEMU, CheriBSD, create a disk image and boot that in QEMU.
+In these examples the `cheribuild.py` program is run from within the
+`cheribuild/` directory as `./cheribuild.py` with options following
+the main command.  Please note the leading `./` before the command
+name.  It is also possible to add the `cheribuild/` directory to your
+`PATH` or to link the binary somewhere such as in a `bin/` diretory of
+your choosing.
```
jrtc27, line 49: Thanks, this new wording looks good to me; anyone who has a clue about Unix will read this and do what they want, and those who don't know what it means can ignore it and blindly follow the basic approach

README.md
```diff
@@ -41,9 +41,25 @@ dnf install libtool clang-devel bison cmake ninja-build samba flex texinfo glib2
...
+
+If you want to start up a QEMU VM running CheriBSD on top of the CHERI-RISCV
```
jrtc27, line 51:
```suggestion
If you want to start up a QEMU VM running CheriBSD on top of the CHERI-RISC-V
```

arichardson (approved):

README.md
```diff
@@ -41,9 +41,25 @@ dnf install libtool clang-devel bison cmake ninja-build samba flex texinfo glib2
...
 # Basic usage
 
-If you want to start up a QEMU VM running CheriBSD run `cheribuild.py run-riscv64-purecap -d` (-d means build all dependencies).
-If you would like the VM to have all userspace binaries to be built as plain RISC-V binaries instead of CHERI pure-capability ones use `cheribuild.py run-riscv64-hybrid -d`.
-This will build the CHERI compiler, QEMU, CheriBSD, create a disk image and boot that in QEMU.
+In these examples the `cheribuild.py` program is run from within the
+`cheribuild/` directory as `./cheribuild.py` with options following
+the main command.  Please note the leading `./` before the command
+name.  It is also possible to add the `cheribuild/` directory to your
+`PATH` or to link the binary somewhere such as in a `bin/` diretory of
```
jrtc27, line 48:
```suggestion
`PATH` or to link the binary somewhere such as in a `bin/` directory of
```

PR #245, opened 2021-10-06
Fix Cherios

LawrenceEsswood:
Change the branches to use to build / run cherios, and fix arguments for running QEMU.

### DIFF
--- a/pycheribuild/qemu_utils.py
+++ b/pycheribuild/qemu_utils.py
@@ -109,9 +109,10 @@ class QemuOptions:
 
         if self.virtio_disk:
             # RISC-V doesn't support virtio-blk-pci, we have to use virtio-blk-device
-            device_kind = "virtio-blk-device" if self.xtarget.is_riscv(include_purecap=True) or \
-                                                 self.force_virtio_blk_device                   \
-                                              else "virtio-blk-pci"
+            if self.xtarget.is_riscv(include_purecap=True) or self.force_virtio_blk_device:
+                device_kind = "virtio-blk-device"
+            else:
+                device_kind = "virtio-blk-pci"
             return ["-drive", "if=none,file=" + str(image) + ",id=drv,format=" + image_format,
                     "-device", device_kind + ",drive=drv"]
         else:


pycheribuild/projects/build_qemu.py
```diff
@@ -338,7 +338,7 @@ def qemu_cheri_binary(cls, caller: SimpleProject, xtarget: CrossCompileTarget =
 
 
 class BuildCheriOSQEMU(BuildQEMU):
-    repository = GitRepository("https://github.com/CTSRD-CHERI/qemu.git", default_branch="cherios", force_branch=True)
+    repository = GitRepository("https://github.com/CTSRD-CHERI/qemu.git", default_branch="qemu-cheri", force_branch=True)
```
jrtc27, line 341: Should be able to use the normal CHERI QEMU build now and remove this target?

LawrenceEsswood: Gonna have it float around for now because I might point it to the morello branch tomorrow so I can play around with cherios-morello. But yes, soon!

arichardson (approved):

pycheribuild/qemu_utils.py
```diff
@@ -108,7 +109,9 @@ def disk_image_args(self, image: Path, image_format: str) -> list:
 
         if self.virtio_disk:
             # RISC-V doesn't support virtio-blk-pci, we have to use virtio-blk-device
-            device_kind = "virtio-blk-device" if self.xtarget.is_riscv(include_purecap=True) else "virtio-blk-pci"
+            device_kind = "virtio-blk-device" if self.xtarget.is_riscv(include_purecap=True) or \
+                                                 self.force_virtio_blk_device                   \
+                                              else "virtio-blk-pci"
```
arichardson, line 114: Not keen on backslash line continuations, maybe just convert it to an if?
```suggestion
            if self.xtarget.is_riscv(include_purecap=True) or self.force_virtio_blk_device:
                device_kind = "virtio-blk-device"
            else:
                device_kind = "virtio-blk-pci"
```

Although I think we can hopefully drop the RISC-V special case soon? @jrtc27 ?

In that case the following should look better:

```suggestion
            device_kind = "virtio-blk-device" if self.force_virtio_blk_device else "virtio-blk-pci"
```

jrtc27: Yes, fixed as of CTSRD-CHERI/cheribsd@fb7c6d17d9dedc30b2ad7ea4ba186c966fa756e8

PR #249, opened 2021-10-15
remove --enable-slirp=git option from build.

gwynethsc:
No longer valid option; slirp enabled by default?
(note that cheri/cherios-qemu/configure defaults to slirp=”yes”)

Encountered when attempting to build run-cherios target clean.


jrtc27:
And cherios-qemu doesn't even _exist_ as a target in that branch...

LawrenceEsswood:
As a of a couple weeks ago, some of the CheriOS branches were quite stale. I would advise doing a `cheribuild cherios -d --clean` and say to change every time it suggests you are on the wrong branch.

PR #253, opened 2021-11-01
Support RISCV CheriOS

LawrenceEsswood:
Add a CheriOS RISCV target cross compile target and have it be supported for cherios and run-cherios targets. I also deleted the CheriOS qemu target because it is no longer needed.

### DIFF
--- a/pycheribuild/projects/cherios.py
+++ b/pycheribuild/projects/cherios.py
@@ -38,17 +38,16 @@ class BuildCheriOS(CMakeProject):
     default_build_type = BuildType.DEBUG
     repository = GitRepository("https://github.com/CTSRD-CHERI/cherios.git", default_branch="master")
     _default_install_dir_fn = ComputedDefaultValue(
-        function=lambda config, cls: config.output_root / ("cherios" + config.mips_cheri_bits_str),
+        function=lambda config, cls: config.output_root / ("cherios" + cls._xtarget.build_suffix(config,include_os=False)),
         as_string="<OUTPUT_ROOT>/cherios128")
     needs_sysroot = False
-    supported_architectures = [CompilationTargets.CHERIOS_MIPS_PURECAP]
+    supported_architectures = [CompilationTargets.CHERIOS_MIPS_PURECAP, CompilationTargets.CHERIOS_RISCV_PURECAP]
 
     @classmethod
     def setup_config_options(cls, **kwargs):
         super().setup_config_options(**kwargs)
         cls.smp_cores = cls.add_config_option("smp-cores", default=1, kind=int)
         cls.build_net = cls.add_bool_option("build-net", default=False)
-
     def __init__(self, config: CheriConfig):
         super().__init__(config)
         self.add_cmake_options(CHERI_SDK_DIR=self.target_info.sdk_root_dir)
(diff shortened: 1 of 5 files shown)


pycheribuild/projects/cherios.py
```diff
@@ -38,10 +38,11 @@ class BuildCheriOS(CMakeProject):
     default_build_type = BuildType.DEBUG
     repository = GitRepository("https://github.com/CTSRD-CHERI/cherios.git", default_branch="master")
     _default_install_dir_fn = ComputedDefaultValue(
-        function=lambda config, cls: config.output_root / ("cherios" + config.mips_cheri_bits_str),
+        function=lambda config, cls: config.output_root / ("cherios" + cls._xtarget.build_suffix(config,
+                                                                                                 include_os=False)),
         as_string="<OUTPUT_ROOT>/cherios128")
```
jrtc27, line 43: Outdated

pycheribuild/projects/cross/llvm.py
```diff
@@ -589,7 +589,7 @@ class BuildCheriOSLLVM(BuildLLVMMonoRepoBase):
     hide_options_from_help = True
 
     def configure(self, **kwargs):
-        self.add_cmake_options(LLVM_TARGETS_TO_BUILD="Mips;host")
+        self.add_cmake_options(LLVM_TARGETS_TO_BUILD="RISCV;Mips;host")
```
jrtc27, line 592: Sort

pycheribuild/projects/run_qemu.py
```diff
@@ -621,7 +621,7 @@ def __init__(self, config: CheriConfig):
 
     def setup(self):
         super().setup()
-        self.qemu_binary = BuildCheriOSQEMU.qemu_binary(self)
+        self.qemu_binary = BuildQEMU.qemu_cheri_binary(self)
```
jrtc27, line 624: This is the default

pycheribuild/projects/cross/llvm.py
LawrenceEsswood: Alphabetically?

pycheribuild/projects/cherios.py
LawrenceEsswood: What is preferred?

pycheribuild/projects/cross/llvm.py
jrtc27: Yes, though host is always at the end in the file (and I note that for Morello LLVM we put ARM before AArch64... booo)

pycheribuild/projects/cherios.py
jrtc27: Something like `cherios-{mips64,riscv64}` which I believe are the two values you'll get (though check that...)

jrtc27: Though as you're changing this line, switch it to $OUTPUT_ROOT, the <...> syntax is mostly phased out

LawrenceEsswood: Ah wait, are we talking about the outdated cherios128? cherios-mips and cherios-riscv64 are the two values I otherwise get.

jrtc27: Yeah. And I see, that's because CHERIOS_MIPS_PURECAP never got updated to have its suffix be mips64 rather than mips like all the others did. Should probably just fix that.

LawrenceEsswood: Updated the suffix

jrtc27 (approved):

PR #254, opened 2021-11-01
openssl: enable non-fett build

brooksdavis:
Update Configure arguments to work with 1.1.1 and rework support for
multiple versions.

Move disabling parallel builds to the fett target where 1.0.2 is used
and has a broken Makefile.

Fix header install location for fett openssl.

This is a reimplementation of 25aa9947e4288cfa9801d42ef76ba182aa64a297
using tuples rather than the packaging version type.

### DIFF
--- a/pycheribuild/projects/cross/openssl.py
+++ b/pycheribuild/projects/cross/openssl.py
@@ -34,14 +34,19 @@ from .crosscompileproject import (CrossCompileProject, DefaultInstallDir, FettPr
 
 
 class BuildOpenSSL(CrossCompileProject):
-    # Just add the FETT target below for now.
-    do_not_add_to_targets = True
     build_in_source_dir = True
+    # nginx looks in /usr/local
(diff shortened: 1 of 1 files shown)


pycheribuild/projects/cross/openssl.py
```diff
@@ -34,15 +34,19 @@
 
 
 class BuildOpenSSL(CrossCompileProject):
-    # Just add the FETT target below for now.
-    do_not_add_to_targets = True
     build_in_source_dir = True
+    path_in_rootfs = "/usr/local"
```
jrtc27, line 38: Could benefit from a comment why the default doesn't work (namely that nginx's home-grown configure scripts suck...)

jrtc27:
Looks fine other than the desire for a comment

jrtc27:
Your force push doesn't seem to be any different to the original commit

jrtc27 (approved):
Thanks

PR #257, opened 2021-11-04
Add jsc tests

brettferdosi:
Fix webkit build for new versions of icu and add a test runner for sunspider benchmarks.


gvnn3 (approved):
LGTM

PR #260, opened 2021-11-11
device-model fixes

bukinr:
addressing jrtc27's comments


pycheribuild/projects/disk_image.py
```diff
@@ -1425,6 +1426,12 @@ class BuildFreeBSDWithDefaultOptionsDiskImage(BuildFreeBSDImage):
     hide_options_from_help = True
 
 
+class BuildCheriBSDDeviceModelDiskImage(BuildCheriBSDDiskImage):
+    target = "disk-image-cheribsd-device-model"
```
jrtc27, line 1430: CheriBSD is a bit special in that if no OS is mentioned it's implied to be CheriBSD, so disk-image targets don't mention it; i.e. this should just be disk-image-device-model (which then also matches what your run target is).

arichardson:
This has been open for many years without changes, closing it now.

PR #261, opened 2021-11-16
CheriOS not using bbl

LawrenceEsswood:
Bootloaders just get in the way.

### DIFF
--- a/pycheribuild/projects/run_qemu.py
+++ b/pycheribuild/projects/run_qemu.py
@@ -626,6 +626,10 @@ class LaunchCheriOSQEMU(LaunchQEMUBase):
                 self.run_cmd("dd", "if=/dev/zero", "of=" + str(self.disk_image), size_flag, "count=1")
         super().process()
 
+    def get_riscv_bios_args(self) -> typing.List[str]:
+        # CheriOS bundles its kernel with its own bootloader
+        return ["-bios", "none"]
+
 
 class LaunchRtemsQEMU(LaunchQEMUBase):
     target = "run-rtems"


jrtc27:
No, you definitely want to have BBL. For QEMU you can get away with it, for real hardware you absolutely cannot.

jrtc27:
(The supervisor binary interface is _the_ abstraction that sits between an OS and everything beneath it, and like Arm's EL3 doesn't exist in a virtualised environment)

LawrenceEsswood:
So, what if I want my nanokernel to occupy that space?

jrtc27:
You don't, because otherwise you need one nanokernel per hardware platform, and have to reimplement all of that platform's firmware, and are unable to run in a virtualised environment, only an emulated one.

jrtc27:
The only time M-mode gets used for the actual OS is when you're writing an RTOS for a system that has no S-mode. If you're using S-mode then M-mode is for the platform's firmware, not you.

LawrenceEsswood:
I think the nanokernel _is_ a part of a particular platform's firmware. I don't imagine it would be possible to have one binary that would fit all. I'm happy for the CheriOS kernel to sit in S-mode.

jrtc27:
The nanokernel is an extension of your OS down towards the hardware, tightly coupled to CheriOS's interface and requirements. Firmware that lives in M-mode is an extension of the hardware that provides a generic abstraction for any OS to sit above.

jrtc27:
So, taking the Morello board as an example, how would you make CheriOS's nanokernel work there, when TF-A has all manner of platform-specific stuff to do (configuring all the control bits specific to that core, implementing errata workarounds, etc.)? Do you really intend to ship your own board firmware just for CheriOS, rather than make it work without EL3 access?

LawrenceEsswood:
So, I suppose there are a couple of things. The first is TCB minimisation. I want to sit the nanokernel as low as possible to reduce that. The second is how the nanokernel is shipped. Within reason, it shouldn't be possible to run a given platform without a nanokernel. A user should be able to verify they are running on a nanokernel-fied platform (the nanokernel contains some keys in a similar way that SGX enabled cores have a key embedded). The nanokernel really feels like it belongs as a part of trusted firmware. I'm not sure I want to (naively) statically link the nanokernel into existing firmware (it would make the TCG larger than I want it to be), but nor do I want to re-implement everything existing firmware already does. Possibly, the nanokernel should be a cheri-enforced compartment within existing firmware (the only one with the access system permission), and then I would just modify TF-A / bbl etc to use that.

PR #262, opened 2021-11-16
CheriBSD main

bsdjhb:


pycheribuild/projects/cross/gdb.py
```diff
@@ -240,7 +240,10 @@ class BuildGDB(BuildGDBBase):
     repository = GitRepository(
         "https://github.com/CTSRD-CHERI/gdb.git",
         # Branch name is changed for every major GDB release:
-        default_branch="mips_cheri-8.3", force_branch=True,
+        default_branch="mips_cheri-8.3",
+        old_branches={"mips_cheri_7.12": "mips_cheri-8.3",
```
bsdjhb, line 244: It would be handy to use self.default_branch as the value in the dictionary perhaps.

jrtc27: I guess you could assign special meaning to None as self.default_branch

pycheribuild/projects/project.py
```diff
@@ -1770,24 +1780,22 @@ def update(self, current_project: "Project", *, src_dir: Path, base_project_sour
...
-            if status.stdout.startswith(b"## ") and not status.stdout.startswith(
-                    b"## " + default_branch.encode("utf-8") + b"..."):
-                current_branch = status.stdout[3:status.stdout.find(b"...")].strip()
-                current_project.warning("You are trying to build the", current_branch.decode("utf-8"),
+        if src_dir.exists() and (self.force_branch or self.old_branches):
+            if self.force_branch:
+                assert self.old_branches is None, "cannot set both force_branch and old_branches"
+                default_branch = self.get_default_branch(current_project, include_per_target=True)
+                assert default_branch, "default_branch must be set if force_branch is true!"
+            current_branch = self.get_current_branch(src_dir).decode("utf-8")
```
jrtc27, line 1788: It deviates more from the original code, but maybe hoist this above the force_branch case? Then you have the two cases setting default_branch grouped together

bsdjhb: Yeah, I almost did that and honestly prefer it that way as well.

pycheribuild/projects/project.py
```diff
@@ -1770,24 +1780,22 @@ def update(self, current_project: "Project", *, src_dir: Path, base_project_sour
...
+            if self.old_branches:
+                default_branch = self.old_branches.get(current_branch)
```
jrtc27, line 1790: This won't work if current_branch isn't in old_branches; it won't raise an exception like `[]`, instead it'll give None and blow up later. Easiest solution is to add current_branch as the optional default value argument.

pycheribuild/projects/cross/gdb.py
bsdjhb: Hmm, that's an interesting idea, but that seems kind of odd?  It also makes it a bit more of a pain to deal with in the implementation.  The other option I've thought about for the gdb case is just using a local default_branch variable in this function.

pycheribuild/projects/project.py
jrtc27: Oh, never mind, you already check for None... though if you want to have `"foo": None` mean `self._default_branch` then you do need to be able to distinguish the cases

bsdjhb: I check for None below (I purposely didn't want an exception)

bsdjhb: Yeah, this is why using None to mean default branch is a bit messier.

pycheribuild/projects/cross/gdb.py
jrtc27: Yeah that's less gross

jrtc27 (approved):
(with the obvious caveat of "don't land the final commit until we're ready"; others can go in earlier so it's just the CheriBSD change pending)

jrtc27:
(GitHub doesn't update the diff / list of commits unless you change the base...)

PR #266, opened 2021-12-23
Add a recipe for Neomutt command line email client

tmarkettos:
This adds basic support for building the Neomutt email client.  Some nice-to-have functionality is not enabled (header caches, SASL, NLS, international domain names) but as-is it's basically usable for talking to IMAP/SMTP servers.

Tested with Fastmail with the following config flags to avoid SASL:
```
set smtp_authenticators="login"
set ssl_starttls = yes
set ssl_force_tls = yes
```

### DIFF
--- a/pycheribuild/projects/cross/neomutt.py
+++ b/pycheribuild/projects/cross/neomutt.py
@@ -24,10 +24,8 @@
 #
 
 from .crosscompileproject import CrossCompileAutotoolsProject
-from .crosscompileproject import CheriConfig, CompilationTargets, CrossCompileProject
+from .crosscompileproject import CheriConfig, CompilationTargets
 from ..project import GitRepository
-from ...config.compilation_targets import CompilationTargets
-
 
 class BuildNeomutt(CrossCompileAutotoolsProject):
     repository = GitRepository("https://github.com/neomutt/neomutt.git")
@@ -44,4 +42,4 @@ class BuildNeomutt(CrossCompileAutotoolsProject):
 
         # enable OpenSSL (in base system), disable internationalisation libs we don't have
         self.configure_args.extend(['--disable-nls', '--disable-idn',
-          '--disable-doc', '--ssl'])
+            '--disable-doc', '--ssl'])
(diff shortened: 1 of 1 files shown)


pycheribuild/projects/cross/neomutt.py
```diff
@@ -0,0 +1,47 @@
...
+
+from .crosscompileproject import CrossCompileAutotoolsProject
+from .crosscompileproject import CheriConfig, CompilationTargets, CrossCompileProject
+from ..project import GitRepository
+from ...config.compilation_targets import CompilationTargets
+
+
+class BuildNeomutt(CrossCompileAutotoolsProject):
+    repository = GitRepository("https://github.com/neomutt/neomutt.git")
+    supported_architectures = CompilationTargets.ALL_FREEBSD_AND_CHERIBSD_TARGETS + [CompilationTargets.NATIVE]
```
jrtc27, line 33: Eh, I'd just leave it as the default, which is CheriBSD+native, unless you really need it to also build for FreeBSD? Arguably the default should include FreeBSD, but it adds a whole load more targets that slows things down and clutters everything up.

pycheribuild/projects/cross/neomutt.py
```diff
@@ -0,0 +1,47 @@
...
+    dependencies = ['gettext', 'libxml2']
+
+    def __init__(self, config: CheriConfig):
+        super().__init__(config)
```
jrtc27, line 37:
```suggestion
    def setup(self):
        super().setup(config)
```
so this gets deferred until the target is actually built

pycheribuild/projects/cross/neomutt.py
```diff
@@ -0,0 +1,47 @@
...
+
+        # neomutt's build system doesn't use autotools, it justs pretends to look the same
+        # - but it doesn't implement the --target option, so we strip it
+        indices = [i for i, s in enumerate(self.configure_args) if '--target=' in s]
```
jrtc27, line 41: I've been wanting to ditch --target by default as you don't normally specify it when cross-compiling, and it can cause issues with some other targets (specifically, htop will triple-prefix the binaries by default if you do that, even if host == target, due to some part of autotools it's using). Might make more sense to at least make it an opt-out class property, but this is fine for now.

Though this would be better written as:

```
self.configure_args[:] = [arg for arg in self.configure_args if not arg.startswith('--target=')]
```

PR #267, opened 2022-01-12
Add gkermit cross project

nwf:
Serial ports are eternal.

### DIFF
--- a/pycheribuild/projects/cross/gkermit.py
+++ b/pycheribuild/projects/cross/gkermit.py
@@ -1,0 +1,61 @@
+#
+# Copyright (c) 2022 Microsoft Corporation
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+# OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+# OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+# SUCH DAMAGE.
+
+from pathlib import Path
+
+from ..project import ExternallyManagedSourceRepository
+from .crosscompileproject import CrossCompileMakefileProject, DefaultInstallDir
+
+
+class BuildGKermit(CrossCompileMakefileProject):
+    build_in_source_dir = False
+    build_via_symlink_farm = True
+    native_install_dir = DefaultInstallDir.BOOTSTRAP_TOOLS
+    repository = ExternallyManagedSourceRepository()
+    set_commands_on_cmdline = True
+
+    def setup(self):
+        super().setup()
+        self.common_warning_flags.append("-Wno-unused-value")
+        self.common_warning_flags.append("-Wno-non-literal-null-conversion")
+        self.make_args.set_env(KFLAGS=self.commandline_to_str(
+            self.default_compiler_flags + ["-include", "unistd.h"]))
+
+    def update(self):
+        filename = "gku201.tar.gz"
+        sha256 = "19f9ac00d7b230d0a841928a25676269363c2925afc23e62704cde516fc1abbd"
+        url_prefix = "https://www.kermitproject.org/ftp/kermit/archives/"
+
+        self.makedirs(self.source_dir)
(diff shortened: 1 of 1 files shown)


pycheribuild/projects/cross/gkermit.py
```diff
@@ -0,0 +1,63 @@
...
+            self.run_cmd(["tar", "xzvf", self.source_dir / filename, "-C", self.source_dir])
+
+    def needs_configure(self):
+        return not (self.build_dir / "makefile").exists()
+
+    def configure(self, **kwargs):
+        if not self.should_run_configure():
+            return
+
+        self.write_file(self.build_dir / "makefile", overwrite=True, mode=0o644, contents=f"""
```
jrtc27, line 55: What's wrong with upstream's Makefile?

pycheribuild/projects/cross/gkermit.py
```diff
@@ -0,0 +1,63 @@
+#
+# Copyright (c) 2021 Microsoft
```
jrtc27, line 2: 2022?
And surely it's more than just Microsoft? (Microsoft Corporation?)

pycheribuild/projects/cross/gkermit.py
```diff
@@ -0,0 +1,63 @@
...
+# OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+# SUCH DAMAGE.
+
+from ..project import ExternallyManagedSourceRepository
+from .crosscompileproject import CrossCompileMakefileProject, DefaultInstallDir
+
+
+class BuildGKermit(CrossCompileMakefileProject):
+    build_in_source_dir = False
+    cross_install_dir = DefaultInstallDir.ROOTFS_OPTBASE
```
jrtc27, line 31: What's wrong with localbase?

pycheribuild/projects/cross/gkermit.py
```diff
@@ -0,0 +1,63 @@
...
+    native_install_dir = DefaultInstallDir.BOOTSTRAP_TOOLS
+    repository = ExternallyManagedSourceRepository()
+
+    def clean(self):
+        return self.async_clean_directory(self.build_dir, keep_root=True)
```
jrtc27, line 36: This is the default?

pycheribuild/projects/cross/gkermit.py
nwf: It expects in-source-tree builds and to be edited to set `CC`.

pycheribuild/projects/cross/gkermit.py
nwf: Didn't know which was preferred.  localbase it is, then.

pycheribuild/projects/cross/gkermit.py
jrtc27: `build_via_symlink_farm = True` deals with the former (it has its own issues, but those shouldn't matter for something like kermit that's not frequently changing), `make CC=foo` deals with the latter.

pycheribuild/projects/cross/gkermit.py
jrtc27: The preferred is "leave it at the default" :)

pycheribuild/projects/cross/gkermit.py
nwf: Time keeps on slipping, doesn't it.  And yea, probably "Microsoft Corporation" though we don't appear to be perfectly consistent.

nwf:
Here's a spin using the existing infrastructure (`build_via_symlink_farm` and `set_commands_on_cmdline`) with just a little bit of local flavour.

pycheribuild/projects/cross/gkermit.py
```diff
@@ -0,0 +1,64 @@
...
+        filename = "gku201.tar.gz"
+        sha256 = "19f9ac00d7b230d0a841928a25676269363c2925afc23e62704cde516fc1abbd"
+        url_prefix = "https://www.kermitproject.org/ftp/kermit/archives/"
+
+        self.makedirs(self.source_dir)
+        if not (self.source_dir / filename).is_file():
+            self.download_file(self.source_dir / filename, url=url_prefix + filename, sha256=sha256)
+            self.run_cmd(["tar", "xzvf", self.source_dir / filename, "-C", self.source_dir])
+
+    def needs_configure(self):
```
jrtc27, line 55: The symlink farm creation is a configure step, so this breaks builds unless you use --clean or --reconfigure (presumably you tested with --clean). But this shouldn't be needed, Project has configure_command = None and MakefileProject doesn't override that, so it'll automatically do the right thing (i.e. nothing).

pycheribuild/projects/cross/gkermit.py
```diff
@@ -0,0 +1,64 @@
...
+        return False
+
+    def compile(self, cwd: Path = None, parallel: bool = True):
+        if cwd is None:
+            cwd = self.build_dir
+        self.run_make("gkermit", cwd=cwd, parallel=parallel)
+
+    def install(self, **kwargs):
+        self.install_file(self.build_dir / "gkermit", self.install_dir / "gkermit")
```
jrtc27, line 64: Missing a bin for the install dir

pycheribuild/projects/cross/gkermit.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def update(self):
+        filename = "gku201.tar.gz"
+        sha256 = "19f9ac00d7b230d0a841928a25676269363c2925afc23e62704cde516fc1abbd"
+        url_prefix = "https://www.kermitproject.org/ftp/kermit/archives/"
+
+        self.makedirs(self.source_dir)
+        if not (self.source_dir / filename).is_file():
+            self.download_file(self.source_dir / filename, url=url_prefix + filename, sha256=sha256)
+            self.run_cmd(["tar", "xzvf", self.source_dir / filename, "-C", self.source_dir])
```
jrtc27, line 53: Technically this command should be gated on a file in the tarball in case it gets interrupted between downloading and extracting (either by external events or by something going wrong)

nwf: Ah, and `download_file` uses the checksum for idempotence.  Great.

pycheribuild/projects/cross/gkermit.py
```diff
@@ -0,0 +1,61 @@
...
+    native_install_dir = DefaultInstallDir.BOOTSTRAP_TOOLS
+    repository = ExternallyManagedSourceRepository()
+    set_commands_on_cmdline = True
+
+    def setup(self):
+        super().setup()
+        self.common_warning_flags.append("-Wno-unused-value")
+        self.common_warning_flags.append("-Wno-non-literal-null-conversion")
+        self.make_args.set_env(KFLAGS=self.commandline_to_str(
+            self.default_compiler_flags + ["-include", "string.h", "-include", "unistd.h"]))
```
jrtc27, line 43: string.h is in gkermit.h, it's just unistd.h that's missing? (which Debian carries a local patch for)

nwf: Oh, huh.  I guess my notes about building probably date back to pre-2.0.

jrtc27 (approved):
Can't obviously spot any other issues

PR #269, opened 2022-01-28
Perform python version check earlier

LawrenceEsswood:
Thought it would be nice to just have a single file where we bump the version to the maximum of all the requirements we have (that can safely be included first), for at least a couple of the top-level scripts. I left a more permissive noqa rather than noqa: E402 on some of the imports because nothing is actually imported and this would fail another flake8 check.

### DIFF
--- a/tests/test_metalog.py
+++ b/tests/test_metalog.py
@@ -12,6 +12,7 @@ except ImportError:
 from pathlib import Path
 
 sys.path.append(str(Path(__file__).parent.parent))
+import pycheribuild.check_version  # noqa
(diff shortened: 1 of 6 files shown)


tests/test_metalog.py
```diff
@@ -12,6 +12,7 @@
 from pathlib import Path
 
 sys.path.append(str(Path(__file__).parent.parent))
+import pycheribuild.check_version  # noqa
```
arichardson, line 15: We probably don't need this for the tests, but adding it to the main entry point seems reasonable.

arichardson:
No longer needed.

PR #270, opened 2022-02-01
Trust src.opt.mk

bsdjhb:
I have a similar change to stop forcing OFED off, but CheriBSD isn't ready for that yet.


jrtc27:
Should also do the same with kernel modules at some point... hard-coding mlx* as disabled isn't an issue but linuxulator and armv8crypto should really be done solely by CheriBSD

brooksdavis (approved):

PR #271, opened 2022-02-07
Simplify handling of linux/input.h dependency

arichardson:
Instead of adding the libinput source directory as an include path, we can
create linux/{input,input-event-codes,uinput}.h headers that simply
#include the corresponding dev/evdev/ header.

The evdev-proto port downloads the Linux headers and patches those instead,
but it seems to me that creating a file that includes the native header is
less fragile as it doesn't rely on ioctl() numbers being compatible, etc.

According to https://bugreports.qt.io/browse/QTBUG-79130 and
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=240964#c19 the evdev
headers sould not be used outside of the base system, but I'm not
convinced that using the Linux kernel definitions is any better?


arichardson:
This should be fine according to freebsd-xorg IRC.

PR #277, opened 2022-03-03
Disable vectorization for benchmarks

arichardson:
This ensures that we have a more comparable baseline that just compares
CHERI overheads rather than highlighting that some bits of LLVM still need
to be updated. Interestingly clang enables vectorization for
`-fno-vectorize -O2/-O3`, but not for `-O2/-O3 -fno-vectorize`. This means
we have to override the default CMAKE_{C,CXX}_FLAGS variable since
otherwise -O3 -DNDEBUG will be appended after the flags that cheribuild
sets and thereby overrides the custom optimization_flags.


bsdjhb:
Do we want to make it easier to undo this in the future or for specific runs?  I guess that means, do we want a new command line knob to toggle this.  Perhaps not, but I worry that if in the future the toolchain is updated to do this better then we'd still have this hanging around.

arichardson:
> Do we want to make it easier to undo this in the future or for specific runs? I guess that means, do we want a new command line knob to toggle this. Perhaps not, but I worry that if in the future the toolchain is updated to do this better then we'd still have this hanging around.

That is a good point - maybe this should be opt-in rather than opt-out. If it's opt-in I could also move it out of the BenchmarkMixin and have it apply to all projects instead (probably makes sense to build CheriBSD without vectorization as well although there probably isn't too much that changes).

arichardson:
It doesn't make much sense to merge this.

PR #278, opened 2022-03-04
fett: remove all support for fett targets

brooksdavis:
The FETT bug bounty is long over and users who want to run on AWS F1
instances are better served by GFE targets which will remain.

Enable targets for non-FETT variants of things like sqlbox that
otherwise would have no targets. It may make sense to GC some of these
targets entierly.


jrtc27:
The kernel configs are useful to keep since the GFE kernels are MFS ones and the QEMU kernels are linked at the wrong address. All the other crud can go though.

pycheribuild/projects/cross/openssl.py
```diff
@@ -57,25 +57,8 @@ def setup(self):
         self.configure_args.append("BSD-generic64")
         self.configure_args.append("-shared")
 
-        if self.get_version() <= (1, 1, 0):
```
arichardson, line 60: get_version can be removed too?

arichardson (approved):
Generally LGTM, I can rebase my MIPS removal on top of this (or merge that first if you prefer).

PR #281, opened 2022-03-23
Install optional configuration with --subdir.

kwitaszczyk:
When using --subdir, we should execute install and installconfig tragets to install optional configuration.

Note that installconfig doesn't fail if there is no configuration in a subdirectory.


arichardson (approved):

jrtc27:
s/tragets/targets/ in the commit message

PR #282, opened 2022-03-25
CheriBSD/Morello supports MFS-ROOT kernels, too

nwf:


pycheribuild/projects/cross/cheribsd.py
```diff
@@ -1759,7 +1766,7 @@ class BuildCheriBsdMfsKernel(BuildCHERIBSD):
     target = "cheribsd-mfs-root-kernel"
     dependencies = ["disk-image-mfs-root"]
     repository = ReuseOtherProjectRepository(source_project=BuildCHERIBSD, do_update=True)
-    supported_architectures = CompilationTargets.ALL_CHERIBSD_RISCV_TARGETS
+    supported_architectures = CompilationTargets.ALL_CHERIBSD_TARGETS_WITH_HYBRID
```
jrtc27, line 1769: This includes amd64 which doesn't have an MFS root config committed

nwf: Thanks.  Is `ALL_CHERIBSD_RISCV_TARGETS + ALL_CHERIBSD_MORELLO_TARGETS` the right thing to use?

jrtc27: That doesn't include plain AArch64 :)

pycheribuild/projects/disk_image.py
```diff
@@ -975,7 +975,7 @@ def add_unlisted_files_to_metalog(self):
         self.mtree.add_dir("var/db", print_status=self.config.verbose)
         self.mtree.add_dir("var/empty", print_status=self.config.verbose)
 
-        if self.is_x86 or self.compiling_for_aarch64(include_purecap=True):
+        if self.include_boot and (self.is_x86 or self.compiling_for_aarch64(include_purecap=True)):
```
jrtc27, line 978: This does mean that kgdb and dtrace will be sadder (though without the .full kernel maybe one or both already are), but that's not why this code was here, it's here based on whether you're direct-booting kernels or not, so that's a separate issue.

nwf: At a glance, `/boot/kernel/kernel` does contain a `.SUNW_ctf` and `.symtab` section, so I could see a reason to want those even on some `self.include_boot = False` builds.  Sounds like the MFS disk image classes might want a build flag.

arichardson (approved):
Seems reasonable to me, @jrtc27 are you happy with this?

jrtc27:
Hm, include_boot_more isn't a great name, and the include_boot_kernel dance is a bit odd in the subclasses (plus non-default MFS config builds will embed the wrong kernel... but I guess that's an existing issue?)

nwf:
Yeah, that's fair; I'll rename it to the probably better `include_boot_files`.  As to which kernel gets included, that is, indeed, an existing problem and probably worth fixing as I would also like to add support for `NODEBUG` kernels.

arichardson (approved):

PR #285, opened 2022-04-11
Don't ship {src, ports}.txz; expect users to check them out.

rwatson:


pycheribuild/projects/cross/cheribsd.py
```diff
@@ -1912,6 +1912,11 @@ def process(self):
         release_args = self.installworld_args.copy()
         release_args.update(self.kernel_make_args_for_config(kernconf, None).copy())
 
+        # Don't build src.txz or ports.txz into our releases.  We prefer users
+        # check them out properly, or use a tool like portsnap.
+        release_args.set(NOSRC=True)
+        release_args.set(NOPORTS=True)
```
jrtc27, line 1918: ports.txz is more complicated, it's not just a case of turning it back on, it's also a case of checking out the right thing and setting PORTSDIR to point to it, not picking up whatever's in /usr/ports if you happen to be building on FreeBSD and doing nothing otherwise

PR #287, opened 2022-04-13
Add rootfs-tarball targets

nwf:
These are rather like disk images, only without the filesystem,
partitions, or disks.  They are particularly useful as sources of NFS
exports for netbooted, diskless machines.

### DIFF
--- a/pycheribuild/projects/disk_image.py
+++ b/pycheribuild/projects/disk_image.py
@@ -1374,6 +1374,40 @@ class BuildCheriBSDDiskImage(BuildDiskImageBase):
         self.minimum_image_size = "256m"  # let's try to shrink the image size
 
 
+def _default_tar_name(config: CheriConfig, directory: Path, project: "BuildDiskImageBase"):
+    xtarget = project.get_crosscompile_target(config)
+    return directory / (project.disk_image_prefix + project.build_configuration_suffix(xtarget) + ".tar.xz")
+
+
+class BuildCheriBSDTarball(BuildCheriBSDDiskImage):
+    target = "rootfs-tarball"
+
+    default_disk_image_path = ComputedDefaultValue(
+        function=lambda conf, proj: _default_tar_name(conf, conf.output_root, proj),
+        as_string=lambda cls: "$OUTPUT_ROOT/" + cls.disk_image_prefix + "-<TARGET>.tar.xz depending on architecture")
+
+    def __init__(self, config: CheriConfig):
+        super().__init__(config)
+        self.add_required_system_tool("bsdtar", cheribuild_target="bsdtar", apt="libarchive-tools")
+
+    def make_disk_image(self):
+        # write out the manifest file:
+        self.mtree.write(self.manifest_file, pretend=self.config.pretend)
+        bsdtar_path = shutil.which("bsdtar")
+        if not bsdtar_path:
+            if not self.config.pretend:
+                raise LookupError("Could not find bsdtar command in PATH")
+            bsdtar_path = "bsdtar"
+        try:
+            self.run_cmd([bsdtar_path, "acf", self.disk_image_path, "@"+str(self.manifest_file)], cwd=self.rootfs_dir)
+        except Exception:
+            self.warning("bsdtar failed, if it reports an issue with METALOG report a bug (could be either cheribuild"
+                         " or cheribsd) and attach the METALOG file.")
+            self.query_yes_no("About to delete the temporary directory. Copy any files you need before pressing enter.",
+                              yes_no_str="")
+            raise
+
+
 class BuildFreeBSDImage(BuildDiskImageBase):
     target = "disk-image-freebsd"
     include_os_in_target_suffix = False


pycheribuild/projects/disk_image.py
```diff
@@ -818,7 +834,10 @@ def __process(self):
             # Add/symlink GDB (if requested).
             self.add_gdb()
             # finally create the disk image
-            self.make_disk_image()
+            if self.is_tarball:
```
jrtc27, line 837: You could just override make_disk_image in the subclass?

pycheribuild/projects/disk_image.py
```diff
@@ -642,6 +643,21 @@ def make_efi_partition(self, efi_partition: Path):
...
+    def make_rootfs_tarball(self):
+        # write out the manifest file:
+        self.mtree.write(self.manifest_file, pretend=self.config.pretend)
+        try:
+            bsdtar_path = shutil.which(str(self.bsdtar_cmd))
+            if not bsdtar_path:
+                bsdtar_path = str(self.bsdtar_cmd)
+            self.run_cmd([bsdtar_path, "cf", self.disk_image_path, "@"+str(self.manifest_file)], cwd=self.rootfs_dir)
+        except Exception:
+            self.warning("makefs failed, if it reports an issue with METALOG report a bug (could be either cheribuild"
```
jrtc27, line 655: Not makefs

pycheribuild/projects/disk_image.py
```diff
@@ -642,6 +643,21 @@ def make_efi_partition(self, efi_partition: Path):
                 self.run_cmd(mtools_bin / "mdir", "-i", efi_partition, "-/", "::")
                 # self.run_cmd(mtools_bin / "mdu", "-i", efi_partition, "-a", "::")
 
+    def make_rootfs_tarball(self):
+        # write out the manifest file:
+        self.mtree.write(self.manifest_file, pretend=self.config.pretend)
+        try:
+            bsdtar_path = shutil.which(str(self.bsdtar_cmd))
+            if not bsdtar_path:
```
jrtc27, line 651: Should probably be a fatal error for !pretending?

pycheribuild/projects/disk_image.py
```diff
@@ -642,6 +643,21 @@ def make_efi_partition(self, efi_partition: Path):
...
+                bsdtar_path = str(self.bsdtar_cmd)
+            self.run_cmd([bsdtar_path, "cf", self.disk_image_path, "@"+str(self.manifest_file)], cwd=self.rootfs_dir)
```
jrtc27, line 653: Is it worth compressing? Maybe as an option?

pycheribuild/projects/disk_image.py
```diff
@@ -1374,6 +1393,25 @@ def __init__(self, config: CheriConfig):
...
+    return directory / (project.disk_image_prefix + project.build_configuration_suffix(xtarget) + ".tar")
+
+
+class BuildCheriBSDTarball(BuildCheriBSDDiskImage):
+    is_tarball = True
+    target = "rootfs-tarball"
+
+    default_disk_image_path = ComputedDefaultValue(
+        function=lambda conf, proj: _default_tar_name(conf, conf.output_root, proj),
+        as_string=lambda cls: "$OUTPUT_ROOT/" + cls.disk_image_prefix + "-<TARGET>-disk.img depending on architecture")
```
jrtc27, line 1407: Help string is outdated

pycheribuild/projects/disk_image.py
nwf: Fair; I just copied this from `cheribsd.py`, which might merit a similar change.

pycheribuild/projects/disk_image.py
nwf: I had shied away from that because I thought we might want `minimal` tarballs, too, but yeah, that's probably a better plan.

pycheribuild/projects/disk_image.py
nwf: Ah ha, `bsdtar` has a `-a` option.  How cute, let's go with that.

jrtc27: > if it is unknown suffix or no suffix, creates a new archive with restricted pax format and bzip2 compression.

This seems undesirable...

jrtc27: Though that seems inaccurate in my testing, where it does what you'd actually want, i.e. leave it uncompressed

jrtc27: So yeah let's go with -a :)

pycheribuild/projects/disk_image.py
```diff
@@ -1374,6 +1374,41 @@ def __init__(self, config: CheriConfig):
...
+class BuildCheriBSDTarball(BuildCheriBSDDiskImage):
+    target = "rootfs-tarball"
+
+    default_disk_image_path = ComputedDefaultValue(
+        function=lambda conf, proj: _default_tar_name(conf, conf.output_root, proj),
+        as_string=lambda cls: "$OUTPUT_ROOT/" + cls.disk_image_prefix + "-<TARGET>.tar.xz depending on architecture")
+
+    def __init__(self, config: CheriConfig):
+        super().__init__(config)
+        self.bsdtar_cmd = "bsdtar"
```
jrtc27, line 1391: This seems unnecessary?

nwf: Indeed.  Another bit of copy-paste.

jrtc27 (approved):

pycheribuild/projects/disk_image.py
```diff
@@ -1374,6 +1374,40 @@ def __init__(self, config: CheriConfig):
...
+        self.add_required_system_tool("bsdtar", cheribuild_target="bsdtar", apt="libarchive-tools")
+
+    def make_disk_image(self):
+        # write out the manifest file:
+        self.mtree.write(self.manifest_file, pretend=self.config.pretend)
+        bsdtar_path = shutil.which("bsdtar")
+        if not bsdtar_path:
+            if not self.config.pretend:
+                raise LookupError("Could not find bsdtar command in PATH")
+            bsdtar_path = str(self.bsdtar_cmd)
```
jrtc27, line 1400: Uh wait this is now wrong

jrtc27: (and hence why Jenkins fails)

nwf: m(  `git add -p` fail!

PR #289, opened 2022-05-05
Give large disk images even more free space

brooksdavis:
With the previous settings, we could install the llvm package, but
we'd have negative free space and even with things like deleting the
package cache don't have enough space to add GDB.  Empty space is cheap
so just double to 2g.


arichardson (approved):
Might want to remove the kyua comment that no longer applies now that it's in base but other than that LGTM.

arichardson (approved):

PR #290, opened 2022-05-11
morello: Use new var-args and stock TLS on newer CheriBSD.

bsdjhb:


bsdjhb:
I went with using the re module rather than shelling out to awk.

pycheribuild/config/compilation_targets.py
```diff
@@ -716,11 +729,16 @@ def toolchain_targets(cls, target: "CrossCompileTarget", config: "CheriConfig"):
         return ["morello-llvm-native"]
 
     @classmethod
-    def essential_compiler_and_linker_flags_impl(cls, *args, xtarget, **kwargs):
-        result = super().essential_compiler_and_linker_flags_impl(*args, xtarget=xtarget, **kwargs)
+    def essential_compiler_and_linker_flags_impl(cls, instance: "CheriBSDTargetInfo", *args, xtarget, **kwargs):
+        result = super().essential_compiler_and_linker_flags_impl(instance, *args, xtarget=xtarget, **kwargs)
+        version = instance.cheribsd_version()
+        if version >= 20220511:
```
bsdjhb, line 735: This does default to the old flags if there is no sys/param.h, but I think that is ok.

jrtc27 (approved):
Targets that poke `essential_compiler_and_linker_flags` in `__init__` rather than `setup` can be broken by this if cheribsd is being built during the same cheribuild invocation and stepping across the ABI break, but there's nothing we can do to avoid that, this is as good as we can get, and things _shouldn't_ be messing with CFLAGS in `__init__`.

bsdjhb:
That is good motivation to remove the fallback in the future.

PR #292, opened 2022-06-29
rsync: don't build in source directory

nwf:
This seems to work fine; tested locally with `rsync-morello-purecap`, `rsync-riscv64-purecap`, and `rsync-native`.


arichardson (approved):
LGTM if it works for you :)

PR #293, opened 2022-07-01
Fix build for bluespec-compiler

### LINKED ISSUE #291: bluespec-compiler fails to build on Linux host
bluespec-compiler fails to build on Ubuntu 21.10 because bsc have renamed their Makefile to GNUmakefile.  cheribuild expects those to have a 'distclean' rule, which bsc doesn't have:

```
cheribuild$ ./cheribuild.py bluespec-compiler
Checking /home/atm26/.config/cheribuild.json since /local/scratch/atm26/CTSRD-CHERI/cheribuild/cheribuild.json doesn't exist
Note: Configuration file /home/atm26/.config/cheribuild.json does not exist, using only command line arguments.
Sources will be stored in /home/atm26/cheri
Build artifacts will be stored in /home/atm26/cheri/output
cd /local/scratch/atm26/CTSRD-CHERI/cheribuild && git fetch
cd /home/atm26/cheri/bsc && git fetch
Skipping update: Current HEAD is up-to-date or ahead of upstream.
cd /home/atm26/cheri/bsc && gmake distclean
gmake: *** No rule to make target 'distclean'. Stop.
Warning: 'gmake distclean' FD flags for <stdin> changed from 0x80002 to 0x80402 - resetting them.
Fatal error (in target bluespec-compiler): Command `gmake distclean` failed with non-zero exit code 2. Working directory was /home/atm26/cheri/bsc
```

tmarkettos:
Upstream bsc have a new GNUmakefile which is substantially different from
the kind that cheribuild expects.  Adjust our build rules to suit.

Currently we don't build the documentation since that implies lots of Latex
dependencies, that aren't strictly necessary to use the bsc compiler

Fixes #291

### DIFF
--- a/pycheribuild/projects/bluespec_compiler.py
+++ b/pycheribuild/projects/bluespec_compiler.py
@@ -52,8 +52,7 @@ class BuildBluespecCompiler(Project):
         try:
             self.run_make("full_clean")
         except Exception:
-            self.info("Cleaning failed, continuing anyway");
-
+            self.info("Cleaning failed, continuing anyway")
 
     def compile(self, **kwargs):
         try:
--- a/pycheribuild/projects/bluespec_compiler.py
+++ b/pycheribuild/projects/bluespec_compiler.py
@@ -45,11 +45,21 @@ class BuildBluespecCompiler(Project):
         self.add_required_system_tool("cabal", apt="cabal-install", homebrew="cabal-install")
         for i in ("autoconf", "gperf", "bison", "flex"):
             self.add_required_system_tool(i, homebrew=i)
+        self.add_required_system_header("tcl/tcl.h", apt="tcl-dev")
         self.make_args.set(PREFIX=self.install_dir)
+
+    def clean(self, **kwargs):
+        try:
+            self.run_make("full_clean")
+        except Exception:
+            self.info("Cleaning failed, continuing anyway");
+
 
     def compile(self, **kwargs):
         try:
-            self.run_make("all")
+            # a full install includes documentation which drags in
+            # lots of Latex dependencies, so just build the tools
+            self.run_make("install-src")
         except Exception:
             self.info("Compilation failed. If it complains about missing packages try running:\n"
                       "\tcabal install regex-compat syb old-time split\n"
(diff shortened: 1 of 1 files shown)


pycheribuild/projects/bluespec_compiler.py
```diff
@@ -45,11 +45,20 @@ def __init__(self, config: CheriConfig):
         self.add_required_system_tool("cabal", apt="cabal-install", homebrew="cabal-install")
         for i in ("autoconf", "gperf", "bison", "flex"):
             self.add_required_system_tool(i, homebrew=i)
+        self.add_required_system_header("tcl/tcl.h", apt="tcl-dev")
         self.make_args.set(PREFIX=self.install_dir)
 
+    def clean(self, **kwargs):
```
arichardson, line 51: It looks like latest version support out-of-source builds, if we can use that instead we can avoid changing `clean()`.

pycheribuild/projects/bluespec_compiler.py
```diff
@@ -45,11 +45,20 @@ def __init__(self, config: CheriConfig):
         self.add_required_system_tool("cabal", apt="cabal-install", homebrew="cabal-install")
         for i in ("autoconf", "gperf", "bison", "flex"):
             self.add_required_system_tool(i, homebrew=i)
+        self.add_required_system_header("tcl/tcl.h", apt="tcl-dev")
         self.make_args.set(PREFIX=self.install_dir)
```
arichardson, line 49: If you remove `build_in_source_dir = True`, the following might be sufficient:
```suggestion
        self.make_args.set(PREFIX=self.install_dir, BUILDDIR=self.build_dir)
```

arichardson: Maybe I should just remove the `distclean` workaround and always use `git clean` for in-source clean. For now using this seems fine.

arichardson: I wonder if it works if you set BUILDIR in the environment instead of on the make command line?

arichardson:
Is this still needed?

tmarkettos:
The current version with recent changes is working for me, so closing this PR.

PR #294, opened 2022-07-20
morello-syzkaller build

mazalan01:


pycheribuild/projects/syzkaller.py
```diff
@@ -62,6 +67,9 @@ def setup_config_options(cls, **kwargs):
             "run-sysgen", show_help=True,
             help="Rerun syz-extract and syz-sysgen to rebuild generated Go "
                  "syscall descriptions.")
+        cls.use_go = cls.add_config_option("use-go", kind=GOROOTOptions, default=GOROOTOptions.DEFAULT,
+                                             enum_choice_strings=[t.value for t in GOROOTOptions],
+                                             help="Specify GOROOT default is system")
```
qwattash, line 72: Indentation

pycheribuild/projects/syzkaller.py
```diff
@@ -95,15 +103,17 @@ def needs_configure(self) -> bool:
...
-            GOPATH=self.gopath.expanduser(),
-            CC=self.CC, CXX=self.CXX,
-            PATH=self._new_path)
+        args = {
+            "HOSTARCH":"amd64",
+            "TARGETARCH":"arm64" if self.crosscompile_target.cpu_architecture.value == "aarch64"
+                else self.crosscompile_target.cpu_architecture.value,
+            "TARGETOS":"freebsd",
+            "CC":self.CC, "CXX":self.CXX,
+            "PATH":self._new_path}
```
qwattash, line 112: `}` in a new line, I think

pycheribuild/projects/syzkaller.py
```diff
@@ -134,19 +144,28 @@ def install(self, **kwargs):
...
+        if self.use_go == GOROOTOptions.UPSTREAM:
+            args["GOROOT"] = self.goroot.expanduser()
+            args["GOPATH"] = self.gopath.expanduser()
+        self.make_args.set_env(**args)
 
         self.run_make("clean", parallel=False, cwd=self.gosrc)
         joiner = super().clean()
         return joiner
 
+class BuildMorelloSyzkaller(BuildSyzkaller):
```
qwattash, line 163: This looks good to me

pycheribuild/projects/syzkaller.py
```diff
@@ -41,6 +42,10 @@
 from ..qemu_utils import QemuOptions
 from ..utils import ThreadJoiner
 
+class GOROOTOptions(Enum):
```
kwitaszczyk, line 45: I'd call it `GoType`.

pycheribuild/projects/syzkaller.py
```diff
@@ -62,6 +67,13 @@ def setup_config_options(cls, **kwargs):
             "run-sysgen", show_help=True,
             help="Rerun syz-extract and syz-sysgen to rebuild generated Go "
                  "syscall descriptions.")
+        cls.use_go = cls.add_config_option("use-go", kind=GOROOTOptions,
+            default=GOROOTOptions.DEFAULT, enum_choice_strings=[t.value for t in GOROOTOptions],
+            help="Specify GOROOT default is system")
```
kwitaszczyk: To make it consistent with `use-qemu`:
```
help="The Go type to run with."
```
The default value should be printed by `cheribuild --help` and doesn't have to be included in `help`.

pycheribuild/projects/syzkaller.py
```diff
@@ -95,15 +106,19 @@ def needs_configure(self) -> bool:
...
-            GOPATH=self.gopath.expanduser(),
-            CC=self.CC, CXX=self.CXX,
-            PATH=self._new_path)
+        print(self._new_path)
+        args = {
+            "HOSTARCH":"amd64",
+            "TARGETARCH":"arm64" if self.crosscompile_target.cpu_architecture.value == "aarch64"
+                else self.crosscompile_target.cpu_architecture.value,
+            "TARGETOS":"freebsd",
+            "CC":self.CC, "CXX":self.CXX,
```
kwitaszczyk: Let's move `CXX` to a separate line to have one key-value per line.

pycheribuild/projects/syzkaller.py
```diff
@@ -62,6 +67,13 @@ def setup_config_options(cls, **kwargs):
             "run-sysgen", show_help=True,
             help="Rerun syz-extract and syz-sysgen to rebuild generated Go "
                  "syscall descriptions.")
+        cls.use_go = cls.add_config_option("use-go", kind=GOROOTOptions,
+            default=GOROOTOptions.DEFAULT, enum_choice_strings=[t.value for t in GOROOTOptions],
+            help="Specify GOROOT default is system")
+
+    def create_new_path(self, config):
```
kwitaszczyk, line 75: I'd call it `get_path` to match other `get_*` methods that have similar meaning in other classes.

kwitaszczyk, line 75: `config` is unused. You use `self.config` that `config` is assigned to before this function is called.

pycheribuild/projects/syzkaller.py
```diff
@@ -62,6 +67,13 @@ def setup_config_options(cls, **kwargs):
...
+        self._new_path = (str(self.config.cheri_sdk_dir / "bin") + ":" +
+                          str(self.config.dollar_path_with_other_tools))
```
kwitaszczyk, line 77: Instead of setting the value, I'd return it. Then, instead of creating a field `_new_path` and maintaining it, we can simply call `get_path()` when we construct `args`.

pycheribuild/projects/syzkaller.py
```diff
@@ -95,15 +106,19 @@ def needs_configure(self) -> bool:
...
-
-        self.make_args.set_env(
-            HOSTARCH="amd64",
-            TARGETARCH=self.crosscompile_target.cpu_architecture.value,
-            TARGETOS="freebsd",
-            GOROOT=self.goroot.expanduser(),
-            GOPATH=self.gopath.expanduser(),
-            CC=self.CC, CXX=self.CXX,
-            PATH=self._new_path)
+        print(self._new_path)
```
kwitaszczyk, line 110: Unneeded debug.

pycheribuild/projects/syzkaller.py
```diff
@@ -95,15 +106,19 @@ def needs_configure(self) -> bool:
...
+            "TARGETARCH":"arm64" if self.crosscompile_target.cpu_architecture.value == "aarch64"
+                else self.crosscompile_target.cpu_architecture.value,
+            "TARGETOS":"freebsd",
+            "CC":self.CC, "CXX":self.CXX,
+            "PATH":self._new_path
+        }
+        if self.use_go == GOROOTOptions.UPSTREAM:
+            args["GOROOT"] = self.goroot.expanduser()
+            args["GOPATH"] = self.gopath.expanduser()
+        self.make_args.set_env(**args)
```
kwitaszczyk: We duplicate args in `clean()`. Instead of that, we should introduce `_setup_make_args()` as in `pycheribuild/projects/cross/cheribsd.py`, set args there and call `_setup_make_args()` in `compile()` and `clean()`.

pycheribuild/projects/syzkaller.py
```diff
@@ -134,19 +149,34 @@ def install(self, **kwargs):
...
 
+class BuildMorelloSyzkaller(BuildSyzkaller):
+    target = "morello-syzkaller"
+    github_base_url = "https://github.com/CTSRD-CHERI/"
+    repository = GitRepository(github_base_url + "cheri-syzkaller.git", force_branch=True,
+        default_branch="morello-syzkaller")
+
+    def create_new_path(self, config):
+        # Replace sdk with morello-sdk:
+        self._new_path = (str(self.config.cheri_sdk_dir)[0:-3] + "morello-sdk/bin" + ":" +
```
kwitaszczyk, line 179: Use `self.config.morello_sdk_dir` instead. Also, `self.config.morello_sdk_dir` is Path so you can concatenate paths with `/`, e.g. `self.config.morello_sdk_dir / "bin"`.

kwitaszczyk (changes requested):

pycheribuild/projects/syzkaller.py
qwattash: Also there should be a `morello_sdk_bindir` in the config object that points already to the binary directory

pycheribuild/projects/syzkaller.py
```diff
@@ -62,6 +68,13 @@ def setup_config_options(cls, **kwargs):
             "run-sysgen", show_help=True,
             help="Rerun syz-extract and syz-sysgen to rebuild generated Go "
                  "syscall descriptions.")
+        cls.use_go = cls.add_config_option("use-go", kind=GOROOTOptions, default=GOROOTOptions.DEFAULT,
+                                           enum_choice_strings=[t.value for t in GOROOTOptions],
+                                           help="Specify GOROOT default is system")
+
+    def create_new_path(self, config):
+        self._new_path = (str(self.config.cheri_sdk_dir / "bin") + ":" +
```
qwattash, line 76: Same as below, it should now use `cheri_sdk_bindir` since it's already there

pycheribuild/projects/syzkaller.py
```diff
@@ -62,92 +69,135 @@ def setup_config_options(cls, **kwargs):
...
                  "syscall descriptions.")
+        cls.use_go = cls.add_config_option("use-go", kind=GoType, default=GoType.DEFAULT,
+                                           enum_choice_strings=[t.value for t in GoType],
+                                           help="The Go type to run with.")
+
+    def get_path(self):
+        return (str(self.get_sdk_bindir()) + ":" +
+                str(self.config.dollar_path_with_other_tools))
+
+    def get_sdk_dir(self, config):
```
qwattash, line 80: The config is always the one from the current object, we probably can just use `self.config here`

pycheribuild/projects/syzkaller.py
```diff
@@ -62,92 +69,135 @@ def setup_config_options(cls, **kwargs):
...
+    def get_sdk_bindir(self):
+        return self.config.cheri_sdk_bindir
 
     def __init__(self, config):
-        self._install_prefix = config.cheri_sdk_dir
-        self._install_dir = config.cheri_sdk_dir
+        self._install_prefix = self.get_sdk_dir(config)
+        self._install_dir = self.get_sdk_dir(config)
         self.destdir = ""
         super().__init__(config)
```
qwattash, line 90: There is no reason for late calling super `__init__()` I'd call it first so we have `self.config` available everywhere. Also we can call the getters everywhere instead of having private variables for those, but this is more flavor.

pycheribuild/projects/syzkaller.py
```diff
@@ -62,92 +69,135 @@ def setup_config_options(cls, **kwargs):
...
         self.make_args.set_env(CFLAGS=" ".join(cflags))
         self.run_make(parallel=False, cwd=self.gosrc)
 
     def generate(self):
-        with self.set_env(PATH=self._new_path, SOURCEDIR=self.cheribsd_dir):
+        with self.set_env(PATH=self.get_path(), SOURCEDIR=self.cheribsd_dir):
             self.run_make("extract", parallel=False, cwd=self.gosrc)
             self.run_make("generate", parallel=False, cwd=self.gosrc)
 
+    def get_every_file_path(self, dir):
```
qwattash, line 149: Maybe rename to something more meaningful, e.g. `get_install_files`

pycheribuild/projects/syzkaller.py
```diff
@@ -62,92 +69,135 @@ def setup_config_options(cls, **kwargs):
...
+        file_paths = []
+        for path in dir.iterdir():
+            if path.is_dir():
+                sub_files = self.get_every_file_path(path)
+                dir_name = path.parts[-1]
+                for s in sub_files:
+                    s[1] = dir_name + "/" + s[1]
```
qwattash, line 156: use pathlib path concatenation, this is not portable

pycheribuild/projects/syzkaller.py
```diff
@@ -62,92 +69,135 @@ def setup_config_options(cls, **kwargs):
...
+        with self.set_env(PATH=self.get_path(), SOURCEDIR=self.cheribsd_dir):
             self.run_make("extract", parallel=False, cwd=self.gosrc)
             self.run_make("generate", parallel=False, cwd=self.gosrc)
 
+    def get_every_file_path(self, dir):
+        file_paths = []
+        for path in dir.iterdir():
+            if path.is_dir():
+                sub_files = self.get_every_file_path(path)
+                dir_name = path.parts[-1]
```
qwattash, line 154: use `path.parent.name`, more idiomatic

pycheribuild/projects/syzkaller.py
```diff
@@ -62,92 +69,135 @@ def setup_config_options(cls, **kwargs):
...
+                for s in sub_files:
+                    s[1] = dir_name + "/" + s[1]
+                    file_paths.append(s)
+            else:
+                file_paths.append([path, path.parts[-1]])
```
qwattash, line 159: same here

pycheribuild/projects/syzkaller.py
```diff
@@ -62,92 +69,135 @@ def setup_config_options(cls, **kwargs):
...
+        return file_paths
+
     def install(self, **kwargs):
         # XXX-AM: should have a propert install dir configuration
         native_build = self.source_dir / "bin"
-        mips64_build = native_build / "freebsd_mips64"
-        syz_remote_install = self.syzkaller_install_path() / "freebsd_mips64"
+        build = native_build
```
qwattash, line 165: Either use `build` or `native_build`. I'd pick one and probably also rename it to something more meaningful e.g. `build_output_path`

pycheribuild/projects/syzkaller.py
```diff
@@ -62,92 +69,135 @@ def setup_config_options(cls, **kwargs):
...
 
         if not self.config.pretend:
-            # mips64_build does not exist if we preted, so skip
-            for fname in os.listdir(str(mips64_build)):
-                fpath = mips64_build / fname
-                if os.path.isfile(fpath):
-                    self.install_file(fpath, syz_remote_install / fname, mode=0o755)
+            # build does not exist if we preted, so skip
+            for fpath in self.get_every_file_path(build):
+                if os.path.isfile(fpath[0]):
```
qwattash, line 175: We should probably stop using `os.path` and replace this with pathlib `Path.is_file()` since you already have a pathlib Path here.

pycheribuild/projects/syzkaller.py
```diff
@@ -62,92 +69,135 @@ def setup_config_options(cls, **kwargs):
...
+                    self.install_file(fpath[0], syz_remote_install / fpath[1], mode=0o755)
```
qwattash, line 176: you can avoid having to move pairs of `[path, name]` by using `path.name` to grab the name here as well.

pycheribuild/projects/syzkaller.py
```diff
@@ -62,92 +69,135 @@ def setup_config_options(cls, **kwargs):
...
         return joiner
 
 
+class BuildMorelloSyzkaller(BuildSyzkaller):
+    target = "morello-syzkaller"
+    github_base_url = "https://github.com/CTSRD-CHERI/"
+    repository = GitRepository(github_base_url + "cheri-syzkaller.git", force_branch=True,
+                               default_branch="morello-syzkaller")
+
+    def get_sdk_dir(self, config):
```
qwattash, line 194: same as above

pycheribuild/projects/syzkaller.py
```diff
@@ -42,6 +44,12 @@
 from ..utils import ThreadJoiner
 
 
+class GoType(Enum):
+    DEFAULT = "system"
```
arichardson, line 48: This `DEFAULT` member is unnecessary?

pycheribuild/projects/syzkaller.py
```diff
@@ -62,92 +70,134 @@ def setup_config_options(cls, **kwargs):
             "run-sysgen", show_help=True,
             help="Rerun syz-extract and syz-sysgen to rebuild generated Go "
                  "syscall descriptions.")
+        cls.use_go = cls.add_config_option("use-go", kind=GoType, default=GoType.DEFAULT,
```
arichardson, line 73:
```suggestion
        cls.use_go = cls.add_config_option("use-go", kind=GoType, default=GoType.SYSTEM,
```

pycheribuild/projects/syzkaller.py
```diff
@@ -62,92 +70,134 @@ def setup_config_options(cls, **kwargs):
...
         self.gopath = self.build_dir
         self.gosrc = self.source_dir
 
-        self._new_path = (str(self.config.cheri_sdk_dir / "bin") + ":" +
-                          str(self.config.dollar_path_with_other_tools))
+        self.targetarch = self.crosscompile_target.cpu_architecture.value
+        if self.targetarch == "aarch64":
+            self.targetarch = "arm64"
+
+        self.get_path()
```
arichardson, line 106: This line doesn't do anything.

pycheribuild/projects/syzkaller.py
```diff
@@ -62,92 +70,134 @@ def setup_config_options(cls, **kwargs):
...
 
-        self.make_args.set_env(
-            HOSTARCH="amd64",
-            TARGETARCH=self.crosscompile_target.cpu_architecture.value,
-            TARGETOS="freebsd",
-            GOROOT=self.goroot.expanduser(),
-            GOPATH=self.gopath.expanduser(),
-            CC=self.CC, CXX=self.CXX,
-            PATH=self._new_path)
+        self._setup_make_args()
```
arichardson, line 137: This should probably be called inside `setup()` instead.

pycheribuild/projects/syzkaller.py
```diff
@@ -62,92 +70,134 @@ def setup_config_options(cls, **kwargs):
...
-        self.make_args.set_env(
-            HOSTARCH="amd64",
-            TARGETARCH=self.crosscompile_target.cpu_architecture.value,
-            TARGETOS="freebsd",
-            GOROOT=self.goroot.expanduser(),
-            GOPATH=self.gopath.expanduser(),
-            CC=self.CC, CXX=self.CXX,
-            PATH=self._new_path)
+
+        self._setup_make_args()
```
arichardson, line 181: If this is called in `setup()` it should be correct in both `clean()` and `compile()`

pycheribuild/projects/syzkaller.py
mazalan01: Thanks, I made the changes.

pycheribuild/projects/syzkaller.py
```diff
@@ -62,92 +68,128 @@ def setup_config_options(cls, **kwargs):
...
+        return self.get_sdk_dir() / "syzkaller"
 
     def syzkaller_binary(self):
-        return self.config.cheri_sdk_bindir / "syz-manager"
+        return self.get_sdk_bindir() / "syz-manager"
 
     def needs_configure(self) -> bool:
         return False
 
+    def _setup_make_args(self):
```
arichardson, line 118:
```suggestion
    def setup(self):
        super().setup()
```

The setup() function will be called automatically after __init__(), so no need to call it from there.
Any setup logic that doesn't set instance variables should ideally be moved to setup(). __init__() is always called, but setup() is only called when the target is actually executed, so moving as much as possible to setup() reduces startup time (and things like tab-completion).

pycheribuild/projects/syzkaller.py
```diff
@@ -62,92 +68,128 @@ def setup_config_options(cls, **kwargs):
...
-                          str(self.config.dollar_path_with_other_tools))
+        self.targetarch = self.crosscompile_target.cpu_architecture.value
+        if self.targetarch == "aarch64":
+            self.targetarch = "arm64"
 
         cheribsd_target = self.get_crosscompile_target(config).get_rootfs_target()
         self.cheribsd_dir = BuildCHERIBSD.get_source_dir(self, cross_target=cheribsd_target)
 
+        self._setup_make_args()
+
```
arichardson, line 108:
```suggestion
```

arichardson (changes requested):

pycheribuild/projects/syzkaller.py
```diff
@@ -62,92 +68,127 @@ def setup_config_options(cls, **kwargs):
...
+                file_paths.append(path)
+        return file_paths
+
     def install(self, **kwargs):
         # XXX-AM: should have a propert install dir configuration
-        native_build = self.source_dir / "bin"
-        mips64_build = native_build / "freebsd_mips64"
-        syz_remote_install = self.syzkaller_install_path() / "freebsd_mips64"
+        build_output_path = self.source_dir / "bin"
+        syz_remote_install = self.syzkaller_install_path()
```
qwattash, line 158: Mayber rename syz_remote_install -> sdk_install_path

pycheribuild/projects/syzkaller.py
```diff
@@ -62,95 +68,161 @@ def setup_config_options(cls, **kwargs):
...
-        return self.config.cheri_sdk_bindir / "syz-manager"
+        return self.get_sdk_bindir() / "syz-manager"
 
     def needs_configure(self) -> bool:
         return False
 
+    def setup(self):
+        super().setup()
+        args = {
+            "HOSTARCH": "amd64",
```
arichardson, line 119: This probably shouldn't be hardcoded, but I guess is fine for now.

pycheribuild/projects/syzkaller.py
```diff
@@ -62,95 +68,161 @@ def setup_config_options(cls, **kwargs):
...
+                               default_branch="morello-syzkaller")
+
+    def get_sdk_dir(self):
+        return self.config.morello_sdk_dir
+
+    def get_sdk_bindir(self):
+        return self.config.morello_sdk_bindir
+
+
+class VMSetting(Enum):
```
arichardson, line 192: TargetSytemSettings or something like that? Not all of these are virtual machines.

pycheribuild/projects/syzkaller.py
```diff
@@ -62,95 +68,161 @@ def setup_config_options(cls, **kwargs):
...
+                "image_device": "drive index=0,media=disk,format=raw,file=",
+                "count": 1,
+                "cpu": 1,
+                "mem": 2048,
+                    }
+            }
+    MORELLO = {
+            "type": "morello",
+            "vm": {
+                "mbox_address": "morello7-dev.sec.cl.cam.ac.uk",
```
arichardson, line 211: I'm not sure it's a good idea to hardcode hostnames here, maybe just use a placeholder like `my-morello-hostname.sec.cl.cam.ac.uk`?

pycheribuild/projects/syzkaller.py
mazalan01: I used vm for consistency. In syzkaller, vm is used to refer to them, and in the config file you specify the system settings under "vm". But I agree that TargetSystemSettings is more accurate, should I change it?

qwattash:
Superseded by #373

Issue #297, opened 2022-08-19
openssh: make args not honored on unclean builds


brooksdavis:
In #295 I added DESTDIR to make arguments so DESTDIR is honored. This works on a clean build (e.g., `cheribuild openssh-morello-purecap --clean`) where `gmake install` is called with:
```
Installing openssh-morello-purecap (-morello-purecap) ... 
cd /home/bed22/cheri/build/openssh-morello-purecap-build && env PKG_CONFIG_PATH=/home/bed22/cheri/output/rootfs-morello-purecap/usr/libdata/pkgconfig:/home/bed22/cheri/output/rootfs-morello-purecap/usr/local/morello-purecap/lib/pkgconfig:/home/bed22/cheri/output/rootfs-morello-purecap/usr/local/morello-purecap/share/pkgconfig:/home/bed22/cheri/output/rootfs-morello-purecap/usr/local/morello-purecap/libdata/pkgconfig:/home/bed22/cheri/output/rootfs-morello-purecap/usr/local/lib/pkgconfig:/home/bed22/cheri/output/rootfs-morello-purecap/usr/local/share/pkgconfig:/home/bed22/cheri/output/rootfs-morello-purecap/usr/local/libdata/pkgconfig PKG_CONFIG_LIBDIR=/home/bed22/cheri/output/rootfs-morello-purecap/usr/libdata/pkgconfig:/home/bed22/cheri/output/rootfs-morello-purecap/usr/local/morello-purecap/lib/pkgconfig:/home/bed22/cheri/output/rootfs-morello-purecap/usr/local/morello-purecap/share/pkgconfig:/home/bed22/cheri/output/rootfs-morello-purecap/usr/local/morello-purecap/libdata/pkgconfig:/home/bed22/cheri/output/rootfs-morello-purecap/usr/local/lib/pkgconfig:/home/bed22/cheri/output/rootfs-morello-purecap/usr/local/share/pkgconfig:/home/bed22/cheri/output/rootfs-morello-purecap/usr/local/libdata/pkgconfig PKG_CONFIG_SYSROOT_DIR=/home/bed22/cheri/output/rootfs-morello-purecap DESTDIR=/home/bed22/cheri/output/rootfs-morello-purecap nice gmake install DESTDIR=/home/bed22/cheri/output/rootfs-morello-purecap
```
but when I run without `--clean` we get:
```
Installing openssh-morello-purecap (-morello-purecap) ... 
[...]

PR #298, opened 2022-08-19
Fix DESTDIR passing for openssh

brooksdavis:
Set configure and make arguments in setup so they are alwasy set and
make configure() only handle autoreconf.

Fixes:		#297
Suggested by:	@jrtc27


jrtc27 (approved):
Assuming it works of course

PR #299, opened 2022-08-19
sail: catch up with upstream

nwf:
Ever since the patch series of
https://github.com/rems-project/sail/commit/09e69c98c60b6b8467e39ed17fd5dff01011cb88
through
https://github.com/rems-project/sail/commit/6ba9f1d726bf8e1ed826a1886eaa8e8d79c93135
landed, sail has, from the looks of it, had a dune-based build system with opam
backwards compatibility.  Rather than teach cheribuild about dune, take
advantage of the latter.

### DIFF
--- a/pycheribuild/projects/sail.py
+++ b/pycheribuild/projects/sail.py
@@ -411,9 +411,7 @@ class BuildSailFromSource(OcamlProject):
         pass
 
     def install(self, **kwargs):
-        # Use ./opam to just build sail, not coq-sail.  Using '.' will try to
-        # build both and we probably don't need all of coq installed just now
-        self.run_in_ocaml_env("opam install -y ./opam")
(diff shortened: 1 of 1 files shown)


arichardson (approved):
This looks fine to me. I was actually considering removing this target since I always use the "sail" target with --sail/use-git-version, but it looks like there are users after all :)

nwf:
If I should be using `sail --sail/use-git-version` instead, this target can go.  I just didn't know about that option.

arichardson:
> If I should be using `sail --sail/use-git-version` instead, this target can go. I just didn't know about that option.

I'm not sure if that works for everyone. Let me know if it does and if so we can remove the old (more fragile) targets.

nwf:
Seems to work for me!  I'll close this in favor of that, then.

PR #300, opened 2022-08-24
CheriBSD: Default to enabling MALLOC_PRODUCTION

arichardson:
Can be disabled by building with --cheribsd/build-type=Debug


arichardson:
I think the default should also be changed in CheriBSD, but it might make sense to make this dependent on the cheribuild build type as well?

brooksdavis:
I'm ok with switching the default here and/or in CheriBSD.

PR #308, opened 2022-10-19
Update CHERI GDB to the cheri-12 branch.

bsdjhb:
- The rule to add "gmp" in dependencies() was inverted.

- The check for "gmp" in dependencies later also did not work as it would not match on dependencies like "gmp-riscv64-purecap-for-rootfs".  However, we can just make this unconditional now.

- Use --with-gmp and --with-libgmp-prefix variables rather than directly setting CFLAGS and LDFLAGS.  Note that GDB master (soon to be 13) follows binutils in using --with-gmp only for this.

- Change the default branch to cheri-12 across all architectures removing logic for morello-specific branches.

### DIFF
--- a/pycheribuild/projects/cross/gdb.py
+++ b/pycheribuild/projects/cross/gdb.py
@@ -34,7 +34,7 @@ from pathlib import Path
 from .crosscompileproject import (BuildType, CheriConfig, CompilationTargets, CrossCompileAutotoolsProject,
                                   DefaultInstallDir, GitRepository, Linkage, MakeCommandKind)
 from .gmp import BuildGmp
-from ..project import TargetBranchInfo, ComputedDefaultValue
+from ..project import ComputedDefaultValue
 from ...processutils import run_command
 from ...utils import OSInfo, status_update
 
@@ -80,8 +80,8 @@ class BuildGDBBase(CrossCompileAutotoolsProject):
     @classmethod
     def dependencies(cls, config: CheriConfig) -> "list[str]":
         deps = super().dependencies(config)
-        # For the native-hybrid build gmp must be installed via ports.
-        if not cls.get_crosscompile_target().is_cheri_hybrid():
+        # For the native and native-hybrid builds gmp must be installed via ports.
+        if not cls.get_crosscompile_target().is_native():
             deps.append("gmp")
         return deps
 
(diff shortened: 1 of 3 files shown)


pycheribuild/projects/cross/gdb.py
```diff
@@ -80,8 +80,8 @@ class BuildGDBBase(CrossCompileAutotoolsProject):
     @classmethod
     def dependencies(cls, config: CheriConfig) -> "list[str]":
         deps = super().dependencies(config)
-        # For the native-hybrid build gmp must be installed via ports.
-        if not cls.get_crosscompile_target().is_cheri_hybrid():
+        # For the native build gmp must be installed via ports.
+        if cls.get_crosscompile_target().is_cheri_hybrid():
```
jrtc27, line 84: I think this fixes your use case but breaks Alex's gdb-native-hybrid build?

pycheribuild/projects/cross/gdb.py
```diff
@@ -160,14 +160,14 @@ def setup(self) -> None:
...
-            if "gmp" in self.cached_full_dependencies():
-                self.COMMON_FLAGS.append("-I" + str(BuildGmp.get_install_dir(self) / "include"))
-                self.LDFLAGS.append("-L" + str(BuildGmp.get_install_dir(self) / "lib"))
-                # Autoconf stupidly decides which to use based on file existence
-                # before trying to actually use the thing, so our passing of
-                # -static means the .so fails to link in and thus the build fails.
-                self.configure_args.append("--with-libgmp-type=static")
+            self.configure_args.append("--with-gmp=" + str(BuildGmp.get_install_dir(self)))
+            # GDB > 12 only uses --with-gmp
+            self.configure_args.append("--wqith-libgmp-prefix=" + str(BuildGmp.get_install_dir(self)))
```
jrtc27, line 165:
```suggestion
            self.configure_args.append("--with-libgmp-prefix=" + str(BuildGmp.get_install_dir(self)))
```

pycheribuild/projects/cross/gdb.py
bsdjhb: Hmm, is this to build natively on a Morello host?  I'm not sure what the right condition to check is.  Previously it enabled "gmp" for "gdb-native" but not for targets like "gdb-riscv64-hybrid" or "gdb-risc64-hybrid-for-purecap-rootfs".  I think what we want is to enable "gmp" for all but "gdb-native" and "gdb-native-hybrid".

jrtc27: Probably `if not (cls.get_crosscompile_target().is_native() and cls.get_crosscompile_target().is_cheri_hybrid())`

jrtc27: But yes; the reason is that there's no NATIVE_HYBRID (which is what NATIVE_NON_PURECAP is on a Morello host) gmp target (but maybe that would be the better approach?..)

arichardson: I think that's what I originally tried but I think I ran into problems with conflicting install paths and just decided to use the ports versions instead on native cheribsd.

pycheribuild/projects/cross/gdb.py
```diff
@@ -80,8 +80,8 @@ class BuildGDBBase(CrossCompileAutotoolsProject):
     @classmethod
     def dependencies(cls, config: CheriConfig) -> "list[str]":
         deps = super().dependencies(config)
-        # For the native-hybrid build gmp must be installed via ports.
-        if not cls.get_crosscompile_target().is_cheri_hybrid():
+        # For the native and native-hybrid builds gmp must be installed via ports.
+        if not cls.get_crosscompile_target().is_native():
```
arichardson, line 84: I guess this is fine too, we can require gmp to use packages on Linux and macos as well easier than bootstrapping it by default.

arichardson (approved):
LGTM if it works for you

arichardson:
The CI failure is a pre-existing bug exposed by my latest commit and can be ignored

bsdjhb:
For the `gdb-native-hybrid` case on Morello I did have to set PKG_CONFIG_PATH explicitly to point to /usr/local64.  It seems like cheribuild should do that automatically for all `*-native-hybrid` targets perhaps.

Command I used to build was `env PKG_CONFIG_PATH=/usr/local64/libdata/pkgconfig python3 ~/work/git/cheribuild/cheribuild.py  gdb-native-hybrid`

This did give me a gdb binary on Morello, but it has the --debug-directory incorrect (points to $SDK/lib/debug instead of /usr/lib/debug).

arichardson:
> For the `gdb-native-hybrid` case on Morello I did have to set PKG_CONFIG_PATH explicitly to point to /usr/local64. It seems like cheribuild should do that automatically for all `*-native-hybrid` targets perhaps.
> 
> Command I used to build was `env PKG_CONFIG_PATH=/usr/local64/libdata/pkgconfig python3 ~/work/git/cheribuild/cheribuild.py gdb-native-hybrid`
> 
> This did give me a gdb binary on Morello, but it has the --debug-directory incorrect (points to $SDK/lib/debug instead of /usr/lib/debug).

That's odd - there is an explicit workaround to add that path. And the debug dir should have been fixed by a recent commit (unless I forgot to push it)

PR #309, opened 2022-10-21
gdb.py: Always use /usr/lib/debug as separate debug dir on FreeBSD.

bsdjhb:


bsdjhb:
This fixes the debug directory setting for `gdb-native` and `gdb-native-hybrid` when built on FreeBSD or CheriBSD.

jrtc27:
This duplicates code that already exists in the !native case, but this probably does make sense, I don't think we're installing anything to ~/cheri/output/sdk/lib/debug whether on FreeBSD or not

bsdjhb:
Arguably we should always set this as even on MacOS or Linux you would use this along with `set sysroot` to cross-debug coredumps or the like and would want it to look in `/usr/lib/debug` in the sysroot.  The question is more not the target host the GDB binary is planning to run on, but what target the GDB binary is planning to examine (which is almost always CheriBSD for cheribuild's purposes)

jrtc27:
Hm, that's true, what the native system does isn't really relevant except when the native system is a CHERI one

arichardson (approved):

Issue #182, opened 2021-03-02
SIGTTOU if cheribuild on a compilation target is invoked twice in a shell script


nwf:
I have a cheribuild wrapping script that builds the matrix of `mrs` configurations used in revocation benchmarks; because `cheribuild` doesn't really understand building the same target over and over, this script steers the system by hand by varying `--mrs/build-directory` for each invocation.  In any case, "recently" (as in, in the past few months?), if I attempt to build more than one configuration per run of the script, the second `cheribuild` process hangs with a `SIGTTOU` in response to `TCSBRK` after reaping a child:
```
[pid 109817] wait4(109818, [{WIFEXITED(s) && WEXITSTATUS(s) == 0}], 0, NULL) = 109818                                                                        
[pid 109817] ioctl(0, TCSBRK, 1)        = ? ERESTARTSYS (To be restarted if SA_RESTART is set)                                                               
[pid 109817] --- SIGTTOU {si_signo=SIGTTOU, si_code=SI_KERNEL} ---                                                                                           [pid 109817] --- stopped by SIGTTOU ---                             
```

I have applied the following hack which lets things run again, but it's clearly going against the spirit of *something* as now `gave_tty_control` is an unused parameter (and `SIGTTOU` is always suppressed during restore):
```
--- a/pycheribuild/processutils.py
+++ b/pycheribuild/processutils.py
@@ -190,7 +190,7 @@ def keep_terminal_sane(gave_tty_control=False, name=""):
     finally:
         # Can seemingly get unwanted SIGTTOU's whilst restoring so just ignore
         # them temporarily.
-        with suppress_sigttou(suppress=gave_tty_control):
+        with suppress_sigttou():
             stdin_state.restore()
             stdout_state.restore()
             stderr_state.restore()
```

I don't understand what's different the second time around relative to the first, which is further evidence that the above is quite dubious.  Better suggestions appreciated.  These particular UNIX dragons have always managed to bamboozle me. :(

arichardson:
I know ~nothing about SIGTTOU, so I'm not sure what's going on here. `git blame` says this is not my code, so I'll defer to @jrtc27 .

jrtc27:
Oh boy. This was probably affected by 63da029de3cd7e3e750015e5ef4366c79488a6b0, which made things less broken for _interactive_ use when called from a non-interactive shell (i.e. script), but could well have caused new issues for scripts. I hate this part of POSIX.

jrtc27:
[...]

PR #310, opened 2022-10-25
Only attempt to restore terminal attributes when in the foreground

arichardson:
Otherwise we can end up blocking indefinitely in tcdrain().

Fixes: https://github.com/CTSRD-CHERI/cheribuild/issues/182


nwf:
I think this makes sense.

PR #311, opened 2022-10-27
Fix ci and misc refactorings

arichardson:


pycheribuild/projects/cross/gdb.py
```diff
@@ -185,8 +188,6 @@ def setup(self) -> None:
             self.LDFLAGS.append("-lelf")
             self.LDFLAGS.append("-lmd")
             self.configure_environment.update(CONFIGURED_M4="m4", CONFIGURED_BISON="byacc", TMPDIR="/tmp", LIBS="")
-            # Look in /usr/lib not /usr/local/lib
```
bsdjhb, line 188: Ah, I guess it was only when building a cross-binary before my change?

arichardson: Yeah and since native builds were not a thing before we didn't notice.

PR #313, opened 2022-11-02
Add --presume-connectivity option

nwf-msr:
If we're heavily firewalled, the attempt to check whether we're online doesn't really indicate whether or not we'll be able to reach, for example, github.com .  Operating in such an environment, I have historically just stubbed out the test locally, but rather than continuing to do that, having an option could be nice.

### DIFF
--- a/pycheribuild/config/defaultconfig.py
+++ b/pycheribuild/config/defaultconfig.py
@@ -103,6 +103,11 @@ class DefaultCheriConfig(CheriConfig):
             "confirm-clone", help="Ask for confirmation before cloning repositories.")
         self.force_update = loader.add_bool_option("force-update", help="Always update (with autostash) even if there "
                                                                         "are uncommitted changes")
+
+        self.presume_connectivity = loader.add_bool_option(
+            "presume-connectivity",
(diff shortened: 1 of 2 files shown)


nwf-msr:
I don't think the CI failure is my fault, as it shows up in #311, too.  It looks like df90883bdc2ef8f726d4e8fbd5abdf1a2da79e51 removed the `# noinspection PyTypeChecker` directive from `add_bool_option` when it moved the function from `project.py` to `simple_project.py`?

arichardson (approved):
This seems okay to me although I do wonder if there is a better check that also succeeds in your environment?

I also wonder if we should assume the same for `ssh_host_accessible`, although that is more questionable?

arichardson:
> I don't think the CI failure is my fault, as it shows up in #311, too. It looks like [df90883](https://github.com/CTSRD-CHERI/cheribuild/commit/df90883bdc2ef8f726d4e8fbd5abdf1a2da79e51) removed the `# noinspection PyTypeChecker` directive from `add_bool_option` when it moved the function from `project.py` to `simple_project.py`?

Should hopefully be fixed now

arichardson (approved):

PR #318, opened 2022-11-18
Add more CHERI-relevant tests from Juliet test suite

eupharina:
- Add the following CWE entries in addition to already present in `juliet_test_suite` project of `cheribuild`
  - 122 Heap-based Buffer Overflow
  - 124 Buffer Underwrite ('Buffer Underflow')
  - 127 Buffer Under-read
  - 134 Uncontrolled Format String
  - 188 Reliance on Data/Memory Layout
  - 587 Assignment of a Fixed Address to a Pointer
  - 588 Attempt to Access Child of a Non-structure Pointer
  - 685 Function Call With Incorrect Number of Arguments
  - 843 Access of Resource Using Incompatible Type ('Type
Confusion')

- Treat compiler detected bugs as warnings, not errors, for relevant CWE entries.
- Provide user input from stdtin, files and environment that triggers illegal memory access.

### DIFF
--- a/pycheribuild/projects/cross/juliet_test_suite.py
+++ b/pycheribuild/projects/cross/juliet_test_suite.py
@@ -111,10 +111,58 @@ class BuildJulietCWE121(BuildJulietCWESubdir):
                                              subdirectory="testcases/CWE121_Stack_Based_Buffer_Overflow")
 
 
+class BuildJulietCWE122(BuildJulietCWESubdir):
+    target = "juliet-cwe-122"
+    cwe_number = 122
+    repository = ReuseOtherProjectRepository(BuildJulietTestSuite,
+                                             subdirectory="testcases/CWE122_Heap_Based_Buffer_Overflow")
+
+
+class BuildJulietCWE124(BuildJulietCWESubdir):
+    target = "juliet-cwe-124"
+    cwe_number = 124
+    repository = ReuseOtherProjectRepository(BuildJulietTestSuite,
+                                             subdirectory="testcases/CWE124_Buffer_Underwrite")
+
+
 class BuildJulietCWE126(BuildJulietCWESubdir):
     target = "juliet-cwe-126"
     cwe_number = 126
-    repository = ReuseOtherProjectRepository(BuildJulietTestSuite, subdirectory="testcases/CWE126_Buffer_Overread")
+    repository = ReuseOtherProjectRepository(BuildJulietTestSuite,
+                                             subdirectory="testcases/CWE126_Buffer_Overread")
+    cwe_setup_commands = [
+                "echo 500 > /tmp/in.txt"
+            ]
+
+
+class BuildJulietCWE127(BuildJulietCWESubdir):
+    target = "juliet-cwe-127"
+    cwe_number = 127
+    repository = ReuseOtherProjectRepository(BuildJulietTestSuite,
+                                             subdirectory="testcases/CWE127_Buffer_Underread")
+    cwe_setup_commands = [
+                "echo -500 > /tmp/in.txt"
+            ]
+
+
+class BuildJulietCWE134(BuildJulietCWESubdir):
+    target = "juliet-cwe-134"
+    cwe_number = 134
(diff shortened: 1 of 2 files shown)


pycheribuild/projects/cross/juliet_test_suite.py
```diff
@@ -66,13 +68,20 @@ class BuildJulietCWESubdir(CrossCompileCMakeProject):
     do_not_add_to_targets = True
     cwe_number = None
     default_install_dir = DefaultInstallDir.DO_NOT_INSTALL
+    cwe_warning_flags = []
```
brooksdavis, line 71: If you add this before adding the CWEs you won' have a commit with contents that are immediately removed.

pycheribuild/projects/cross/juliet_test_suite.py
```diff
@@ -94,6 +103,12 @@ def run_tests(self):
             args.append("--ld-preload-path")
             args.append(self.ld_preload_path)
 
+        for cmd in self.cwe_setup_commands:
+            args.append("--test-setup-command=" + cmd)
+
+        # For stdin redirection
+        args.append("--test-setup-command=" + "touch /tmp/in.txt")
```
brooksdavis, line 110: You can't just write to fixed paths in /tmp. It's unsafe. You need to create a proper temporary file or directory and add hooks to clean it up on exit. You should be able to do this with tempfile.mkstemp().

pycheribuild/projects/cross/juliet_test_suite.py
eupharina: Swapped the commits

pycheribuild/projects/cross/juliet_test_suite.py
brooksdavis: Ah, in that case, this is fine. I missed that it was inside a on-off qemu instance.

arichardson (commented):
Thanks for this pull request, this generally looks good just a few minor suggestions.
I haven't run the test suite in a long time but if it works for you I'm happy to merge this.

pycheribuild/projects/cross/juliet_test_suite.py
```diff
@@ -111,12 +111,46 @@ class BuildJulietCWE121(BuildJulietCWESubdir):
...
+    cwe_number = 134
+    repository = ReuseOtherProjectRepository(BuildJulietTestSuite, subdirectory="testcases/CWE134_Uncontrolled_Format_String")
+    cwe_warning_flags = ["-Wno-error=format-security"]
+
+
+class BuildJulietCWE188(BuildJulietCWESubdir):
+    target = "juliet-cwe-188"
+    cwe_number = 188
+    repository = ReuseOtherProjectRepository(BuildJulietTestSuite, subdirectory="testcases/CWE188_Reliance_on_Data_Memory_Layout")
+    cwe_warning_flags = ["-cheri-bounds=subobject-safe"]
```
arichardson, line 178: This is not really a warning flag and will only work when building for purecap, maybe just override setup to something like this:
```
def setup():
    super().setup()
    if self.compiling_for_cheri():
        self.CFLAGS.append("-cheri-bounds=subobject-safe")
```

pycheribuild/projects/cross/juliet_test_suite.py
```diff
@@ -94,6 +101,12 @@ def run_tests(self):
             args.append("--ld-preload-path")
             args.append(self.ld_preload_path)
 
+        for cmd in self.cwe_setup_commands:
+            args.append("--test-setup-command=" + cmd)
+
+        # For stdin redirection
+        args.append("--test-setup-command=" + "touch /tmp/in.txt")
```
arichardson, line 108:
```suggestion
        args.append("--test-setup-command=touch /tmp/in.txt")
```

arichardson (commented):
LGTM with the -cheri-bounds flag fixed.

arichardson (approved):
Thanks!

arichardson:
Looks like flak8 is unhappy - will merge this once CI passes.

PR #320, opened 2022-11-22
build_qemu: move check for commit to `configure` to ensure the repository is cloned

JOT85:
Previously, building without the source directory would fail.

This is the same solution used for samba: https://github.com/CTSRD-CHERI/cheribuild/blob/9b7f601aae30cd0d6ae2e51436b168c9a5c78fa9/pycheribuild/projects/samba.py#L95


arichardson (approved):

JOT85:
Thanks! :)

PR #322, opened 2022-11-22
utils: ignore blank lines and comments in os-release file

JOT85:
Here's an example of the os-release file for CentOS 7:
```
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"
```

The blank line broke the previous parser.

The man page for os-release (`OS-RELEASE(5)`) specifies:

> Lines beginning with "#" are treated as comments. Blank lines are
> permitted and ignored.

This modified parser will still fail if a line doesn't contain '=', I beleive this is acceptable since OS-RELEASE(5) specifies that each line should be an assignment, so a line without '=' that isn't a comment or blank would be invalid.

### DIFF
--- a/pycheribuild/utils.py
+++ b/pycheribuild/utils.py
@@ -414,7 +414,10 @@ class OSInfo(object):
         with Path("/etc/os-release").open(encoding="utf-8") as f:
             d = {}
             for line in f:
-                k, v = line.rstrip().split("=", maxsplit=1)
+                line = line.strip()
+                if line == '' or line[0] == '#':
+                    continue
+                k, v = line.split("=", maxsplit=1)
                 # .strip('"') will remove if there or else do nothing
                 d[k] = v.strip('"')
         return d


pycheribuild/utils.py
```diff
@@ -414,7 +414,10 @@ def __parse_etc_os_release() -> "dict[str, str]":
         with Path("/etc/os-release").open(encoding="utf-8") as f:
             d = {}
             for line in f:
-                k, v = line.rstrip().split("=", maxsplit=1)
+                line = line.strip()
```
jrtc27, line 417: I don't know why it was just an rstrip before but that shouldn't need changing for this?

JOT85: With only rstrip comments with preceding space won't be ignored, although I doubt that will ever come up in the wild.
In addition, using rstrip will have the key include any prefixing whitespace if there is any, which isn't desirable? I can't think of a reason it would be?

arichardson: I think the only reason it was rstrip before is that I assumed that was sufficient. I agree it should be changed to strip.

arichardson: And it's probably not worth splitting this out into a separate commit

PR #329, opened 2022-12-08
Support hybrid targets for 'webkit'.

jacobbramley:
This is a prerequisite for some CapableVMs work (https://github.com/capablevms/webkit/issues/11).

### DIFF
--- a/pycheribuild/projects/cross/qt5.py
+++ b/pycheribuild/projects/cross/qt5.py
@@ -2,6 +2,8 @@
 # Copyright (c) 2017 Alex Richardson
 # All rights reserved.
 #
+# Copyright (c) 2022 Arm Limited and/or its affiliates <open-source-office@arm.com>
+#
 # This software was developed by SRI International and the University of
 # Cambridge Computer Laboratory under DARPA/AFRL contract FA8750-10-C-0237
 # ("CTSRD"), as part of the DARPA CRASH research programme.
@@ -753,6 +755,10 @@ class BuildICU4C(CrossCompileAutotoolsProject):
     native_install_dir = DefaultInstallDir.CHERI_SDK
     make_kind = MakeCommandKind.GnuMake
     needs_native_build_for_crosscompile = True
+    supported_architectures = list(
+        set(CompilationTargets.ALL_SUPPORTED_CHERIBSD_AND_HOST_TARGETS) |
+        set(CompilationTargets.ALL_CHERIBSD_MORELLO_TARGETS)
+    )
 
     def setup(self):
         super().setup()
--- a/pycheribuild/projects/cross/webkit.py
+++ b/pycheribuild/projects/cross/webkit.py
@@ -1,5 +1,6 @@
 #
 # Copyright (c) 2020 Alex Richardson
+# Copyright (c) 2022 Arm Limited and/or its affiliates <open-source-office@arm.com>
 #
 # This software was developed by SRI International and the University of
 # Cambridge Computer Laboratory (Department of Computer Science and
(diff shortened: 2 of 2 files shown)


pycheribuild/projects/cross/webkit.py
```diff
@@ -29,7 +30,7 @@
 #
 
 from enum import Enum
-
+from ...config.compilation_targets import CompilationTargets
```
jrtc27, line 33: Can be imported from crosscompileproject

pycheribuild/projects/cross/qt5.py
```diff
@@ -753,6 +755,10 @@ class BuildICU4C(CrossCompileAutotoolsProject):
     native_install_dir = DefaultInstallDir.CHERI_SDK
     make_kind = MakeCommandKind.GnuMake
     needs_native_build_for_crosscompile = True
+    supported_architectures = list(
```
jrtc27, line 758: I think we'd normally spell this `supported_architectures = CompilationTargets.ALL_CHERIBSD_TARGETS_WITH_HYBRID + [CompilationTargets.NATIVE]`

pycheribuild/projects/cross/qt5.py
```diff
@@ -2,6 +2,8 @@
 # Copyright (c) 2017 Alex Richardson
 # All rights reserved.
 #
+# Copyright (c) 2022 Arm Limited and/or its affiliates <open-source-office@arm.com>
```
jrtc27, line 5: Is this change really copyrightable (especially once every part of it is rewritten by me...)?

jacobbramley: I am not a lawyer, but my understanding is broadly that copyrightability depends on jurisdiction, and presumably many other things. In any case, Arm asks us (by default) to update copyright statements on _all_ contributions so we don't need to apply our own personal heuristics, which will inevitably be wrong somewhere.

> especially once every part of it is rewritten by me
Well that makes things much easier, yes :-)

arichardson:
Would passing "--enable-hybrid-for-purecap-rootfs-targets" work for you or exporting `CHERIBUILD_ENABLE_HYBRID_FOR_PURECAP_ROOTFS_TARGETS=1` in the environment?

pycheribuild/projects/cross/qt5.py
```diff
@@ -753,6 +755,10 @@ class BuildICU4C(CrossCompileAutotoolsProject):
     native_install_dir = DefaultInstallDir.CHERI_SDK
     make_kind = MakeCommandKind.GnuMake
     needs_native_build_for_crosscompile = True
+    supported_architectures = list(
+        set(CompilationTargets.ALL_SUPPORTED_CHERIBSD_AND_HOST_TARGETS) |
```
arichardson, line 759: Are you sure you intend to build against a hybrid rootfs and not the lib64 libraries in a purecap rootfs? In the latter case I'd suggest `    supported_architectures = CompilationTargets.ALL_CHERIBSD_TARGETS_WITH_HYBRID_FOR_PURECAP_ROOTFS + [
        CompilationTargets.NATIVE]`

arichardson (changes requested):

pycheribuild/projects/cross/qt5.py
jacobbramley: We do actually run hybrid tests on a hybrid rootfs.

arichardson: Is there a specific reason you are using a hybrid rootfs or could you change over to a purecap one?

jacobbramley: The original reason was simply that it seemed like the obvious, simplest thing to do. However, it does seem that we can run our hybrid builds perfectly well with `morello-webkit-morello-hybrid-for-purecap-rootfs`.

Given that, and also that @jrtc27 wanted to rewrite it differently anyway, I propose just closing this PR.

PR #330, opened 2022-12-12
Remove workaround for old GDB build system

arichardson:
This should no longer be needed after updating to GDB12 and it appears to result in problems with newer macOS systems where moving binaries results
in errors: https://theevilbit.github.io/posts/amfi_launch_constraints/


gcjenkinson (approved):
LGTM

PR #331, opened 2022-12-17
Expand --cheribsd/caprevoke-kernel behaviors

nwf:
This will let Jenkins build the caprevoke kernels with names, rather than just being the default `/boot/kernel`, while still letting `--cheribsd/caprevoke-kernel` default to doing what it's always done, displacing the default `/boot/kernel`.


arichardson (approved):
LGTM once tests pass

jrtc27:
And this definitely doesn't break which kernel gets picked when running, whether as a user or for tests in CI?

PR #335, opened 2022-12-30
GDB: register dependency on texinfo

ppaeps:
On a fresh installation of FreeBSD, the GDB build will fail in configure when it can't find makeinfo.  This is part of the texinfo package.


arichardson (commented):
CI is failing because the other callsite that adds a makeinfo dependency (gmp.py line 42) only sets the package name for FreeBSD, could you update that call to pass `default="texinfo"`?

ppaeps:
> CI is failing because the other callsite that adds a makeinfo dependency (gmp.py line 42) only sets the package name for FreeBSD, could you update that call to pass `default="texinfo"`?

Thanks for looking into that for me.  I don't have access to CI yet.  Your suggestion seems to have fixed it.  CI is happy with this.

PR #337, opened 2023-01-11
Remove support for the BESSPIN evaluation tool

brooksdavis:
We're unlikely to need to run this again so remove the clutter.


arichardson (approved):

jrtc27 (approved):
"Commence primary ignition", or something

PR #342, opened 2023-01-17
Use --modversion instead of --exists to check library is installed

### LINKED ISSUE #340: Failing freebsd and cheribsd builds on macOS. pkg-config --exists libarchive
There is a bug in the ports install of libarchive on macOS which causes builds of cheribsd to fail when cheribuild.py checks the dependencies. Specifically, at `pycheribuild/projects/cross/cheribsd.py:430`. The workaround for this is documented in https://github.com/Homebrew/homebrew-core/issues/117869 .

A work-around for cheribuild if this doesn't get fixed soon is to use the `--modversion` option for pkg-config rather than the `--exists` option. I'm not certain if this workaround would break things further down. The one listed on homebrew-core should behave though.

graymalkin:
pkg-config --exists also checks dependencies are present on macOS, and due to some packaging issues with upstream libarchive this is breaking the build for cheribsd and freebsd.

Closes #340.

See https://github.com/Homebrew/homebrew-core/issues/120526 for tracking of this issue as it relates to macOS.

### DIFF
--- a/pycheribuild/projects/simple_project.py
+++ b/pycheribuild/projects/simple_project.py
@@ -821,7 +821,7 @@ class SimpleProject(AbstractProject, metaclass=ABCMeta if typing.TYPE_CHECKING e
             return  # already checked
         self._validate_cheribuild_target_for_system_deps(instructions.cheribuild_target)
         try:
-            self.run_cmd(["pkg-config", "--exists", package])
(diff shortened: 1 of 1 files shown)


arichardson (approved):
Thanks!

I'm not convinced checking for the dependencies is useful, and this change will make it easier to add "minimum pkg-config version" checks in the future.

PR #343, opened 2023-01-29
Add expat and gmp Debian and Arch package lists

JOT85:
I've added expat and gmp to the package lists for Debian and Arch. I needed these to build the riscv64 SDK.

expat is already included in Fedora, but not gmp? Perhaps it's a dependency of something else already or should it be added?


arichardson:
Thanks! GMP is needed by newer versions of GDB, but it was not needed when this list was originally added to the README.

arichardson (approved):

PR #346, opened 2023-02-09
split on only the first `=` when parsing build options

deadly-platypus:
If you have a build option such as `OPTION=\"-DFOO=BAR\"` then the parsing of the config file fails. This just splits the string on the first `=` so the full value (with the second `=`) remains unchanged


jrtc27:
Please rebase your branch, your history has merges that we don't want

deadly-platypus:
is this better?

jrtc27:
Yes, thanks

PR #347, opened 2023-02-11
Update README.md

benlaurie:
It seems that other packages installing odd versions of gcc can cause libstdc++ to be missing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ If you are building CHERI on a Debian/Ubuntu-based machine, the following comman
 the most commonly used cheribuild targets:
 
 ```shell
-apt install autoconf automake libtool pkg-config clang bison cmake mercurial ninja-build samba flex texinfo time libglib2.0-dev libpixman-1-dev libarchive-dev libarchive-tools libbz2-dev libattr1-dev libcap-ng-dev libexpat1-dev libgmp-dev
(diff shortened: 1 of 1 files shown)


jrtc27:
My *guess* is that:
1. You had gcc-12 installed (and possibly older gcc/g++ packages, likely -11), but not g++-12
2. Clang saw 12 as the newest GCC version
3. Clang used GCC 12's headers (it likes to use the newest version available)

```
burn:~ jrtc4% docker run --rm -it ubuntu:22.04
root@1b5489962837:/# apt update
...
root@1b5489962837:/# apt install -y gcc g++ clang
...
root@1b5489962837:/# echo '#include <new>' | clang++ -x c++ - -fsyntax-only
root@1b5489962837:/# apt install -y gcc-12
...
root@1b5489962837:/# echo '#include <new>' | clang++ -x c++ - -fsyntax-only
<stdin>:1:10: fatal error: 'new' file not found
#include <new>
         ^~~~~
1 error generated.
```

This documentation certainly isn't the right fix; libstdc++-12-dev is extremely specific to your situation. I don't think there's much we can really do here, unfortunately, this is a long-standing deficiency in Clang.

jrtc27 (changes requested):
As above

benlaurie:
OK, so I documented the issue instead.

benlaurie:
Maybe we should fix clang?

arichardson (approved):
LGTM, @jrtc27 are you okay with the current wording?

PR #351, opened 2023-02-23
Add --run-fvp/trace-unbuffered.

jacobbramley:
When reading a trace file from a live model, it's usually important that the trace file is unbuffered. This is particularly useful in combination with ToggleMTIPlugin.


arichardson (approved):
CI failure seems unrelated and caused by new checks. Will merge this once I've fixed CI.

PR #357, opened 2023-03-06
Use `intercept-build` instead of `bear` for generating a compilation database

deadly-platypus:
This PR uses `clang/tools/scan-build-py/bin/intercept-build` included in the CheriLLVM source to generate a compilation database, since `bear` is unsupported for building `cheribsd`.


deadly-platypus:
closing pull request after further testing indicates that the functionality is not quite correct

deadly-platypus:
Figured out that the compilation database is actually created with `_GENERIC-MORELLO-PURECAP` added to the filename

arichardson:
`bear` tends to work better than `intercept-build`, so I would prefer to use that. Where does it say `bear` is unsupported for CheriBSD?

deadly-platypus:
`bear` currently cannot be used to build CheriBSD, specifically `cheribuild.py --compilation-db cheribsd-morello-purecap` will fail. However, with this patch, it builds and generates a compilation database that looks correct.  What is better about `bear`?  The reason I like `intercept-build` is that it comes with `cherillvm` and no additional dependencies are required.

arichardson:
> `bear` currently cannot be used to build CheriBSD, specifically `cheribuild.py --compilation-db cheribsd-morello-purecap` will fail. However, with this patch, it builds and generates a compilation database that looks correct. What is better about `bear`? The reason I like `intercept-build` is that it comes with `cherillvm` and no additional dependencies are required.

Hmm you are right it looks like bear fails with cheribsd due to some dodgy quoting/env vars. The reason I prefer bear is that it's a maintained version of what was committed to LLVM as intercept-build. If you want a rough approximation of compilation commands you can also run `cmake -DCMAKE_C_COMPILER=/usr/bin/clang -DCMAKE_CXX_COMPILER=/usr/bin/clang++ -DTARGET=arm64 -DTARGET_ARCH=aarch64 -DCMAKE_EXPORT_COMPILE_COMMANDS=ON`.

deadly-platypus:
This just makes generating a compilation database work with `cheribuild` again.  If this isn't the correct approach though, then go ahead and close this PR.

qwattash:
I previously attempted to use bear on the purecap kernel build and also found that it was failing, however IIRC it is due to some weird quoting, it does not seem to be a problem with `bear`.

qwattash:
I think the only issue with CheriBSD is that by overriding the value of `MAKE` to point to `tools/build/make.py` we break the FreeBSD build files because they assume that it is always a path and do weird things to it during bootstrapping.
I'm not convinced that we actually need to override `$MAKE` for `bear`, I attempted to just take it out and cheribsd builds fine.
I think the compilationdb is correct, at least in the case where a single kernel config is specified via `--cheribsd-xxx-yyy/kernel-config MY-KCONFIG`. I have not tested it with buildworld.

arichardson:
> I think the only issue with CheriBSD is that by overriding the value of `MAKE` to point to `tools/build/make.py` we break the FreeBSD build files because they assume that it is always a path and do weird things to it during bootstrapping. I'm not convinced that we actually need to override `$MAKE` for `bear`, I attempted to just take it out and cheribsd builds fine. I think the compilationdb is correct, at least in the case where a single kernel config is specified via `--cheribsd-xxx-yyy/kernel-config MY-KCONFIG`. I have not tested it with buildworld.

I think it used to be required since the builds system would fork off a new make that cleared the environment, but I believe bear will continue tracking now. I think we should revert the MAKE override and see what happens.

arichardson:
This should be fixed by https://github.com/CTSRD-CHERI/cheribuild/commit/b17cdb825dfd32e65776c487837d69a0ea2ebeba, please reopen (or file a new issue) if it still causes problems.

PR #359, opened 2023-03-30
Cleanup PEP 484 variable type annotations in favor of PEP 526.

qwattash:
Minor type annotation cleanup. Should not bring any functional change.


pycheribuild/config/chericonfig.py
```diff
@@ -292,11 +292,11 @@ def __init__(self, loader, action_class) -> None:
             help="When building with --include-dependencies ignore the SDK dependencies. Saves a lot of time "
                  "when building libc++, etc. with dependencies but the sdk is already up-to-date. "
                  "This is like --no-include-toolchain-depedencies but also skips the target that builds the sysroot.")
-        self.skip_dependency_filters = loader.add_option(
+        self.skip_dependency_filters: typing.List[re.Pattern] = loader.add_option(
```
arichardson, line 295:
```suggestion
        self.skip_dependency_filters: "list[re.Pattern]" = loader.add_option(
```

I've been trying to remove typing.List imports, so I'd suggest using the string version.

pycheribuild/projects/disk_image.py
```diff
@@ -198,16 +198,16 @@ def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
         # make use of the mtree file created by make installworld
         # this means we can create a disk image without root privilege
-        self.manifest_file = None  # type: Optional[Path]
-        self.extra_files: "list[Path]" = []
+        self.manifest_file: Optional[Path] = None
+        self.extra_files: List[Path] = []
```
arichardson, line 202: I would keep this then the import can be omitted.

arichardson (changes requested):
Thanks for fixing this.

pycheribuild/projects/disk_image.py
qwattash: If we really want to omit the import and still remove the `# type: ` comments perhaps this can become
```
        self.manifest_file: "Path|None" = None
        self.extra_files: "list[Path]" = []
```
Which I think is better because it eventually can be enabled without the quotes in more recent python versions.

qwattash: Oh perhaps you were only referring to the `List` import, I see.

arichardson (approved):

PR #361, opened 2023-04-20
cheribsd: Ensure that the kernel-config override only builds one kernel.

qwattash:
When the kernel-config is specified explicitly, extra kernels should not be built.

### DIFF
--- a/pycheribuild/config/config_loader_base.py
+++ b/pycheribuild/config/config_loader_base.py
@@ -151,6 +151,7 @@ class ConfigLoaderBase(ABC):
     def reset(self) -> None:
         for option in self.options.values():
             option._cached = None
+            option._is_default_value = False
 
     def debug_msg(self, *args, sep=" ", **kwargs) -> None:
         pass
--- a/pycheribuild/projects/cross/cheribsd.py
+++ b/pycheribuild/projects/cross/cheribsd.py
@@ -27,6 +27,7 @@
 # OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 # SUCH DAMAGE.
 #
+import inspect
 import itertools
 import os
 import shutil
@@ -55,6 +56,7 @@ from ..project import (
 )
 from ..simple_project import SimpleProject, TargetAliasWithDependencies, _clear_line_sequence, flush_stdio
 from ...config.compilation_targets import CompilationTargets, FreeBSDTargetInfo
+from ...config.loader import ConfigOptionBase
 from ...config.target_info import AutoVarInit, CrossCompileTarget
 from ...config.target_info import CompilerType as FreeBSDToolchainKind
 from ...processutils import latest_system_clang_tool, print_command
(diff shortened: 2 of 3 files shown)


pycheribuild/projects/cross/cheribsd.py
```diff
@@ -1678,6 +1678,8 @@ def default_kernel_config(self, platform: "Optional[ConfigPlatform]" = None, **f
 
     def extra_kernel_configs(self) -> "list[CheriBSDConfig]":
         # Everything that is not the default kernconf
+        if self.kernel_config != self.default_kernel_config():
```
jrtc27, line 1681: What if I do --cheribsd/kernel-config=DEFAULT_KERNEL_CONFIG_HERE? That should override it too, surely?

qwattash: Yes and it should, this should only affect the extra kernels that are added on top of the default kernel. I think that the `kernel-config` argument is already taken into account from the `default_kernel` side.

jrtc27: Today, `--cheribsd-morello-purecap/kernel-config=GENERIC-MORELLO cheribsd-morello-purecap` builds GENERIC-MORELLO and GENERIC-MORELLO-PURECAP, whilst `--cheribsd-morello-purecap/kernel-config=GENERIC-MORELLO-PURECAP cheribsd-morello-purecap` builds GENERIC-MORELLO-PURECAP twice

qwattash: Yeah I don't think that we should build the `GENERIC-MORELLO-PURECAP` in this case, if we specify a kernel-config to override. It feels like it is ignoring the override, and also building the kernel twice is wrong.

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -1678,6 +1678,8 @@ def default_kernel_config(self, platform: "Optional[ConfigPlatform]" = None, **f
...
+            return []
```
arichardson, line 1682: How about the following?
```suggestion
        option = inspect.getattr_static(self, "kernel_config")
        assert isinstance(option, ConfigOptionBase)
        if not option.is_default_value:
            return []
```

qwattash: Uhm this seems to break a bunch of tests. I need to understand exactly what is going on there.

qwattash: Upon further analysis, this seems to be tripping on an assertion because `option.is_default_value` can only be used after the `ConfigOptionBase._cached` has been loaded. The check in ConfigOptionBase may be conflating the _cached value with having loaded the value, this seems to be the case when no default is available, e.g. when has_default_buildkernel_kernel_config is set.

The test infrastructure seems to be triggering `ConfigOptionBase.load_option` multiple times, leading to the _is_default_value being retained erroneously. I'm not entirely sure how this is happening, it might be that the cached value is None and we load the value again. This however does not explain why it was marked as default before. This seems like a racy thing that happens only when multiple tests run.

arichardson: Hmm the logic to reset in the unit tests is somewhat hacky so it does sound like something is going wrong there...

arichardson (changes requested):
Change LGTM but I think the reset logic should be moved.

pycheribuild/config/config_loader_base.py
```diff
@@ -192,6 +192,9 @@ def __init__(self, name: str, shortname: Optional[str], default,
 
     def load_option(self, config: "ConfigBase", instance: "Optional[object]", _: type,
                     return_none_if_default=False) -> T:
+        # This may get called multiple times, so reset _is_default_value
+        self._is_default_value = False
```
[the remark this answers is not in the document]
qwattash: Should be fixed now.

arichardson (approved):

PR #368, opened 2023-07-12
seL4: Add new compilation targets

heshamelmatary:


pycheribuild/config/compilation_targets.py
```diff
@@ -761,6 +761,75 @@ def base_sysroot_targets(cls, target: "CrossCompileTarget", config: "CheriConfig
             assert False, "No support for building RTEMS for non RISC-V targets yet"
 
 
+class Sel4TargetInfo(_ClangBasedTargetInfo):
```
arichardson, line 764: You can avoid some duplication by inheriting from BaremetalClangTargetInfo.

pycheribuild/config/compilation_targets.py
```diff
@@ -784,6 +784,59 @@ def uses_softfloat_by_default(cls, xtarget: "CrossCompileTarget"):
         return xtarget.is_riscv(include_purecap=True)
 
 
+class Sel4TargetInfo(BaremetalClangTargetInfo):
+    shortname: str = "sel4"
+
+    @property
+    def cmake_system_name(self) -> str:
+        return "sel4"
```
arichardson, line 792: Is this actually a recognized value? If not you should use Generic so that CMake does the right thing.

heshamelmatary: It doesn't do any weird things for me, I get all the expected *sel4* directories and symlinks. What does Generic do?

arichardson: CMake will use the default baremetal settings (https://github.com/Kitware/CMake/blob/master/Modules/Platform/Generic.cmake) with "Generic", I can't see a file called sel4.cmake, so my guess would be that it will do the wrong thing.

See https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html#toolchain-files:
`If your target is an embedded system without an OS, set CMAKE_SYSTEM_NAME to “Generic.” `

pycheribuild/config/compilation_targets.py
```diff
@@ -784,6 +784,59 @@ def uses_softfloat_by_default(cls, xtarget: "CrossCompileTarget"):
...
+        from ..projects.cross.llvm import BuildCheriLLVM
+        return BuildCheriLLVM
+
+
+class Sel4MorelloTargetInfo(BaremetalClangTargetInfo):
+    shortname: str = "sel4"
+    uses_morello_llvm: bool = True
+
+    @property
+    def cmake_system_name(self) -> str:
```
arichardson, line 818: no need to override this it's the same as the base class now.

pycheribuild/config/compilation_targets.py
```diff
@@ -784,6 +784,59 @@ def uses_softfloat_by_default(cls, xtarget: "CrossCompileTarget"):
         return xtarget.is_riscv(include_purecap=True)
 
 
+class Sel4TargetInfo(BaremetalClangTargetInfo):
+    shortname: str = "sel4"
+
+    @property
+    def cmake_system_name(self) -> str:
+        return "Generic"
```
arichardson, line 792: no need to override this it's the same as the base class now.

pycheribuild/config/compilation_targets.py
```diff
@@ -784,6 +784,59 @@ def uses_softfloat_by_default(cls, xtarget: "CrossCompileTarget"):
...
+
+    @classmethod
+    def is_sel4(cls) -> bool:
+        return True
+
+    @classmethod
+    def triple_for_target(cls, target: "CrossCompileTarget", config: "CheriConfig", *, include_version: bool):
```
arichardson, line 799: Could you move this to the base class to avoid duplicating it twice?

arichardson (commented):
LGTM once the issues I've pointed out have been fixed.

arichardson:
CI failure looks unrelated. Change looks good to me now, although you aren't actually using it yet.

arichardson (commented):
Actually, can you use BaremetalFreestandingTargetInfo for sel4?

heshamelmatary:
> > > CI failure looks unrelated. Change looks good to me now, although you aren't actually using it yet.
> > 
> > 
> > You can actually build `compiler-rt-builtins-sel4-morello-aarch64` and `compiler-rt-builtins-sel4-riscv64` with this PR. Also if you add `-d` it will install "_-none-_-" tools symlinks required for building seL4. But there's still no seL4 target build yet, that I may add in the future.
> 
> True, but couldn't you use the generic baremetal compiler-rt target already? Adding the symlinks should also be possible without the new target info

Yeah I could just use baremetal. But some variants may need different flags (e.g., hard-float vs soft-float which is the default for bare metal?) or extensions as well. I am happy to wait for this PR until I decide if it's a good idea to build seL4 projects for cheribuild or not.

PR #369, opened 2023-07-13
Morello purecap kernel

bsdjhb:
TL;DR: We haven't been testing the Morello purecap kernel in CI since the end of December as the tests are booting the hybrid kernel instead.

### DIFF
--- a/pycheribuild/boot_cheribsd/__init__.py
+++ b/pycheribuild/boot_cheribsd/__init__.py
@@ -72,7 +72,10 @@ SUPPORTED_ARCHITECTURES = {x.generic_target_suffix: x for x in (CompilationTarge
                                                                 CompilationTargets.CHERIBSD_MORELLO_PURECAP,
                                                                 )}
 
-AUTOBOOT_PROMPT = re.compile(r"(H|, h)it \[Enter] to boot ")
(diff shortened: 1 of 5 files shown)


pycheribuild/boot_cheribsd/__init__.py
```diff
@@ -72,7 +72,7 @@
                                                                 CompilationTargets.CHERIBSD_MORELLO_PURECAP,
                                                                 )}
 
-AUTOBOOT_PROMPT = re.compile(r"(H|, h)it \[Enter] to boot ")
+AUTOBOOT_PROMPT = re.compile(r"((H|, h)it \[Enter] to boot |\[Space] to pause)")
```
bsdjhb, line 75: I should maybe add a comment as the 'Hit' version is what the boot loader uses when there is no lua or forth, ', hit' is the old string in lua, and '[Space]' is in the new lua string (and also I think in the old lua string).

jrtc27:
> since the end of December

... 2021

bsdjhb:
> > since the end of December
> 
> ... 2021

Oof, yeah, I had misparsed.  The last dev run which used purecap Morello kernels was https://ctsrd-build.cl.cam.ac.uk/job/CheriBSD-pipeline/job/dev/93.

PR #371, opened 2023-07-31
Stop using fixlinks tool during sysroot installation

arichardson:
CheriBSD has not installed broken absolute symlinks for many years, we can retire this tool now.


arichardson:
I'm pretty sure this is not needed anymore. I looked at a few Jenkins builds and they all print `fixed 0/217 symbolic links`

jrtc27 (approved):
I was actually staring at this the other day thinking "surely this is long dead" :)

PR #372, opened 2023-08-09
Restore boot_and_login default args

PeterRugg:
Hopefully trivial, but just thought I'd do a PR to be sure


PeterRugg:
(vcu118-run.py calls this function without the extra args)

arichardson (approved):

PR #373, opened 2023-08-10
Fix syzkaller

arichardson:


pycheribuild/projects/syzkaller.py
```diff
@@ -37,117 +37,128 @@
...
-        syz_remote_install = self.syzkaller_install_path() / "freebsd_mips64"
+        target_build = native_build / f"freebsd_{self.syz_arch}"
+        syz_remote_install = self.syzkaller_install_path() / f"freebsd_{self.syz_arch}"
 
         self.makedirs(syz_remote_install)
 
         self.install_file(native_build / "syz-manager", self.syzkaller_binary(), mode=0o755)
 
         if not self.config.pretend:
             # mips64_build does not exist if we preted, so skip
```
qwattash, line 146: Probably remove the stale comment or update it. `os.listdir` is also old, I'm not sure why I used it instead of pathlib, maybe this can be simplified.

qwattash (approved):
Looks good to me

PR #378, opened 2023-10-12
Require python 3.8

arichardson:
All older versions are EOL and none of the distributions we support
should have such an old version.


jrtc27:
db30c920b4e61149ea8ca0dd9a0eaf0ec93d5ff7 for some of this, though note the details in the commit message

Issue #377, opened 2023-09-29
Switch parity of --cheribsd/caprevoke-kernel


brooksdavis:
`--cheribsd/no-caprevoke-kernel` should select NOCAPREVOKE kernels and `--cheribsd/caprevoke-kernel` should be a no-opt beyond disabling the former option.

jrtc27:
5f0e1c0fac0a2c3f54b9f8252e572ceb2c770e7b

PR #382, opened 2023-10-31
caprevoke-kernel -> build-nocaprevoke-kernel

brooksdavis:
CheriBSD now includes revocation by default so invert the --cheribsd/caprevoke-kernel support in favor of a slightly awkward --cheribsd/nocaprevoke-kernel option which enables -NOCAPREVOKE kernels.

Addresses #377


brooksdavis:
Reworked to be build-nocaprevoke-kernel as suggested by @jrtc27

jrtc27 (approved):

PR #386, opened 2024-01-24
Pass compiler/linker flags to CMake for -native-hybrid

kwitaszczyk:
morello-llvm-native on CheriBSD/Morello requires -mabi=aapcs in CFLAGS. Otherwise, it would be compiled with -mabi=purecap that cc uses by default on CheriBSD with purecap world.

### DIFF
--- a/pycheribuild/config/target_info.py
+++ b/pycheribuild/config/target_info.py
@@ -860,6 +860,9 @@ class CrossCompileTarget:
         assert self.target_info_cls is not None
         return self.target_info_cls.is_native()
 
+    def is_native_hybrid(self) -> bool:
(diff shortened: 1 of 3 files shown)


pycheribuild/config/target_info.py
```diff
@@ -860,6 +860,9 @@ def is_native(self) -> bool:
         assert self.target_info_cls is not None
         return self.target_info_cls.is_native()
 
+    def is_native_hybrid(self) -> bool:
```
arichardson, line 863: Maybe we could use `is_native_libcompat()`? That would allow using it for e.g. lib32 with -m32 to test this on non-Morello HW.

PR #387, opened 2024-01-28
ensure 0644 perms for `/etc/pam.d/system`

cocoa-xu:
Hi, I noticed that the file permission of `/etc/pam.d/system` is incorrect when I was using the minimal disk image. It prompts that:

```
2024-01-28T07:59:02.046161+00:00 - login 83 - - in openpam_check_desc_owner_perms(): /etc/pam.d/system: insecure ownership or permissions
2024-01-28T07:59:02.059751+00:00 - login 83 - - pam_start(): System error
```

And if I log out the system, I can never log in back.

This PR tries to fix this issue by ensuring the file permission of `/etc/pam.d/system` to be `0644` when building the rootfs. After applying this patch, I can log out and log in again without any issue.

```
Logging in as root...
2024-01-28T09:27:25.194438+00:00 - login 83 - - login on console as root
exec /bin/sh
# ^D

CheriBSD/arm64 (Amnesiac) (ttyu0)

login: root
2024-01-28T09:32:54.297221+00:00 - login 86 - - login on ttyu0 as root
2024-01-28T09:32:54.305668+00:00 - login 86 - - ROOT LOGIN (root) ON ttyu0
#
```

### DIFF
--- a/pycheribuild/projects/disk_image.py
+++ b/pycheribuild/projects/disk_image.py
@@ -307,28 +307,28 @@ class BuildDiskImageBase(SimpleProject):
             fstab_contents += "/dev/gpt/swap none swap sw 0 0\n"
         fstab_contents += self.file_templates.get_fstab_template()
         self.create_file_for_image("/etc/fstab", contents=fstab_contents, 
-                                   mode=0o664, show_contents_non_verbose=True)
+                                   mode=0o644, show_contents_non_verbose=True)
 
         # enable ssh and set hostname
         # TODO: use separate file in /etc/rc.conf.d/ ?
         rc_conf_contents = self.file_templates.get_rc_conf_template().format(hostname=self.hostname)
         self.create_file_for_image("/etc/rc.conf", contents=rc_conf_contents, 
-                                   mode=0o664, show_contents_non_verbose=False)
+                                   mode=0o644, show_contents_non_verbose=False)
 
         cshrc_contents = self.file_templates.get_cshrc_template().format(SRCPATH=self.config.source_root,
                                                                          ROOTFS_DIR=self.rootfs_dir)
-        self.create_file_for_image("/etc/csh.cshrc", contents=cshrc_contents, mode=0o664)
+        self.create_file_for_image("/etc/csh.cshrc", contents=cshrc_contents, mode=0o644)
 
         # Basic .bashrc/.bash_profile template
         dot_bashrc_contents = self.file_templates.get_dot_bashrc_template().format(SRCPATH=self.config.source_root,
                                                                                    ROOTFS_DIR=self.rootfs_dir)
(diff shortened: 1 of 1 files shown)


jrtc27:
Doing this as a one-off for /etc/pam.d/system sounds wrong, surely we should set the right permissions for other files too, i.e. give create_file_for_image a sensible default?

cocoa-xu:
Hi @jrtc27, I'm not sure if give create_file_for_image a default would break other things, and it might be okay to add this one-off for the reasons below. 

Firstly, there're already quite a few one-offs in the codebase, such as:

https://github.com/CTSRD-CHERI/cheribuild/blob/af47e1d206eb89c2872ede22292669d23df4921b/pycheribuild/projects/disk_image.py#L356-L379

https://github.com/CTSRD-CHERI/cheribuild/blob/af47e1d206eb89c2872ede22292669d23df4921b/pycheribuild/projects/disk_image.py#L421

https://github.com/CTSRD-CHERI/cheribuild/blob/af47e1d206eb89c2872ede22292669d23df4921b/pycheribuild/projects/disk_image.py#L440

Meanwhile, there're also a dozen of calls to this function didn't specify the desired file permissions, i.e., 

https://github.com/CTSRD-CHERI/cheribuild/blob/af47e1d206eb89c2872ede22292669d23df4921b/pycheribuild/projects/disk_image.py#L309-329

And for these files that without being explicitly set permissions, they will be created by

https://github.com/CTSRD-CHERI/cheribuild/blob/af47e1d206eb89c2872ede22292669d23df4921b/pycheribuild/filesystemutils.py#L215

which sets default permissions based on current umask value.

And that leads to the final reason that, if the umask value is changed somewhere else during the build, and they depend on this default behaviour, then giving create_file_for_image a default could cause incorrect file permissions.

arichardson:
Since there aren't that many calls I think requiring the mode argument probably makes sense and avoids problems like this in the future

arichardson:
@cocoa-xu would you be able to update this PR with add an explicit argument added to the other calls as well?

cocoa-xu:
> @cocoa-xu would you be able to update this PR with add an explicit argument added to the other calls as well?

Sure thing! I'll do that in a minute

jrtc27:
You seem to have typo'ed 644 as 664 in some places

cocoa-xu:
Hi @jrtc27, sorry I was copying file permissions from the system image built by the script. They should all be corrected now, but please do let me know if there're any more mistakes, and I'll correct them as soon as possible.

And I guess this also suggests that these files have wrong file permissions set all the time, and we indeed need to explicitly specify their permissions.

arichardson (approved):
Thanks this looks good now, @jrtc27 do you agree?

I can remove the default argument in a future change.

jrtc27 (approved):
LGTM subject to cleaning up the commit message on squash

PR #389, opened 2024-02-07
Default to purecap kernels for CheriBSD

brooksdavis:
Switch the default kernel ABI to purecap


arichardson (approved):
Tests need updating otherwise LGTM

jrtc27 (approved):
Was initially concerned this would affect releng/23.11, but past Alex(?) was smart and explicitly passed --cheribsd/default-kernel-abi in Jenkinsfile so that's under CheriBSD's control already.

brooksdavis:
This is annoyingly more complicated that I'd hoped. Options like `--cheribsd/build-fett-kernels` and `--cheribsd/build-bench-kernels` don't have associated PURECAP kernels in the tree so it's not just a matter of flipping orders. It feels like those options want to ignore the default and I'm not sure what the right solution is.

Issue #391, opened 2024-03-08
GMP mercurial host blocks certain IP ranges


dblane-digicatapult:
in `cheribuild/pycheribuild/projects/cross/gmp.py` we are using MercurialRepository with a repository of `repository = MercurialRepository("https://gmplib.org/repo/gmp")` However, the host for this (quite rightly) blocks certain large IP ranges, namely all of Azure and Github Actions runners, this causes any build that has GMP as a dependency to fail.

There is a well maintained mirror hosted on github at `https://github.com/gmp-mirror/gmp` which maintains the entire hg repositories commit history and contains some additional branches.

Would it be possible for us to switch to using the mirror instead?

qwattash:
I don't have anything against it. The commits appear to be in sync on the main branch.

jrtc27:
My recollection was that either such a git mirror didn’t exist or it was outdated back when I added gmp as a mercurial repo. Using git is certainly easier (the MercurialRepository implementation is quite crude and easy to break if you make local changes, but good enough for our use case) provided we know that we can rely on the mirror going forwards. There’s also the issue of needing to convert the user’s checkout from mercurial to git, which I don’t think needs to be any more complex than informing the user (in the same way as we do for switching branches) and then wiping the directory if it’s got a .hg subdirectory, as I highly doubt users have local changes (if they do they should save them elsewhere before saying yes to the prompt).

PR #392, opened 2024-03-11
swapped mercurial for git

Ellenn-A:
# Changes 
#391 
in `cheribuild/pycheribuild/projects/cross/gmp.py` swapps `repository = MercurialRepository("https://gmplib.org/repo/gmp")` for `repository = GitRepository("https://github.com/gmp-mirror/gmp")`

### Reasoning:
in cheribuild/pycheribuild/projects/cross/gmp.py we are using MercurialRepository with a repository of repository = MercurialRepository("https://gmplib.org/repo/gmp") However, the host for this (quite rightly) blocks certain large IP ranges, namely all of Azure and Github Actions runners, this causes any build that has GMP as a dependency to fail.

There is a well maintained mirror hosted on github at https://github.com/gmp-mirror/gmp which maintains the entire hg repositories commit history and contains some additional branches.


jrtc27:
This is the easy bit. The hard bit is dealing with the fact that many users will have an existing gmp directory that's a mercurial checkout and needs replacing with a git one. This patch does not address that.

arichardson:
The easiest approach might just be to use a different directory name (e.g. libgmp) and keep the mercurial checkout untouched

arichardson (approved):
LGTM, I'll merge this with an updated commit message once CI is happy.

PR #394, opened 2024-04-17
Test tidying

brooksdavis:
A bit of reformatting to simplify later diffs plus pacifying newer ruff versions.


brooksdavis:
@arichardson ruff seems to be broken. The first commit here fails when it passed before and the second is the fixes suggested by --fix with ruff from home-brew which also fails :(

arichardson:
> @arichardson ruff seems to be broken. The first commit here fails when it passed before and the second is the fixes suggested by --fix with ruff from home-brew which also fails :(

I think the problem is that CI pulls the latest version of ruff - we should pin it to a fixed version.

PR #395, opened 2024-04-17
Try pinning ruff to an earlier version

brooksdavis:


brooksdavis:
It looks like this has no effect (maybe not until after it's merged?)

arichardson:
Should be fixed by https://github.com/CTSRD-CHERI/cheribuild/pull/396

PR #398, opened 2024-05-15
Build Morello benchmark ABI kernels.

qwattash:
This is inhibited by the no-build-alternate-abi-kernels.

### DIFF
--- a/pycheribuild/projects/cross/cheribsd.py
+++ b/pycheribuild/projects/cross/cheribsd.py
@@ -90,6 +90,7 @@ class KernelABI(Enum):
     NOCHERI = "no-cheri"
     HYBRID = "hybrid"
     PURECAP = "purecap"
+    PURECAP_BENCHMARK = "purecap-benchmark"
 
 
 class ConfigPlatform(Enum):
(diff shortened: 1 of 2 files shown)


jrtc27:
Won't this break building older CheriBSD, including in CI for releng/23.11?

qwattash:
> Won't this break building older CheriBSD, including in CI for releng/23.11?

Right.. I wanted to avoid adding another option specifically for benchmark ABI kernels, but maybe I need a different flag to disable building benchmark ABI kernels. I'm leaning towards building it by default but I could make it off-by-default.

qwattash:
> > > Won't this break building older CheriBSD, including in CI for releng/23.11?
> > 
> > 
> > Right.. I wanted to avoid adding another option specifically for benchmark ABI kernels, but maybe I need a different flag to disable building benchmark ABI kernels. I'm leaning towards building it by default but I could make it off-by-default.
> 
> It needs to be off by default unfortunately, and enabled in CheriBSD CI, but we can flip the default once 24.0x is out.

This version should fix it. It is now enabled with `--cheribsd/build-benchmark-abi-kernels`. There is one test failing due to MFS root options, although I'm puzzled because it seems that it uses the riscv64 target and the option should not be available there (as expected). This seems to be a bug in the way options are inherited and registered. I'm quite confused by the condition here and the use of `fallback_replaceable`. My understanding is that if the current `target` is not in `only_add_for_targets` list we should just bail and not add the option, e.g. just `return None`. However this seems a very intentional workaround to a corner case.
@arichardson Do you have any clue?

https://github.com/CTSRD-CHERI/cheribuild/blob/afea91fcaae4417835ed27bd3d5c84baec5c5c3f/pycheribuild/projects/simple_project.py#L981-L985

jrtc27 (approved):

PR #404, opened 2024-08-08
Add build for B-lang-org/bsc-contrib repo as 'bluespec-libraries' target, also add 'bluespec' alias

tmarkettos:
'bluespec-libraries' builds and installs the B-lang-org/bsc-contrib repo into the local install.

'bluespec' is a dummy target which depends on 'bluespec-libraries' which depends on 'bluespec-compiler', so building 'bluespec' with dependencies builds both.

Upstream bluespec-compiler's makefile tends to rebuild all its BSV files even if nothing changed, so building 'bluespec-libraries' also causes the BSV inside bluespec-compiler to get rebuilt.  Not sure if there's an easy solution for this.


arichardson (approved):
LGTM if it works for you

PR #405, opened 2024-08-19
gmake: bootstrap to fetch gnulib before configuring

nwf:
Without this, the build dies with

```
cd /cheri/source/mainline/gmake && /cheri/source/mainline/gmake/autogen.sh
/cheri/source/mainline/gmake/autogen.sh: 1088: cannot open gnulib/gnulib-tool: No such file
Fatal error (in target gmake-aarch64): Command `/cheri/source/mainline/gmake/autogen.sh` failed with non-zero exit code 2. Working directory was /cheri/source/mainline/gmake
```

### DIFF
--- a/pycheribuild/projects/cross/gmake.py
+++ b/pycheribuild/projects/cross/gmake.py
@@ -54,3 +54,7 @@ class BuildGmake(CrossCompileAutotoolsProject):
             # when the locale is something like en_GB.UTF-8. C.UTF-8 seems to
             # work fine, not just C, so use that for the build.
             self.make_args.set_env(LC_ALL="C.UTF-8")
+
+    def configure(self, **kwargs):
+        self.run_cmd(self.source_dir / "bootstrap", cwd=self.source_dir)
+        super().configure(**kwargs)


arichardson (commented):
Hmm I think this should probably be inside `configure()`? ideally `setup()` should not modify the source dir.

pycheribuild/projects/cross/gmake.py
```diff
@@ -54,3 +54,5 @@ def setup(self):
             # when the locale is something like en_GB.UTF-8. C.UTF-8 seems to
             # work fine, not just C, so use that for the build.
             self.make_args.set_env(LC_ALL="C.UTF-8")
+
+        self.run_cmd(self.source_dir / "bootstrap", cwd=self.source_dir)
```
arichardson, line 58:
```suggestion

    def configure(self, **kwargs)
        self.run_cmd(self.source_dir / "bootstrap", cwd=self.source_dir)
        super().configure()
```

nwf:
Even more ideally that'd work in the build dir, but it does not, so source dir modification in configure() it is.

arichardson (approved):

PR #406, opened 2024-10-22
cheribsd: use llvm-link from the SDK

kwitaszczyk:
llvm-link is used by the kernel-c18n branch to generate an IR file of the kernel.

A corresponding change making use of this is in https://github.com/CTSRD-CHERI/cheribsd/commit/ae7327c781363ae6885e9716971c3052716197a6.


jrtc27 (approved):
I didn't realise FreeBSD already could use it

PR #409, opened 2025-01-31
Remove an obsolete hack from the cheribsd buildworld target

markjdb:
I believe this will fix the last CI failure in https://github.com/CTSRD-CHERI/cheribsd/pull/2309

### DIFF
--- a/pycheribuild/projects/cross/cheribsd.py
+++ b/pycheribuild/projects/cross/cheribsd.py
@@ -45,7 +45,6 @@ from ..project import (
     BuildType,
     CheriConfig,
     ComputedDefaultValue,
-    CPUArchitecture,
     DefaultInstallDir,
     GitRepository,
     MakeCommandKind,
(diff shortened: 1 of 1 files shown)


jrtc27 (approved):
Please merge so long as 24.05 cheribsd-riscv64-hybrid still builds without issue

jrtc27:
Ah, please also fix the flake8 errors, it seems this was the only use of CPUArchitecture so the import is now unused

markjdb:
I tested building cheribsd 24.05 with this patch applied (specifically the `cheribsd-riscv64-hybrid` target), no problems seen.

I'm not sure what the CI failures are about. The run_basic_tests.sh script doesn't show any failures for me on FreeBSD.

arichardson:
> I tested building cheribsd 24.05 with this patch applied (specifically the `cheribsd-riscv64-hybrid` target), no problems seen.
> 
> I'm not sure what the CI failures are about. The run_basic_tests.sh script doesn't show any failures for me on FreeBSD.

Yeah I'm not sure why macos CI is returning values in a different order - the sorting *should* be deterministic. Haven't figured this one out yet...

jrtc27:
Oh that's because our comparison function for targets isn't a proper order so it isn't well-defined to use sort on it.

arichardson:
> Oh that's because our comparison function for targets isn't a proper order so it isn't well-defined to use sort on it.

That makes sense, although I'm pretty sure I tried to ensure that initially.

PR #410, opened 2025-02-03
Add libmd.so.7 to minimal CheriBSD images

bsdjhb:
This is required by libcrypt.so.5 after CheriBSD commit 4ca6f578846d947386821de5e740062c83f8376a.

### DIFF
--- a/pycheribuild/projects/disk_image.py
+++ b/pycheribuild/projects/disk_image.py
@@ -1226,6 +1226,10 @@ class BuildMinimalCheriBSDDiskImage(BuildDiskImageBase):
         ]
         # required, but only recently added so not always present
         optional_libs += ["libsys.so.7"]
+        # required for libcrypto.so.5 versions newer than CheriBSD
+        # commit 4ca6f578846d947386821de5e740062c83f8376a, but
+        # recently bumped from libmd.so.6 so not always present
+        optional_libs += ["libmd.so.7"]
         # required, but versions were bumped with changes to ncurses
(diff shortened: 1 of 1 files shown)


pycheribuild/projects/disk_image.py
```diff
@@ -1226,6 +1226,9 @@ def add_required_libraries(self, libdirs: "list[str]", ignore_required: bool = F
         ]
         # required, but only recently added so not always present
         optional_libs += ["libsys.so.7"]
+        # required for libcrypto.so.5 versions newer than CheriBSD
+        # commit 4ca6f578846d947386821de5e740062c83f8376a
```
jrtc27, line 1230: Can you add to the comment mentioning the soname bump to 7 being recent hence why it’s optional? (The fact libcrypto optionally needs it isn’t a reason to put it in optional, because if it exists then it’ll be included)

bsdjhb:
I don't understand the CI failures.  They don't seem to be related to this change?  On the current `main` branch there are different CI failures for some (but not all) of the same jobs.

arichardson:
> I don't understand the CI failures. They don't seem to be related to this change? On the current `main` branch there are different CI failures for some (but not all) of the same jobs.

One of them should be fixed by https://github.com/CTSRD-CHERI/cheribuild/commit/e4ec915df0ca0692182cbb5edbd7c6ebe555fe4c - the other failures I've seen are not reproducible for me.

jrtc27 (approved):

PR #413, opened 2025-03-07
cheribsd-release: build vm images

brooksdavis:
In conjunction with the changes in
https://github.com/CTSRD-CHERI/cheribsd/pull/2360 this produces ciimages/FreeBSD-15.0-CURRENT-arm64-aarch64c-BASIC-CI-ufs.raw.xz in the output directory for cheribsd-release-morello-purecap.  The result is untested, but likely a usable starting point if cross build actually works.

I think we'll likely require some manual testing before this lands or it needs to got under a switch, but it should let people try it out.


brooksdavis:
Yeah, a bit more poking shows there's no METALOG use in the disk image so ownership will all be wrong so I need to clean that up on the CheriBSD side (or see if the foundation has a WIP far enough along to help).

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -2179,6 +2179,18 @@ def supported_architectures(self) -> "tuple[CrossCompileTarget, ...]":
 
 
 class BuildFreeBSDReleaseMixin(ReleaseMixinBase):
+    @classmethod
+    def setup_config_options(cls, kernel_only_target=False, install_directory_help=None, **kwargs) -> None:
+        super().setup_config_options(
+            use_upstream_llvm=False,
+            kernel_only_target=kernel_only_target,
+        )
```
jrtc27, line 2187:
```suggestion
    def setup_config_options(cls, **kwargs) -> None:
        super().setup_config_options(**kwargs)
```

surely?

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -2179,6 +2179,18 @@ def supported_architectures(self) -> "tuple[CrossCompileTarget, ...]":
...
+        cls.build_vm_images = cls.add_bool_option(
+            "build-vm-images",
+            _allow_unknown_targets=True,
+            help="Build vm images with releases.",
```
jrtc27, line 2191:
```suggestion
            help="Build VM images with releases.",
```

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -2244,6 +2256,10 @@ def process(self) -> None:
         release_args.set_env(INSTALL="sh " + str(self.source_dir / "tools/install.sh"))
         release_args.set_env(XZ_CMD=str(self.objdir / "tmp/legacy/usr/bin/xz -T 0"))
 
+        if self.build_vm_images:
+            release_args.set_env(WITH_CLOUDWARE="yes")
+            release_args.set_env(WITHOUT_QEMU="yes")
```
jrtc27, line 2261: Do these need to be in the env, or does set_with_options(FOO=False) work?

jrtc27, line 2261: Is there a reason not to always pass WITHOUT_QEMU?

brooksdavis: Switched

PR #414, opened 2025-03-12
Stop generating DSA SSH keys

markjdb:
They are obsolete and on sufficiently new versions of FreeBSD support for them is compiled out, resulting in an error from ssh-keygen.


bsdjhb (approved):

arichardson:
Thanks, somehow missed the notification for this PR.

PR #416, opened 2025-03-20
CheriFreeRTOS hmka2 merge

heshamelmatary:
This is a first cut at trying to merge my old CheriFreeRTOS hmka2 branches into main so that others can use it with the latest toolchain, and validate/build on top of it. It could also be useful for CI to build, run, and test something quick (e.g., with toolchain updates, QEMU, or FPGA), and for other parties wanting to submit fixes or features on top of our CheriFreeRTOS.

It should be able to build and run vanilla RISC-V FreeRTOS and purecap CheriFreeRTOS on many targets, with lots of demos, ranging from simple `main_blinky` that only prints a few messages, to full OTA demos and CLI using networking, VirtIO, ethernet, etc. This merge will lack support for compartmentalisation, however, as it requires a specific `gprel` LLVM branch. I left `gprel` conditional there in case someone might want to still use it.


pycheribuild/config/chericonfig.py
```diff
@@ -328,6 +328,17 @@ def __init__(self, loader, action_class: "type[CheribuildActionEnum]") -> None:
             group=loader.cross_compile_options_group,
             help="The floating point/SIMD mode to use for building AArch64 programs",
         )
+
+        self.riscv_baremetal_hardfloat = loader.add_bool_option("riscv-baremetal-hardfloat", default=False,
+            group=loader.cross_compile_options_group,
+            help="Use hard floating point ABI for building CHERI-RISC-V programs")
+        self.riscv_cheri_rvc = loader.add_bool_option("riscv-cheri-rvc", default=False,
```
arichardson, line 335: C is the default behaviour, do you need this diff?

heshamelmatary: @arichardson Not anymore, I added this option when it wasn't the default. I have deleted this diff now.

PR #421, opened 2025-06-04
qemu: silence int-conversion compiler warning

Flowdalic:
Allows compilation with GCC 14.


arichardson:
Where are you seeing this error? We should fix this in qemu instead.

Flowdalic:
> We should fix this in qemu instead.

See https://github.com/CTSRD-CHERI/qemu/pull/281

PR #422, opened 2025-06-20
cheribsd-release: allow to use custom configs

kwitaszczyk:
Use the same routines to get a list of kernel configurations to include in release files as is used when building a CheriBSD rootfs.

This can be either --kernel-config or --cheribsd-release-*/kernel-config.

### DIFF
--- a/pycheribuild/projects/cross/cheribsd.py
+++ b/pycheribuild/projects/cross/cheribsd.py
@@ -2209,7 +2209,7 @@ class BuildFreeBSDReleaseMixin(ReleaseMixinBase):
 
         # release/Makefile needs to install both world and kernel to create
         # images, so start with the arguments for the combination of the two.
-        kernconfs = self.get_kernel_configs(None)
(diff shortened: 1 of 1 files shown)


arichardson (approved):
Not sure what the difference here is but if it fixes things for you LGTM.

jrtc27:
Looking at this further, get_kernel_configs is public API for "get me the list of kernel configs that the project is configured to build (and thus built, in the case of most uses for it) for the given platform" so that something like QEMU can ask that question. Whereas kernconf_list is "get me the full list of kernel configs to build". So yes, kernconf_list is probably the more appropriate API to use in this case, as that's what gets used for buildkernel, but at the same time it's also a bug that it doesn't honour the kernel-config option (presumably `cheribuild {cheribsd,disk-image,run}-riscv64-purecap --cheribsd-riscv64-purecap/kernel-config=MY-CONFIG` doesn't pass the right `-kernel` to QEMU).

PR #423, opened 2025-06-23
disk-image-{minimal,mfs-root}: Include malloc libraries needed by cheribsdtest

PeterRugg:
Seems to make the following tests pass:

Otherwise fail:
```
malloc_revocation_ctl_baseline: child exec failed
malloc_revocation_ctl_elfnote_disable: child exec failed                                                             
malloc_revocation_ctl_elfnote_enable: child exec failed
malloc_revocation_ctl_elfnote_disable_protctl_enable: child exec failed             
malloc_revocation_ctl_elfnote_enable_protctl_disable: child exec failed
malloc_revocation_ctl_suid_baseline: child exec failed                                                               
malloc_revocation_ctl_suid_elfnote_disable: child exec failed
malloc_revocation_ctl_suid_elfnote_enable: child exec failed
```
Otherwise skipped:
```
malloc_early_constructor: couldn't stat helper
```

It also advances the following, though these seem to still fail for other reasons on Toooba: they now report `child is revoking and should not be` and vice versa:
```
malloc_revocation_ctl_suid_elfnote_disable_protctl_enable:  child exec failed
malloc_revocation_ctl_suid_elfnote_enable_protctl_disable:  child exec failed
```

### DIFF
--- a/pycheribuild/projects/disk_image.py
+++ b/pycheribuild/projects/disk_image.py
@@ -1176,13 +1176,14 @@ class BuildMinimalCheriBSDDiskImage(BuildDiskImageBase):
                 self.add_required_libraries(["usr/" + libcompat_dir], ignore_required=ignore_required)
 
         if self.include_cheribsdtest:
-            for test_binary in (self.rootfs_dir / "bin").glob("cheribsdtest-*"):
-                self.add_file_to_image(test_binary, base_directory=self.rootfs_dir)
-            # Libraries needed to run cheribsdtest
-            for test_lib in (self.rootfs_dir / "usr/libexec").glob("malloc_revoke_enabled*"):
-                self.add_file_to_image(test_lib, base_directory=self.rootfs_dir)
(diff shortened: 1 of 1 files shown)


brooksdavis:
It looks like you need to add a mode=0o4555 argument to the add_file_to_image call when the filename contains suid.

PeterRugg:
Can do! It looks like that doesn't actually fix it though, I suspect because we're running as root anyway on MFS builds?

PeterRugg:
Yeah, I just confirmed that running with owner set to 1 means the tests pass. Does `mode=0o4555` carry out the change to owner nobody, or is there another argument to add to the call to fix it?

PeterRugg:
Ah, `user="nobody"`. Cool. I'll check that all works. Thanks for your help!

arichardson:
Is this still needed or can we close it?

jrtc27:
Fixed as of 20c34cd253be583990247cdd5f6b52fb4429f668 (and earlier commits re metadata)

PR #424, opened 2025-06-27
Introduce CHERI-Alliance SDK and start using it for some targets.

qwattash:

PR #428, opened 2025-07-25
Introduce CHERI-Alliance SDK and start using it for some targets

heshamelmatary:
Picked up from https://github.com/CTSRD-CHERI/cheribuild/pull/424 and adding two seL4 targets. This PR is just mainly for CI testing

### DIFF
--- a/pycheribuild/projects/build_qemu.py
+++ b/pycheribuild/projects/build_qemu.py
@@ -579,7 +579,7 @@ class BuildQEMU(BuildQEMUBase):
 class BuildCheriAllianceQEMU(BuildQEMUBase):
     target = "cheri-alliance-qemu"
     repository = GitRepository("https://github.com/CHERI-Alliance/qemu.git",
-                               default_branch="codasip-cheri-riscv_v3",
+                               default_branch="codasip-cheri-riscv.25-03-31",
                                force_branch=True)
     native_install_dir = DefaultInstallDir.CHERI_ALLIANCE_SDK
     default_targets = (


pycheribuild/projects/cross/llvm_test_suite.py
```diff
@@ -63,11 +63,7 @@ def dependencies(cls, config) -> "tuple[str, ...]":
     @classproperty
     def llvm_project(self) -> "type[BuildLLVMBase]":
         target_info = self.get_crosscompile_target().target_info_cls
-        if issubclass(target_info, FreeBSDTargetInfo):
-            # noinspection PyProtectedMember
-            return target_info._get_compiler_project()
-        else:
-            return BuildCheriLLVM
+        return BuildCheriLLVM
```
arichardson, line 66: This looks wrong, should just pass `self.config` to _get_compiler_project()

qwattash: I think this is confusing. The method is a classproperty but it is using `self` as the argument name, which violates PEP8 and should be named `cls`.

Apart from this nitpick, I think we need to deal with the fact that the LLVM project can't be known without a configuration object at this stage.

qwattash: Ok I pushed a fix, it is unclear to me at this point whether we need to patch the compiler_rt and libcxx targets, which seem to use a similar pattern.

pycheribuild/projects/cross/opensbi.py
```diff
@@ -213,19 +216,42 @@ def run_tests(self):
 class BuildAllianceOpenSBI(BuildOpenSBI):
     target = "cheri-alliance-opensbi"
     _default_install_dir_fn = ComputedDefaultValue(
-        function=lambda config, p: config.cheri_sdk_dir / "cheri-alliance-opensbi/riscv64",
-        as_string="$SDK_ROOT/cheri-alliance-opensbi/riscv64",
+        function=lambda config, p: config.cheri_alliance_sdk_dir / "cheri-alliance-opensbi/riscv64",
```
arichardson, line 219:
```suggestion
        function=lambda config, p: config.cheri_alliance_sdk_dir / "opensbi/riscv64",
```
probably don't need a double prefix here?

pycheribuild/projects/build_qemu.py
```diff
@@ -574,3 +574,53 @@ def install(self, **kwargs):
...
+    target = "cheri-alliance-qemu"
+    repository = GitRepository("https://github.com/CHERI-Alliance/qemu.git",
+                               default_branch="codasip-cheri-riscv_v3",
+                               force_branch=True)
+    native_install_dir = DefaultInstallDir.CHERI_ALLIANCE_SDK
+    default_targets = (
+        "riscv64-softmmu,riscv64cheri-softmmu,riscv32-softmmu,riscv32cheri-softmmu,"
+    )
+    # Turn on unaligned loads/stores by default
+    unaligned = BoolConfigOption("unaligned", show_help=False, help="Permit un-aligned loads/stores", default=False)
```
arichardson, line 589: I think we can drop this one. This was just for MIPS.

pycheribuild/config/compilation_targets.py
```diff
@@ -1025,8 +1033,12 @@ class BaremetalFreestandingTargetInfo(BaremetalClangTargetInfo):
     os_prefix: str = "baremetal-"
 
     @classmethod
-    def _get_compiler_project(cls) -> "type[BuildLLVMInterface]":
-        return typing.cast("type[BuildLLVMInterface]", SimpleProject.get_class_for_target_name("llvm", None))
+    def _get_compiler_project(cls, config: CheriConfig) -> "type[BuildLLVMInterface]":
+        if config.riscv_cheri_isa == RiscvCheriISA.STD:
+            llvm_target = SimpleProject.get_class_for_target_name("cheri-alliance-llvm", None)
```
arichardson, line 1038: Won't this break things like building for arm since we only build the RISC-V and host backends in CHERI Alliance LLVM?

pycheribuild/projects/cross/opensbi.py
qwattash: Yeah, good point.

pycheribuild/projects/cross/opensbi.py
```diff
@@ -213,19 +216,42 @@ def run_tests(self):
...
+        self.run_cmd(
+            self.sdk_bindir / "llvm-objcopy",
+            "-S",
+            "-O",
+            "binary",
+            self._fw_jump_path(),
+            qemu_fw_dir / "opensbi-riscv64cheri-virt-fw_jump.bin",
+            print_verbose_only=False,
+        )
+
```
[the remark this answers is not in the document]
qwattash: I think this is redundant at this point. The suffix should probably change depending on the Zcheri ISA switch.
On a separate note, the name and suffix should be kept in sync with both our QEMU and the Cheri Alliance QEMU branch. I wonder whether we can configure this as part of the build, the docs mention an `export OPENSBI=xxx`.

pycheribuild/projects/build_qemu.py
qwattash: Good point.

pycheribuild/projects/cross/opensbi.py
```diff
@@ -140,7 +140,11 @@ def install(self, **kwargs):
         if self.crosscompile_target.is_hybrid_or_purecap_cheri() and not self.build_dir_suffix:
             # Install into the QEMU firware directory so that `-bios default` works
             qemu_fw_dir = self._qemu_install_dir() / "share/qemu/"
-            suffix = "cheristd" if self.crosscompile_target.is_cheri_purecap() else "cheri"
+            suffix = ""
+            if self.crosscompile_target.is_cheri_purecap():
+                suffix = "cheri"
+                if self.config.riscv_cheri_isa == RiscvCheriISA.STD:
+                    suffix += "std"
```
heshamelmatary, line 147: Isn't anything in "CHERI-Alliance SDK" supposed to be standard CHERI-RISC-V by default? Do we expect to build other non-std CHERI-RISC-V with this SDK at all?

qwattash:
> @qwattash Also QEMU will fail to build without your local fix. So we should probably get your fix merged with the Codasip's branch, or use our own (rebased?) with that fix (and I have another fix as well).

Yes, I pushed a PR there https://github.com/CHERI-Alliance/qemu/pull/6.

heshamelmatary:
@qwattash 

> > @qwattash Also QEMU will fail to build without your local fix. So we should probably get your fix merged with the Codasip's branch, or use our own (rebased?) with that fix (and I have another fix as well).
> 
> Yes, I pushed a PR there [CHERI-Alliance/qemu#6](https://github.com/CHERI-Alliance/qemu/pull/6).

This is against `codasip-cheri-riscv.25-03-31` branch, while this PR uses `codasip-cheri-riscv-v3`. It's a bit confusing which one we should use, and which is the latest and has all features/fixes we want. Should we agree on [creating] some branches this PR could use that will be known to work fine for us?

pycheribuild/config/compilation_targets.py
arichardson: Hmm I guess so, I was thinking about things like the picolibc target where aarch32 and aarch64 are supported and use CHERI LLVM.

arichardson: Maybe I should just change those to use upstream LLVM?

arichardson (approved):

qwattash:
> @qwattash
> 
> > > @qwattash Also QEMU will fail to build without your local fix. So we should probably get your fix merged with the Codasip's branch, or use our own (rebased?) with that fix (and I have another fix as well).
> > 
> > 
> > Yes, I pushed a PR there [CHERI-Alliance/qemu#6](https://github.com/CHERI-Alliance/qemu/pull/6).
> 
> This is against `codasip-cheri-riscv.25-03-31` branch, while this PR uses `codasip-cheri-riscv-v3`. It's a bit confusing which one we should use, and which is the latest and has all features/fixes we want. Should we agree on [creating] some branches this PR could use that will be known to work fine for us?

I think I should update the qemu version we are using here. In general I think I want to converge qemu sooner rather than later, I will bring up having a separate branch in the WG meeting, unfortunately I have had no spare time to look into cleaning it up.

PR #430, opened 2025-08-12
Add CHERI-seL4 and its exercises targets

heshamelmatary:
The last 2 commits are fixes when I tried to reproduce CI errors on my local machine (AArch64), I could just drop them but thought could be useful in the future

### DIFF
--- a/pycheribuild/projects/cross/cheri_sel4_exercises.py
+++ b/pycheribuild/projects/cross/cheri_sel4_exercises.py
@@ -1,0 +1,483 @@
+# SPDX-License-Identifier: BSD-2-Clause
+#
+# Author: Hesham Almatary <Hesham.Almatary@cl.cam.ac.uk>
+#
+# THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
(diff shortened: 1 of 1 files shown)


pycheribuild/projects/cross/cheri_microkit.py
```diff
@@ -0,0 +1,304 @@
...
+    def needs_configure(self):
+        return False
+
+    def compile(self, **kwargs):
+        pass
+
+    def install(self, **kwargs):
+        pass
+
+    def process(self):
```
qwattash, line 75: Same as `__init__`

pycheribuild/projects/cross/cheri_microkit.py
```diff
@@ -0,0 +1,304 @@
...
+        CompilationTargets.FREESTANDING_RISCV64_PURECAP,
+        CompilationTargets.FREESTANDING_RISCV64,
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def setup(self):
+        super().setup()
+
```
qwattash, line 106: This is redundant, you can assume that the super implementation will be called automatically.

```suggestion
```

pycheribuild/projects/cross/cheri_microkit.py
```diff
@@ -0,0 +1,304 @@
...
+    @classmethod
+    def setup_config_options(cls, **kwargs):
+        super().setup_config_options(**kwargs)
+        cls.configs: str = typing.cast(
```
qwattash, line 110: I'm confused by the need to typing-cast here. I think the `add_config_option` method returns a handle for the option, not really the base type?

pycheribuild/projects/cross/cheri_sel4_exercises.py
```diff
@@ -0,0 +1,606 @@
...
+                "build_all",
+                show_help=True,
+                default=False,
+                help="Build all CHERI exercises *and* missions.",
+            ),
+        )
+
+    def setup(self):
+        super().setup()
+
```
qwattash, line 120: Unnecessary
```suggestion
```

pycheribuild/projects/cross/cheri_sel4_exercises.py
```diff
@@ -0,0 +1,606 @@
...
+                kernel_file=kernel,
+            )
+            self.run_cmd(cmd)
+        else:
+            self.fatal(
+                f"Can't run tests: neither valid exercise '{self.exercise}' nor mission '{self.mission}'."
+            )
+
+    def process(self) -> None:
+        super().process()
```
qwattash, line 606: Unnecessary
```suggestion
```

pycheribuild/projects/cross/cheri_microkit.py
qwattash: I'm still a bit unsure about this because my experience with the type checker is limited. @arichardson do you have an opinion on this?

PR #433, opened 2025-09-11
Morello/baremetal: match Freestanding's sysroot dir

heshamelmatary:
NewlibBaremetalTargetInfo targets should add the target triple to the sysroot, as this was the layout expected by Newlib’s multilib install layout (and GCC's sysroot). The current MorelloBaremetal targets shouldn't assume a Newlib/GCC sysroot and should follow the FreeStanding target one. If there's a need for GCC's sysroot and multilib layout, this should go to a new NewlibMorelloBaremetalTargetInfo or something.


arichardson (approved):
LGTM, not sure why it was this way before.

heshamelmatary:
@arichardson OK to merge?

PR #434, opened 2025-09-12
Linux: Add a target for building and running

heshamelmatary:
This commit adds preliminary targets for cross compiling the Linux kernel (and nothing else, without userspace), using LLVM, for AArch64 and RISC-V. Unlike CheriBSD, Linux's userspace is scattered and depends on the project. For instance, the same kernel built here could be used as a VM, for Yocto, Debian, BusyBox, etc, so it's left to those projects to build, package, and run userspace with this kernels.

* Morello: fetched from Arm's repos and only using hybrid kernel
* CHERI-Alliance: only builds a purecap

A simple QEMULauch class is also added to run the kernels on QEMU but without userspace. It is still good to as a test eg to see if the kernel boots to the end and tries to drop to userspace.

### DIFF
--- a/pycheribuild/projects/cross/linux.py
+++ b/pycheribuild/projects/cross/linux.py
@@ -1,0 +1,235 @@
+#
+# Copyright (c) 2025 Hesham Almatary
+#
+# This software was developed by SRI International and the University of
+# Cambridge Computer Laboratory (Department of Computer Science and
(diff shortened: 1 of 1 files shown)


heshamelmatary:
https://github.com/CTSRD-CHERI/cheribuild/pull/436: Close and target `main` now that RVY support is merged

PR #435, opened 2025-09-16
New PR for RVY 0.9.3 support

arichardson:


pycheribuild/projects/cross/llvm.py
```diff
@@ -762,17 +764,58 @@ def configure(self, **kwargs):
 
     def install(self, **kwargs):
         super().install(**kwargs)
-        # FIXME: this appears to break the cheribsd build, so let's remove it for now...
```
jrtc27, line 765: This just deletes parts of morello-llvm??

pycheribuild/projects/cross/opensbi.py
```diff
@@ -82,6 +83,8 @@ def setup(self):
         super().setup()
         compflags = " " + self.commandline_to_str(self.essential_compiler_and_linker_flags)
         compflags += " -Qunused-arguments"  # -mstrict-align -no-pie
+        if self.config.riscv_cheri_isa == RiscvCheriISA.STD:
+            compflags += " -mno-relax"
```
jrtc27, line 87: Shouldn't this be always-on now?

jrtc27: (mandated by the compiler driver that is)

pycheribuild/projects/cross/opensbi.py
```diff
@@ -201,3 +217,42 @@ def run_tests(self):
...
+            self.run_make(parallel=False, cwd=self.source_dir, options=args)
+
+
+class BuildAllianceOpenSBIGFE(BuildAllianceOpenSBI):
+    target = "cheri-alliance-opensbi-gfe"
+    repository = ReuseOtherProjectRepository(BuildAllianceOpenSBI, do_update=True)
+
+    def setup(self):
+        super().setup()
+        self.make_args.set(FW_TEXT_START=0xC0000000)
```
jrtc27, line 258: If the code were ported to self-relocate we wouldn't need a separate GFE one, we could just load it to the different address...

pycheribuild/projects/repository.py
```diff
@@ -176,6 +176,7 @@ def __init__(
         force_branch: bool = False,
         temporary_url_override: "Optional[str]" = None,
         url_override_reason: "typing.Any" = None,
+        temporary_branch_override: "Optional[str]" = None,
```
jrtc27, line 179: I'm not convinced temporary_url_override was a great idea rather than a comment, and I'm even less convinced temporary_branch_override is worth introducing versus just changing default_branch and commenting a TODO/XXX/whatever with the why and what it should become

pycheribuild/projects/build_qemu.py
```diff
@@ -570,3 +574,48 @@ def install(self, **kwargs):
...
+        super().setup()
+        if self.statistics:
+            self.COMMON_FLAGS.append("-DDO_CHERI_STATISTICS=1")
+        if self.build_type == BuildType.DEBUG:
+            self.COMMON_FLAGS.append("-DENABLE_CHERI_SANITIY_CHECKS=1")
+        # the capstone disassembler doesn't support CHERI instructions:
+        self.configure_args.append("--disable-capstone")
+        # Linux/BSD-user is not supported for CHERI (yet)
+        self.configure_args.append("--disable-bsd-user")
+        self.configure_args.append("--disable-linux-user")
```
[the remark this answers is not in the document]
arichardson: yeah should probably de-duplicate

arichardson:
> This is intended to target main? I think we need to be very careful if so, that's the point at which we're exposing it to third parties without them having to jump through hoops.

It's hidden behind a flag, but we can make it even more difficult to use by requiring some env var to be set?

arichardson:
> > > This is intended to target main? I think we need to be very careful if so, that's the point at which we're exposing it to third parties without them having to jump through hoops.
> > 
> > 
> > It's hidden behind a flag, but we can make it even more difficult to use by requiring some env var to be set?
> 
> Is that true of all the random codasip repos? And do we do anything to ensure they only build with STD, rather than failing in cryptic ways if you leave the default at V9? I guess ideally CheriBSD would also not allow you to build it for STD without opting in somehow. Once you get to general CHERI C things STD vs V9 doesn't matter, but for those kinds of things it does and I worry we're not capturing that kind of requirement at the moment either.

I'll make sure they are only enabled with the flag

jrtc27:
Also our tool building our forks so we're at least fully responsible for anything that goes wrong, versus our tool building other people's random forks but we'll still be the ones to get all the bug reports.

pycheribuild/projects/cross/opensbi.py
arichardson: Hmm we have this line:

```
            result.append(
                "-mrelax" if _linker_supports_riscv_relaxations(instance.linker, config, xtarget) else "-mno-relax"
            )
```
Should probably just delete that and let the compiler driver decide.

arichardson: ah and that function has
```
        return False
```
so yeah this can be dropped.

pycheribuild/projects/repository.py
arichardson: The url_override does check if the current remote is one of the old_urls and prompts to update, so there is some (minimal) value there. But I agree with the branch override there is no actual checking that the current branch upstream is stale, so probably better off with a comment. Unlike the url_override it only affects cloning, so a comment should be fine.

pycheribuild/qemu_utils.py
```diff
@@ -232,10 +242,13 @@ def qemu_supports_9pfs(qemu: Path, *, config: ConfigBase) -> bool:
...
 
-def riscv_bios_arguments(xtarget: CrossCompileTarget, _, prefer_bbl=True) -> "list[str]":
+def riscv_bios_arguments(
+    xtarget: CrossCompileTarget, cheri_isa: Optional[RiscvCheriISA], prefer_bbl=True
+) -> "list[str]":
     assert xtarget.is_riscv(include_purecap=True)
     if xtarget.is_hybrid_or_purecap_cheri([CPUArchitecture.RISCV64]):
-        # noinspection PyUnreachableCode
+        if cheri_isa == RiscvCheriISA.STD:
+            return ["-bios", "opensbi-riscv64cheristd-virt-fw_jump.bin"]
```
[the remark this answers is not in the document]
arichardson: Yeah this needs to be fixed in QEMU.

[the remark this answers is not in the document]
arichardson: Added a FIXME

pycheribuild/projects/cross/llvm.py
arichardson: Yeah not sure what happened in that commit, well spotted.

pycheribuild/projects/cross/opensbi.py
```diff
@@ -60,7 +61,7 @@ class BuildOpenSBI(Project):
     supported_architectures = (
         CompilationTargets.FREESTANDING_RISCV64_HYBRID,
         CompilationTargets.FREESTANDING_RISCV64,
-        # Won't compile yet: CompilationTargets.FREESTANDING_RISCV64_PURECAP
+        CompilationTargets.FREESTANDING_RISCV64_PURECAP,
```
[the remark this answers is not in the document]
arichardson: reverted

pycheribuild/projects/cross/u_boot.py
```diff
@@ -160,3 +161,31 @@ def run_make(self, *args, **kwargs):
...
+    default_build_type = BuildType.RELWITHDEBINFO
+    supported_architectures = (
+        CompilationTargets.FREESTANDING_RISCV64,
+        CompilationTargets.FREESTANDING_RISCV64_PURECAP,
+    )
+
+    @classmethod
+    def setup_config_options(cls, **kwargs):
+        super().setup_config_options(**kwargs)
+        cls.secure_boot = cls.add_bool_option("secure-boot", default=False, help="Enable secure boot image")
```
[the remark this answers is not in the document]
qwattash: The difference between the secure-boot defconfig and the normal one appears to be minimal, just some stack size change and a couple of env options.

qwattash (approved):

PR #439, opened 2025-10-06
Add experimental support for building and running CHERI-Linux projects (fixed)

arichardson:
Adds some fixes on top of #436, am able to build risc-v purecap kernel and boot it, as well as the morello busybox one.

Think there are still some CI failures but we should be close to ready here.


pycheribuild/projects/cross/linux.py
```diff
@@ -64,31 +65,31 @@ def check_system_dependencies(self) -> None:
...
         elif self.crosscompile_target.is_aarch64(include_purecap=True):
             self.linux_arch = "arm64"
-            self.make_args.set(CROSS_COMPILE="aarch64-unknown-elf-")
 
         self.make_args.set(ARCH=self.linux_arch)
+        self.make_args.set(O=self.build_dir)
 
         # We only support building the kernel with LLVM/Clang
-        self.make_args.set(CC="clang")
-        self.make_args.set(LLVM="1")
```
heshamelmatary: Does that work when you remove it? My understanding is that the Linux build system requires `LLVM=1` when building for Clang. https://docs.kernel.org/kbuild/llvm.html

pycheribuild/projects/cross/muslc.py
```diff
@@ -101,5 +101,5 @@ def setup(self) -> None:
         # FIXME: Morello muslc does not address warnings with implicit functions defined
         # and fails building. This should be fixed in Morello Busybox codebase manually
         # or when they update to recent revisions/releases
-        self.cross_warning_flags.append("-Wno-error=implicit-function-declaration")
+        self.cross_warning_flags.clear()
```
heshamelmatary, line 104: Note: there's now a merged fix for the bad function pointer cast: https://git.morello-project.org/morello/musl-libc/-/merge_requests/154

pycheribuild/projects/cross/linux.py
```diff
@@ -0,0 +1,234 @@
...
+    target = "linux-kernel"
+    repository = GitRepository("https://github.com/torvalds/linux.git")
+    needs_sysroot = False
+    is_sdk_target = False
+    supported_architectures = (
+        CompilationTargets.LINUX_RISCV64,
+        CompilationTargets.LINUX_AARCH64,
+    )
+    _always_add_suffixed_targets = True
+    include_os_in_target_suffix = False  # Avoid adding -picolibc- as we are building picolibc here
```
heshamelmatary, line 57:
```suggestion
    include_os_in_target_suffix = False  # Avoid adding -linux- as we are building linux-kernel here
```

pycheribuild/projects/cross/busybox.py
```diff
@@ -0,0 +1,189 @@
...
+        self.run_make("clean", cwd=self.source_dir)
+
+
+class BuildMorelloBusyBox(BuildBusyBox):
+    target = "morello-busybox"
+    repository = GitRepository("https://git.morello-project.org/morello/morello-busybox.git")
+    supported_architectures = (CompilationTargets.LINUX_MORELLO_PURECAP,)
+
+    @classmethod
+    def dependencies(cls, config) -> "tuple[str, ...]":
```
heshamelmatary, line 118: Need this as well for the base class?

pycheribuild/projects/cross/compiler_rt.py
```diff
@@ -196,42 +206,6 @@ def install(self, **kwargs):
             self.create_symlink(
                 self.install_dir / "lib" / libname, self.install_dir / "lib/libgcc.a", print_verbose_only=False
             )
-        elif self.target_info.is_linux() and not self.compiling_for_host():
-            self.move_file(
-                self.install_dir / "lib/linux" / libname,
-                self.target_info.sysroot_dir / self.target_info.target_triple / "lib" / libname,
-            )
-
-            # Linux Clang driver expects/embeds the following crt files in the linking flags
```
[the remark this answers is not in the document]
arichardson: looks like the end files are not provided by musl and we actually do need compiler-rt. I also avoided the need for gcc_eh by adding --unwindlib=none

PR #440, opened 2025-10-06
Speed up sorting targets

arichardson:
The default python sort is stable, so there is no need to reimplement stable sort in python. I was seeing this simple sort show up in profiling of the total run time with the following flags: `-p  __run_everything__ --clean --build --test --benchmark --source-root=/tmp/fake-src` We now call should_run_before 8828 times instead of the previous ~4.96 million times since the sorting algorithm is much better.


jrtc27:
This was done because we didn't have a total ordering https://github.com/CTSRD-CHERI/cheribuild/commit/b50d66eb93b651dce580877eb30ee1f2a430f2b5

arichardson:
> This was done because we didn't have a total ordering [b50d66e](https://github.com/CTSRD-CHERI/cheribuild/commit/b50d66eb93b651dce580877eb30ee1f2a430f2b5)

Yeah I have a workaround for this now that should be less fragile: just sort in alphabetical order first.

arichardson:
Rewritten to use a topological sort which avoids this n^2 issue and speeds up startup by 5 seconds when running all targets. Order has changed but is now consistenly alphabetical if there are no depedencies.

PR #471, opened 2026-04-08
Morello linux fixes

heshamelmatary:


pycheribuild/config/compilation_targets.py
```diff
@@ -1675,7 +1666,6 @@ class CompilationTargets(BasicCompilationTargets):
         CPUArchitecture.AARCH64,
         MorelloLinuxTargetInfo,
         is_cheri_purecap=True,
-        check_conflict_with=MORELLO_LINUX_AARCH64,
```
arichardson, line 1678: This seems wrong to me, we should not be reusing the same output directory ever.

heshamelmatary:
On Wed, 8 Apr 2026 at 17:38, Alexander Richardson ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In pycheribuild/config/compilation_targets.py
> <https://github.com/CTSRD-CHERI/cheribuild/pull/471#discussion_r3052810599>
> :
>
> > @@ -1675,7 +1666,6 @@ class CompilationTargets(BasicCompilationTargets):
>          CPUArchitecture.AARCH64,
>          MorelloLinuxTargetInfo,
>          is_cheri_purecap=True,
> -        check_conflict_with=MORELLO_LINUX_AARCH64,
>
> This seems wrong to me, we should not be reusing the same output directory
> ever.
>

Any suggestion how to properly fix the build error then?

pycheribuild/config/compilation_targets.py
heshamelmatary: PS this change doesn't mean it will use the same output directory. But the `check_conflict_with` check, for some reason, expects a non_cheri sysroot when building purecap and fails to build as per the error message above. It's worth noting that `CHERI_LINUX_RISCV64_PURECAP_093` doesn't have `check_conflict_with` as well, and adding it triggers the same build error.

arichardson:
Fixed the second issue in #473. Is the first one still needed?

pycheribuild/config/compilation_targets.py
arichardson: resolved in #473

arichardson (changes requested):

heshamelmatary:
@arichardson I removed the second commit, thanks for pushing a fix. The remaining commit isn't needed for any fixes, but is just an enhancement to align with other targets and another conversation on Slack. Also AFAIR, LLVM-19 will fail if there's trailing `_purecap` in the triple at the end, but I only noticed that for RVY

pycheribuild/projects/cross/busybox.py
```diff
@@ -179,6 +179,16 @@ def install(self, **kwargs) -> None:
         if libc_so.exists():
             self.install_file(libc_so, self.install_dir / f"rootfs/lib/ld-musl-{self.triple_arch}.so.1")
 
+            # XXX Morello Linux's driver expects _purecap suffix for the musl's rtld/interpreter, while
+            # RVY does not. We don't support hybrid/legacy userspace in this cheribuild distro, it is either
+            # all purecap or non-CHERI. In this case, removing _purecap from the triple is fine.
+            # Creating a symlink is workaround for Morello LLVM's morello-linux driver.
+            if self.target_info.target.is_cheri_purecap():
```
arichardson: I think this should only be done for Morello?

heshamelmatary: @arichardson Yeah, I thought to just do it in `BuildMorelloBusyBox` but it will need to override most/all of `install`. Also will be duplicated in `BuildAllianceBusyBox` as I am currently working on consolidating Morello builds into CHERI-Linux targets including busybox. I could move it to `BuildMorelloBusyBox` for now, or probably leave it here in the base class but add an `and` eg

`self.target_info.target.is_cheri_purecap() and self.crosscompile_target.is_aarch64(include_purecap=True):
`
What do you suggest?

heshamelmatary: @arichardson This is now only done for Morello with this line in the same place `if self.target_info.target.is_cheri_purecap([CPUArchitecture.AARCH64])`. Please re-open this if that doesn't address your comment properly

paul-metzger:
This fixes an issue I encountered with dynamically linked tests.

arichardson (approved):
~Can you rebase this to remove the merge commit?~ EDIT: I was able to with the web UI.

PR #472, opened 2026-04-09
CHERI-Linux: update to Codasip's 6.18 release

heshamelmatary:


heshamelmatary:
@paulmetzger Likely won't affect your userspace, but would be good to know if you get time to test this update with the updated kernel. It gets me an interactive shell.

arichardson (approved):

paul-metzger:
@heshamelmatary The commit looks good to me. The same test cases pass and fail than on the 6.16 kernel. Two test cases that are skipped on 6.16, due to the issue with zcherilevels being missing in /proc/cpu, are executed with this kernel and pass, as this issue is fixed in 6.18.

We should start to think about using the test suite in a CI setup.

PR #475, opened 2026-04-10
test-scripts: fix race condition in parallel libc++ tests

arichardson:
The main process could sometimes check the barrier's waiting count
before all shards had reached the barrier, even after receiving
the NEXT_STAGE message. This was especially prevalent in single-CPU
or high-load environments.

Fix this by:
1. Adding a loop to wait for mp_barrier.n_waiting == len(processes).
2. Removing the short 1s timeout on mp_barrier.wait(), as the
   barrier already has a 4-hour timeout and a short one is prone
   to spurious failures under load.
3. Adding a check to ensure shards are still alive while waiting.

### DIFF
--- a/test-scripts/run_libcxx_tests.py
+++ b/test-scripts/run_libcxx_tests.py
@@ -400,12 +400,19 @@ def run_parallel_impl(
                         loop_start_time - start_time,
                     )
                 if len(processes_in_next_stage) == len(processes):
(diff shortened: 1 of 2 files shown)


arichardson:
@bsdjhb @jrtc27 This should fix the spurious CI failures we have been seeing. Commits created using gemini-cli but result looks sensible.

PR #477, opened 2026-04-13
musl: allow building CA's muslc for Morello

heshamelmatary:
This commit enables building CHERI Alliance's muslc for Morello (once https://github.com/CHERI-Alliance/musl/pull/4 is merged)

### DIFF
--- a/pycheribuild/projects/cross/muslc.py
+++ b/pycheribuild/projects/cross/muslc.py
@@ -102,10 +102,10 @@ class BuildAllianceLinuxMuslc(BuildMuslc):
 
     def setup(self) -> None:
         if self.crosscompile_target.is_riscv(include_purecap=True):
-            self.configure_args.extend(["--enable-bakewell --enable-debug"])
+            self.configure_args.append("--enable-bakewell")
         elif self.crosscompile_target.is_aarch64(include_purecap=True):
-            self.configure_args.extend(["--enable-morello"])
-
+            self.configure_args.append("--enable-morello")
+        self.configure_args.append("--enable-debug")
         # FIXME Need to add the compiler resource directory as Codasip's muslc includes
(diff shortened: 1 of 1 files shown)


pycheribuild/projects/cross/muslc.py
```diff
@@ -101,7 +101,11 @@ class BuildAllianceLinuxMuslc(BuildMuslc):
     dependencies = ("cheri-std093-compiler-rt-builtins", "cheri-std093-linux-kernel")
 
     def setup(self) -> None:
-        self.configure_args.extend(["--enable-bakewell --enable-debug"])
+        if self.crosscompile_target.is_riscv(include_purecap=True):
+            self.configure_args.extend(["--enable-bakewell --enable-debug"])
+        elif self.crosscompile_target.is_aarch64(include_purecap=True):
+            self.configure_args.extend(["--enable-morello"])
+
```
arichardson:
```suggestion
        if self.crosscompile_target.is_riscv(include_purecap=True):
            self.configure_args.append("--enable-bakewell")
        elif self.crosscompile_target.is_aarch64(include_purecap=True):
            self.configure_args.append("--enable-morello")
        self.configure_args.append("--enable-debug")
```

Surprised this actually works since we are appending `--enable-bakewell --enable-debug` as a single argument. Maybe `--enable-debug` is just being ignored as a optional suffix?

arichardson (approved):
LGTM with minor changes. Not sure about the --enable-debug. Was it there before or was it being ignored?

pycheribuild/projects/cross/muslc.py
heshamelmatary: Probably. I haven't checked what --enable-debug actually does, I copied it over from their default musl README/flags

arichardson (approved):

PR #478, opened 2026-04-13
CHERI-Linux: Refactor target names

heshamelmatary:
Moving towards consolidating RVY + Morello Linux repos/targets and using CHERI Alliance repos only, this PR refactors CHERI-Linux target names with the following assumptions:

* CHERI-Linux (the kernel and userspace) now means CHERI Alliance Linux that is able to build RVY+Morello off the same repos. Morello Linux (gitlab) still exists for now, but is deemed to be abandoned once we sort out all build+runtime issues.  
* Use default project names for CHERI as the default target names, similar to other projects. Building with cheribuild likely implies building CHERI- supported variants/repos. This removes "cheri-std093" prefixes from current CHERI Alliance Linux targets, as they are now misleading given they are also used for Morello.
* Use CHERI Alliance's repos as the default for CHERI-Linux targets.
* Upstream repos/targets are special and not default; therefore, prefix those with "upstream-", similar to other projects (e.g. upstream-llvm, upstream-qemu). This also removes existing "cheri-" prefixes, assuming that cheribuild builds CHERI-enabled projects by default.
* Resulting commands are now simplified, e.g.:
```
  * ./cheribuild.py run-minimal-cheri-linux-riscv64-purecap -d
  * ./cheribuild.py run-minimal-cheri-linux-morello-purecap -d
```

And for building just the kernel:
```
  * ./cheribuild.py linux-kernel-riscv64-purecap -d
  * ./cheribuild.py linux-kernel-morello-purecap -d
```

PS: RVY CHERI-Linux boots fine, but Morello CHERI-Linux only builds completely but userspace currently has run-time issues. 
Preferably don't merge this until https://github.com/CTSRD-CHERI/cheribuild/pull/477 and https://github.com/CTSRD-CHERI/cheribuild/pull/471 are merged.

### DIFF
--- a/pycheribuild/config/compilation_targets.py
+++ b/pycheribuild/config/compilation_targets.py
@@ -897,8 +897,8 @@ class LinuxTargetInfoBase(_ClangBasedTargetInfo, ABC):
 
 class UpstreamLinuxTargetInfo(LinuxTargetInfoBase):
     uses_upstream_llvm: bool = True
-    kernel_target = "linux-kernel"
-    musl_target = "muslc"
+    kernel_target = "upstream-linux-kernel"
+    musl_target = "upstream-muslc"
     compiler_rt_target = "upstream-compiler-rt-builtins"
 
     @property
@@ -907,8 +907,8 @@ class UpstreamLinuxTargetInfo(LinuxTargetInfoBase):
 
 
 class CheriLinuxTargetInfo(LinuxTargetInfoBase):
-    kernel_target = "cheri-std093-linux-kernel"
-    musl_target = "cheri-std093-muslc"
+    kernel_target = "linux-kernel"
+    musl_target = "muslc"
     compiler_rt_target = "cheri-std093-compiler-rt-builtins"
 
     @property
@@ -919,6 +919,7 @@ class CheriLinuxTargetInfo(LinuxTargetInfoBase):
 
 class CheriLinuxWithMorelloCompilerTargetInfo(CheriLinuxTargetInfo):
     uses_morello_llvm = True
+    compiler_rt_target = "morello-compiler-rt-builtins-for-cheri-alliance-linux"
 
 
 class MorelloLinuxTargetInfo(LinuxTargetInfoBase):
--- a/pycheribuild/projects/cross/compiler_rt.py
+++ b/pycheribuild/projects/cross/compiler_rt.py
@@ -255,15 +255,6 @@ class BuildUpstreamCompilerRtBuiltins(BuildCompilerRtBuiltins):
         *CompilationTargets.ALL_UPSTREAM_LINUX_TARGETS,
     )
 
-    @classmethod
-    def dependencies(cls, config) -> "tuple[str, ...]":
-        result = super().dependencies(config)
-        xtarget = cls.get_crosscompile_target()
-        if xtarget.target_info_cls.is_linux() and not xtarget.is_native():
-            # The builtins for Linux need C library and kernel headers available.
-            result += ("muslc-headers",)
-        return result
-
 
 class BuildAllianceCompilerRtBuiltins(BuildCompilerRtBuiltins):
     target = "cheri-std093-compiler-rt-builtins"
@@ -289,11 +280,7 @@ class BuildMorelloCompilerRtBuiltins(BuildCompilerRtBuiltins):
         *CompilationTargets.ALL_MORELLO_LINUX_TARGETS,
     )
 
-    @classmethod
-    def dependencies(cls, config) -> "tuple[str, ...]":
-        result = super().dependencies(config)
-        xtarget = cls.get_crosscompile_target()
-        if xtarget.target_info_cls.is_linux() and not xtarget.is_native():
-            # The builtins for Linux need C library and kernel headers available.
-            result += ("morello-muslc-headers",)
-        return result
+
+class BuildCheriLinuxCompilerRtBuiltins(BuildMorelloCompilerRtBuiltins):
+    target = "morello-compiler-rt-builtins-for-cheri-alliance-linux"
+    _supported_architectures = (CompilationTargets.CHERI_LINUX_MORELLO_PURECAP,)
--- a/pycheribuild/projects/cross/busybox.py
+++ b/pycheribuild/projects/cross/busybox.py
@@ -42,7 +42,7 @@ from ...utils import classproperty
 
 
 class BuildBusyBox(CrossCompileAutotoolsProject):
-    target = "busybox"
+    target = "upstream-busybox"
     repository = GitRepository("https://git.busybox.net/busybox/")
     is_sdk_target = False
     _supported_architectures = CompilationTargets.ALL_UPSTREAM_LINUX_TARGETS
@@ -229,7 +229,7 @@ class BuildMorelloBusyBox(BuildBusyBox):
 
 
 class BuildAllianceBusyBox(BuildBusyBox):
-    target = "cheri-std093-busybox"
+    target = "busybox"
     repository = GitRepository("https://github.com/CHERI-Alliance/busybox.git")
     _supported_architectures = CompilationTargets.ALL_CHERI_LINUX_TARGETS
     _default_architecture = CompilationTargets.CHERI_LINUX_MORELLO_PURECAP
--- a/pycheribuild/projects/cross/linux.py
+++ b/pycheribuild/projects/cross/linux.py
@@ -46,7 +46,7 @@ from ...utils import classproperty
 
 
 class BuildLinux(CrossCompileAutotoolsProject):
-    target = "linux-kernel"
+    target = "upstream-linux-kernel"
     repository = GitRepository("https://github.com/torvalds/linux.git")
     _needs_sysroot = False
     is_sdk_target = False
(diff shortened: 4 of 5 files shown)


arichardson:
`--riscv-cheri-isa=experimental_std093` should not be needed for the linux targets since that is already the default for them, can you confirm that it works without?

pycheribuild/config/compilation_targets.py
```diff
@@ -907,9 +907,9 @@ def sysroot_dir(self) -> Path:
 
 
 class CheriLinuxTargetInfo(LinuxTargetInfoBase):
-    kernel_target = "cheri-std093-linux-kernel"
-    musl_target = "cheri-std093-muslc"
-    compiler_rt_target = "cheri-std093-compiler-rt-builtins"
+    kernel_target = "linux-kernel"
+    musl_target = "muslc"
+    compiler_rt_target = "alliance-compiler-rt-builtins"
```
arichardson: I think for LLVM (and therefore compiler-rt) @jrtc27 was suggesting we should be explicit that it's 0.9.3.

arichardson (commented):
This looks fine to me, but I'd like @jrtc27 to approve.

heshamelmatary:
On Mon, 13 Apr 2026 at 17:41, Alexander Richardson ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In pycheribuild/config/compilation_targets.py
> <https://github.com/CTSRD-CHERI/cheribuild/pull/478#discussion_r3074458176>
> :
>
> > @@ -907,9 +907,9 @@ def sysroot_dir(self) -> Path:
>
>
>  class CheriLinuxTargetInfo(LinuxTargetInfoBase):
> -    kernel_target = "cheri-std093-linux-kernel"
> -    musl_target = "cheri-std093-muslc"
> -    compiler_rt_target = "cheri-std093-compiler-rt-builtins"
> +    kernel_target = "linux-kernel"
> +    musl_target = "muslc"
> +    compiler_rt_target = "alliance-compiler-rt-builtins"
>
> I think for LLVM (and therefore compiler-rt) @jrtc27
> <https://github.com/jrtc27> was suggesting we should be explicit that
> it's 0.9.3.
>
> Yeah cheri-std093-llvm target name should be the same, nothing changed
with that. I just tested building this alliance’s compiler-rt for Morello,
with this repo and it seems to build fine. Though I’m yet to debug
userspace runtime issues for Morello which could expose some runtime bugs
in this alliance compiler-rt, I haven’t yet looked at the CRT differences.
I thought cheri-std093-compiler-rt-builtins-morello-purecap is a bit
odd/misleading, that’s why I sed/cheri-std093/alliance. Happy to change
that anyway.

heshamelmatary:
I am now considering just reverting the alliance-compiler-rt change for morello to avoid cloning the alliance's llvm while building CHERI-Linux for Morello

paul-metzger:
LGTM

heshamelmatary:
@arichardson The latest push Introduces a new CHERI-Linux compiler-rt target to enable building Morello for it (using morello-llvm), and refactors CHERI-Linux dependencies accordingly. Could you please review that again?

I also tested RVY without the `--riscv-cheri-isa=experimental_std093` and it works. If that sounds good to you and @jrtc27, I could also s/cheri-std093-sdk/cheri-alliance-sdk.

pycheribuild/config/compilation_targets.py
```diff
@@ -919,6 +919,7 @@ def sysroot_dir(self) -> Path:
 
 class CheriLinuxWithMorelloCompilerTargetInfo(CheriLinuxTargetInfo):
     uses_morello_llvm = True
+    compiler_rt_target = "cherilinux-compiler-rt-builtins"
```
arichardson: Why can't we use the `morello-compiler-rt-builtins` here? Why the new target?

heshamelmatary: `morello-compiler-rt-builtins` would duplicate two targets names if we add `CHERI_LINUX_MORELLO_PURECAP` to its `_supported_architectures`. The reason is that both `CHERI_LINUX_MORELLO_PURECAP` and `MORELLO_LINUX_MORELLO_PURECAP` define the same `morello-purecap` target name. That's mainly why I added this new target.

arichardson: Ah thanks, that makes sense. Since the cheri-alliance compiler-rt builtins are versioned to work with the cheri-linux target, could we just use the `cheri-std093-compiler-rt-builtins` instead?

arichardson: Or are there morello-specific changes that are needed? If so duplicating the target seems like the best approach

heshamelmatary: Using `cheri-std093-compiler-rt-builtins` for Morello will force checking out CHERI Alliance LLVM (besides Morello LLVM) even if someone just wants to build for Morello, which is why I reverted this change in the first place :(

Having duplicate target names won't build and will error, but I am not sure if that's what you mean? ie we can't have duplicate target names, right? It was complaining for me that `morello-compiler-rt-builtins-morello-purecap` is duplicated (for Morello Linux and for CHERI-Linux), when I tried to add `CHERI_LINUX_MORELLO_PURECAP` to `morello-compiler-rt-builtins`'s `_supported_architectures`

pycheribuild/projects/cross/compiler_rt.py
```diff
@@ -306,3 +306,22 @@ def dependencies(cls, config) -> "tuple[str, ...]":
...
+    llvm_project = BuildMorelloLLVM
+    repository = ReuseOtherProjectDefaultTargetRepository(llvm_project, subdirectory="compiler-rt")
+    _default_architecture = CompilationTargets.CHERI_LINUX_MORELLO_PURECAP
+    _supported_architectures = (
+        *BuildCompilerRtBuiltins._supported_architectures,
+        *CompilationTargets.ALL_CHERI_LINUX_TARGETS,
+    )
+
+    @classmethod
+    def dependencies(cls, config) -> "tuple[str, ...]":
```
arichardson: This should not be needed, the parent class already does it:

```
    @classmethod
    def dependencies(cls, config) -> "tuple[str, ...]":
        result = super().dependencies(config)
        xtarget = cls.get_crosscompile_target()
        if xtarget.target_info_cls.is_linux() and not xtarget.is_native():
            # The builtins for Linux need C library and kernel headers available.
            ti = typing.cast(typing.Type[LinuxTargetInfoBase], xtarget.target_info_cls)
            result += (ti.kernel_target, ti.musl_target + "-headers")
        return result
```

arichardson:
https://github.com/CTSRD-CHERI/cheribuild/pull/480 should make this a bit simpler.

pycheribuild/config/compilation_targets.py
arichardson: Yeah, let's keep the subclass for now. I pushed https://github.com/CTSRD-CHERI/cheribuild/pull/480 to make the remaining diff simpler.

pycheribuild/projects/cross/compiler_rt.py
```diff
@@ -306,3 +306,22 @@ def dependencies(cls, config) -> "tuple[str, ...]":
...
+
+class BuildCheriLinuxCompilerRtBuiltins(BuildCompilerRtBuiltins):
+    target = "cherilinux-compiler-rt-builtins"
+    llvm_project = BuildMorelloLLVM
+    repository = ReuseOtherProjectDefaultTargetRepository(llvm_project, subdirectory="compiler-rt")
+    _default_architecture = CompilationTargets.CHERI_LINUX_MORELLO_PURECAP
+    _supported_architectures = (
+        *BuildCompilerRtBuiltins._supported_architectures,
+        *CompilationTargets.ALL_CHERI_LINUX_TARGETS,
+    )
```
arichardson:
```suggestion
class BuildCheriLinuxCompilerRtBuiltins(BuildMorelloCompilerRtBuiltins):
    target = "morello-compiler-rt-builtins-for-cheri-alliance-linux"
    _supported_architectures = (CompilationTargets.CHERI_LINUX_MORELLO_PURECAP,)
```

I think this should be sufficient with #480 landed.

pycheribuild/config/compilation_targets.py
arichardson:
```suggestion
    compiler_rt_target = "morello-compiler-rt-builtins-for-cheri-alliance-linux"
```

heshamelmatary:
> #480 should make this a bit simpler.

Indeed! Thanks for pushing that. I rebased this PR now to work with main, much smaller and simpler now

pycheribuild/projects/cross/compiler_rt.py
```diff
@@ -261,7 +261,7 @@ def dependencies(cls, config) -> "tuple[str, ...]":
         xtarget = cls.get_crosscompile_target()
         if xtarget.target_info_cls.is_linux() and not xtarget.is_native():
             # The builtins for Linux need C library and kernel headers available.
-            result += ("muslc-headers",)
+            result += ("upstream-muslc-headers",)
```
arichardson: I think we can actually just delete this override, the base class handles it.

arichardson (approved):
Minor suggestion otherwise LGTM

arichardson (approved):

PR #479, opened 2026-04-14
Targets for the portable stand-alone variant of the CheriBSD test suite

paul-metzger:
This PR contains CHERI/Morello Linux targets for a portable stand-alone variant of the CheriBSD test suite. We'd like to be able to use it on CHERI/Morello Linux for testing and to help with cataloguing design differences between the CHERI-related parts of CheriBSD's and CHERI/Morello Linux's OS interfaces.
@heshamelmatary is interested in this PR for testing.

Setup for Morello Linux:
1. Build the test suite: `./cheribuild.py cheri-portable-os-tests --qemu/no-use-smbd -d`
2. Start Morello Linux in Qemu: `./cheribuild.py run-minimal-morello-linux-morello-purecap --qemu/no-use-smbd -d`
3. The test cases are then in `/opt/cheri-portable-os-tests`.
This issue causes a compile time error and needs to be fixed by hand for Morello Muslc: https://github.com/CHERI-Alliance/musl/pull/3
The test suite needs to be built first in order for it to be included into the image built in the second step.
The workflow for RVY CHERI Linux is similar.

The repository with the adapted test cases is private currently, but the intention is to make it publicly available soon.

### DIFF
--- a/pycheribuild/config/compilation_targets.py
+++ b/pycheribuild/config/compilation_targets.py
@@ -1704,10 +1704,10 @@ class CompilationTargets(BasicCompilationTargets):
         _cheri_isa=RiscvCheriISA.EXPERIMENTAL_STD093,
     )
     UPSTREAM_LINUX_AARCH64 = CrossCompileTarget("aarch64", CPUArchitecture.AARCH64, UpstreamLinuxTargetInfo)
-    CHERI_LINUX_AARCH64 = CrossCompileTarget("aarch64", CPUArchitecture.AARCH64, CheriLinuxTargetInfo)
+    CHERI_LINUX_AARCH64 = CrossCompileTarget("wip-aarch64", CPUArchitecture.AARCH64, CheriLinuxTargetInfo)
     MORELLO_LINUX_AARCH64 = CrossCompileTarget("aarch64", CPUArchitecture.AARCH64, MorelloLinuxTargetInfo)
     CHERI_LINUX_MORELLO_PURECAP = CrossCompileTarget(
-        "morello-purecap",
+        "wip-morello-purecap",
         CPUArchitecture.AARCH64,
         CheriLinuxWithMorelloCompilerTargetInfo,
         is_cheri_purecap=True,
--- a/pycheribuild/config/compilation_targets.py
+++ b/pycheribuild/config/compilation_targets.py
@@ -1704,18 +1704,22 @@ class CompilationTargets(BasicCompilationTargets):
         _cheri_isa=RiscvCheriISA.EXPERIMENTAL_STD093,
     )
     UPSTREAM_LINUX_AARCH64 = CrossCompileTarget("aarch64", CPUArchitecture.AARCH64, UpstreamLinuxTargetInfo)
-    CHERI_LINUX_AARCH64 = CrossCompileTarget("wip-aarch64", CPUArchitecture.AARCH64, CheriLinuxTargetInfo)
-    MORELLO_LINUX_AARCH64 = CrossCompileTarget("aarch64", CPUArchitecture.AARCH64, MorelloLinuxTargetInfo)
+    CHERI_LINUX_AARCH64 = CrossCompileTarget("aarch64", CPUArchitecture.AARCH64, CheriLinuxTargetInfo)
+    MORELLO_LINUX_AARCH64 = CrossCompileTarget("legacy-aarch64", CPUArchitecture.AARCH64, MorelloLinuxTargetInfo)
     CHERI_LINUX_MORELLO_PURECAP = CrossCompileTarget(
-        "wip-morello-purecap",
+        "morello-purecap",
         CPUArchitecture.AARCH64,
         CheriLinuxWithMorelloCompilerTargetInfo,
         is_cheri_purecap=True,
         check_conflict_with=CHERI_LINUX_AARCH64,
         non_cheri_target=CHERI_LINUX_AARCH64,
     )
+    # This target is kept for historical reasons as Morello Linux (from Arm) was the first Linux kernel
+    # and target that added CHERI support. CHERI-Linux (from CHERI Alliance) has built on top of Morello
+    # Linux to include more features, targets (RISC-V), and is the currently active project and should
+    # be used instead of the following target unless there is a good reason not to.
     MORELLO_LINUX_MORELLO_PURECAP = CrossCompileTarget(
-        "morello-purecap",
+        "legacy-morello-purecap",
         CPUArchitecture.AARCH64,
         MorelloLinuxTargetInfo,
         is_cheri_purecap=True,
--- a/pycheribuild/config/compilation_targets.py
+++ b/pycheribuild/config/compilation_targets.py
@@ -191,6 +191,10 @@ class _ClangBasedTargetInfo(TargetInfo, ABC):
         return self._compiler_dir / "ld.lld"
 
     @property
+    def objcopy(self) -> Path:
+        return self._compiler_dir / "objcopy"
+
+    @property
     def ar(self) -> Path:
         return self._compiler_dir / "llvm-ar"
 
--- a/pycheribuild/projects/cross/libbsd.py
+++ b/pycheribuild/projects/cross/libbsd.py
@@ -40,10 +40,12 @@ class BuildLibbsd(CrossCompileAutotoolsProject):
     )
     _default_architecture = CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093
     make_kind = MakeCommandKind.GnuMake
-    # `default_branch` is set because the build scripts assume that
-    # a git tag is checked out. In more detail ./get-version returns an
-    # empty string if main is checked out.
-    repository = GitRepository("https://gitlab.freedesktop.org/libbsd/libbsd.git", default_branch="0.12.2")
+    repository = GitRepository(
+        "https://gitlab.freedesktop.org/libbsd/libbsd.git",
+        temporary_url_override="https://gitlab.freedesktop.org/paul-metzger/libbsd-private-fork-pmetzger.git",
+        default_branch="0_12_2_with_alignment_macro_fix",
+        force_branch=True,
+    )
 
     @classproperty
     def default_install_dir(self):
@@ -53,28 +55,6 @@ class BuildLibbsd(CrossCompileAutotoolsProject):
     def dependencies(cls, config) -> "tuple[str, ...]":
         ti = typing.cast(typing.Type[LinuxTargetInfoBase], cls.get_crosscompile_target().target_info_cls)
         return ti.musl_target, "libmd"
-
-    def _apply_patch(self) -> None:
-        patch = """
-diff --git a/src/merge.c b/src/merge.c
-index 3f1b3fb..f8cb602 100644
---- a/src/merge.c
-+++ b/src/merge.c
-@@ -84,8 +84,8 @@ static void insertionsort(unsigned char *, size_t, size_t,
-  */
- /* Assumption: PSIZE is a power of 2. */
- #define EVAL(p) (unsigned char **)					\\
--	 (((unsigned char *)p + PSIZE - 1 -				\\
--	   (unsigned char *)0) & ~(PSIZE - 1))
-+	 __builtin_cheri_address_set(p, ((__builtin_cheri_address_get(p) + PSIZE - 1 -				\\
-+	   0) & ~(PSIZE - 1)))
-
- /*
-  * Arguments are as for qsort.
-"""
-        self.write_file(self.source_dir / "merge.patch", patch, overwrite=True)
-        self.run_cmd("git", "restore", ".", cwd=self.source_dir)
-        self.run_cmd("git", "apply", "merge.patch", cwd=self.source_dir)
 
     def setup(self) -> None:
         super().setup()
@@ -89,7 +69,6 @@ index 3f1b3fb..f8cb602 100644
         super().configure(**kwargs)
 
     def compile(self, **kwargs):
-        self._apply_patch()
         self.run_make()
         super().compile()
 
--- a/pycheribuild/projects/cross/libbsd.py
+++ b/pycheribuild/projects/cross/libbsd.py
@@ -1,0 +1,98 @@
+#
+# Copyright (c) 2025-2026 Paul Metzger
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+# OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+# OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+# SUCH DAMAGE.
+#
+
+import typing
+
+from .crosscompileproject import CrossCompileAutotoolsProject, DefaultInstallDir, GitRepository, MakeCommandKind
+from ...config.compilation_targets import CompilationTargets, LinuxTargetInfoBase
+from ...utils import classproperty
+
+
+class BuildLibbsd(CrossCompileAutotoolsProject):
+    _always_add_suffixed_targets = True
+    _can_use_autogen_sh = True
+    _supported_architectures = (
+        CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093,
+        CompilationTargets.MORELLO_LINUX_MORELLO_PURECAP,
+    )
+    _default_architecture = CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093
+    make_kind = MakeCommandKind.GnuMake
+    # `default_branch` is set because the build scripts assume that
+    # a git tag is checked out. In more detail ./get-version returns an
+    # empty string if main is checked out.
+    repository = GitRepository("https://gitlab.freedesktop.org/libbsd/libbsd.git", default_branch="0.12.2")
+
+    @classproperty
+    def default_install_dir(self):
+        return DefaultInstallDir.ROOTFS_LOCALBASE
+
+    @classmethod
+    def dependencies(cls, config) -> "tuple[str, ...]":
+        ti = typing.cast(typing.Type[LinuxTargetInfoBase], cls.get_crosscompile_target().target_info_cls)
+        return ti.musl_target, "libmd"
+
+    def _apply_patch(self) -> None:
+        patch = """
+diff --git a/src/merge.c b/src/merge.c
+index 3f1b3fb..f8cb602 100644
+--- a/src/merge.c
++++ b/src/merge.c
+@@ -84,8 +84,8 @@ static void insertionsort(unsigned char *, size_t, size_t,
+  */
+ /* Assumption: PSIZE is a power of 2. */
+ #define EVAL(p) (unsigned char **)					\\
+-	 (((unsigned char *)p + PSIZE - 1 -				\\
+-	   (unsigned char *)0) & ~(PSIZE - 1))
++	 __builtin_cheri_address_set(p, ((__builtin_cheri_address_get(p) + PSIZE - 1 -				\\
++	   0) & ~(PSIZE - 1)))
+
+ /*
+  * Arguments are as for qsort.
+"""
+        self.write_file(self.source_dir / "merge.patch", patch, overwrite=True)
+        self.run_cmd("git", "restore", ".", cwd=self.source_dir)
+        self.run_cmd("git", "apply", "merge.patch", cwd=self.source_dir)
+
+    def setup(self) -> None:
+        super().setup()
+        # Remove dependency on libgcc_eh
+        self.COMMON_LDFLAGS.append("--unwindlib=none")
+        # Remove dependcy on libgcc_s
+        self.COMMON_LDFLAGS.append("-Wc,--unwindlib=none")
+
+    def configure(self, **kwargs):
+        if not self.configure_command.exists():
+            self.run_cmd(self.source_dir / "autogen", cwd=self.source_dir)
+        super().configure(**kwargs)
+
+    def compile(self, **kwargs):
+        self._apply_patch()
+        self.run_make()
+        super().compile()
+
+    def install(self, **kwargs):
+        self.run_make_install()
+        super().install(**kwargs)
--- a/pycheribuild/projects/cross/libbsd.py
+++ b/pycheribuild/projects/cross/libbsd.py
@@ -36,6 +36,7 @@ class BuildLibbsd(CrossCompileAutotoolsProject):
     _can_use_autogen_sh = True
     _supported_architectures = (
         CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093,
+        CompilationTargets.CHERI_LINUX_MORELLO_PURECAP,
         CompilationTargets.MORELLO_LINUX_MORELLO_PURECAP,
     )
     _default_architecture = CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093
--- a/pycheribuild/projects/cross/libmd.py
+++ b/pycheribuild/projects/cross/libmd.py
@@ -36,6 +36,7 @@ class BuildLibmd(CrossCompileAutotoolsProject):
     _can_use_autogen_sh = True
     _supported_architectures = (
         CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093,
+        CompilationTargets.CHERI_LINUX_MORELLO_PURECAP,
         CompilationTargets.MORELLO_LINUX_MORELLO_PURECAP,
     )
     _default_architecture = CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093
--- a/pycheribuild/projects/cross/libmd.py
+++ b/pycheribuild/projects/cross/libmd.py
@@ -1,0 +1,73 @@
+#
+# Copyright (c) 2025-2026 Paul Metzger
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+# OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+# OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+# SUCH DAMAGE.
+#
+
+import typing
+
+from .crosscompileproject import CrossCompileAutotoolsProject, DefaultInstallDir, GitRepository, MakeCommandKind
+from ...config.compilation_targets import CompilationTargets, LinuxTargetInfoBase
+from ...utils import classproperty
+
+
+class BuildLibmd(CrossCompileAutotoolsProject):
+    _always_add_suffixed_targets = True
+    _can_use_autogen_sh = True
+    _supported_architectures = (
+        CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093,
+        CompilationTargets.MORELLO_LINUX_MORELLO_PURECAP,
+    )
+    _default_architecture = CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093
+    make_kind = MakeCommandKind.GnuMake
+    is_sdk_target = False
+    repository = GitRepository("https://git.hadrons.org/git/libmd.git")
+
+    @classproperty
+    def default_install_dir(self):
+        return DefaultInstallDir.ROOTFS_LOCALBASE
+
+    @classmethod
+    def dependencies(cls, config) -> "tuple[str, ...]":
+        ti = typing.cast(typing.Type[LinuxTargetInfoBase], cls.get_crosscompile_target().target_info_cls)
+        return (ti.musl_target,)
+
+    def setup(self) -> None:
+        super().setup()
+        # Remove dependency on libgcc_eh
+        self.COMMON_LDFLAGS.append("--unwindlib=none")
+        # Remove dependcy on libgcc_s
+        self.COMMON_LDFLAGS.append("-Wc,--unwindlib=none")
+
+    def configure(self, **kwargs):
+        if not self.configure_command.exists():
+            self.run_cmd(self.source_dir / "autogen", cwd=self.source_dir)
+        super().configure(**kwargs)
+
+    def compile(self, **kwargs):
+        self.run_make()
+        super().compile(**kwargs)
+
+    def install(self, **kwargs):
+        self.run_make_install()
+        super().install(**kwargs)
--- a/pycheribuild/projects/cross/libxo.py
+++ b/pycheribuild/projects/cross/libxo.py
@@ -36,6 +36,7 @@ class BuildLibxo(CrossCompileAutotoolsProject):
     _can_use_autogen_sh = True
     _supported_architectures = (
         CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093,
+        CompilationTargets.CHERI_LINUX_MORELLO_PURECAP,
         CompilationTargets.MORELLO_LINUX_MORELLO_PURECAP,
     )
     _default_architecture = CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093
--- a/pycheribuild/projects/cross/libxo.py
+++ b/pycheribuild/projects/cross/libxo.py
@@ -1,0 +1,87 @@
+#
+# Copyright (c) 2025-2026 Paul Metzger
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+# OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+# OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+# SUCH DAMAGE.
+#
+
+import typing
+
+from .crosscompileproject import CrossCompileAutotoolsProject, DefaultInstallDir, GitRepository, MakeCommandKind
+from ...config.compilation_targets import CompilationTargets, LinuxTargetInfoBase
+from ...utils import classproperty
+
+
+class BuildLibxo(CrossCompileAutotoolsProject):
+    _always_add_suffixed_targets = True
+    _can_use_autogen_sh = True
+    _supported_architectures = (
+        CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093,
+        CompilationTargets.MORELLO_LINUX_MORELLO_PURECAP,
+    )
+    _default_architecture = CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093
+    make_kind = MakeCommandKind.GnuMake
+    repository = GitRepository("https://github.com/Juniper/libxo.git")
+
+    @classproperty
+    def default_install_dir(self):
+        return DefaultInstallDir.ROOTFS_LOCALBASE
+
+    @classmethod
+    def dependencies(cls, config) -> "tuple[str, ...]":
+        ti = typing.cast(typing.Type[LinuxTargetInfoBase], cls.get_crosscompile_target().target_info_cls)
+        return ti.musl_target, "libbsd", "libmd"
+
+    def _patch_to_use_libbsd(self, path):
+        self.run_cmd("sed", "-i", "s|<sys/queue.h>|<bsd/sys/queue.h>|g", path, cwd=self.source_dir)
+
+    def _patch_includes(self) -> None:
+        self._patch_to_use_libbsd("libxo/xo_encoder.c")
+        self._patch_to_use_libbsd("xopo/xopo.c")
+
+    def _patch_configure_ac(self) -> None:
+        self.run_cmd("sed", "-i", "s|AC_FUNC_REALLOC|#AC_FUNC_REALLOC|g", "configure.ac", cwd=self.source_dir)
+        self.run_cmd("sed", "-i", "s|AC_FUNC_MALLOC|#AC_FUNC_MALLOC|g", "configure.ac", cwd=self.source_dir)
+
+    def configure(self, **kwargs):
+        # Don't depend on libgcc_s. If this isn't set then clang wants to link
+        # with libgcc_s, which is not available on Morello Linux.
+        self.LDFLAGS.append("--unwindlib=none")
+        # This prompts libtool to pass '--unwindlib=none' to clang during
+        # linking. Libtool ignores "--unwindlib=none" and needs
+        # "-Wc,--unwindlib=none" instead. "-Wc,--unwindlib=none" is turned
+        # into "--unwindlib=none" by libtool when it invokes clang.
+        self.CFLAGS.append("-Wc,--unwindlib=none")
+
+        self._patch_configure_ac()
+        self._patch_includes()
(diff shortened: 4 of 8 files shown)


paul-metzger:
After a conversation with Hesham, we want to have the binaries in a subdirectory of /opt/.

heshamelmatary:
> After a conversation with Hesham, we want to have the binaries in a subdirectory of /opt/.

Probably under `/opt/cheri-api-tests` or such?

paulmetzger:
The target name is tentative -- please feel free to make suggestions.

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,180 @@
...
+from .crosscompileproject import CrossCompileMakefileProject, DefaultInstallDir, GitRepository, MakeCommandKind
+from ...config.compilation_targets import CompilationTargets
+from ...utils import classproperty
+
+
+class BuildCheriOSAPITestSuite(CrossCompileMakefileProject):
+    _always_add_suffixed_targets = True
+    do_not_add_to_targets = True
+    make_kind = MakeCommandKind.BsdMake
+    repository = GitRepository("git@github.com:CTSRD-CHERI/portable-cheribsd-test-suite.git")
```
heshamelmatary, line 41: We usually use https instead of git

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,180 @@
...
+    compiler_rt_dependency = None
+
+    @classproperty
+    def default_install_dir(self):
+        return DefaultInstallDir.ROOTFS_LOCALBASE
+
+    def compile(self, **kwargs):
+        # Search for the build directory of compiler-rt-builtins
```
heshamelmatary: Can't you just add compiler-rt-builtins to dependencies? Why do you need to search for it?

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,180 @@
...
+                break
+
+        # Musl libc's alltypes.h doesn't have a header guard by design and redefinition
+        # errors caused by this are false positives.
+        self.cross_warning_flags.append("-Wno-error=typedef-redefinition")
+
+        # We are including cheriintrin.h for some of the definitions
+        # that are on CheriBSD in other header files. However, cheric.h
+        # and cheriintrin.h define cheri_is_subset() but a comment in
+        # cheric.h indicates that the intrinsic is still unstable. Therefore,
```
heshamelmatary: Is that a CheriBSD's `cheric.h`?

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,180 @@
...
+        # we use the one in cheric.h and override the definition in
+        # cheriintrin.h if necessary.
+        self.cross_warning_flags.append("-Wno-error=macro-redefined")
+
+        # Musl libc's endian.h causes these warnings
```
heshamelmatary, line 67: Probably worth bookkeepings those somewhere so that we can properly fix in musl later

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,180 @@
...
+        # the CHERI/Morello Linux sysroot
+        self.make_args.set(
+            _DP_c="",
+            _DP_sys="",
+            _DP_thr="c",
+            _DP_pthread="c",
+        )
+
+        # We will put the binaries here
+        self.destdir = self.destdir / "rootfs" / "root"
```
heshamelmatary: Discussed offline, an enhancement is to put them in something like `/opt/cheri-os-tests` or something instead of under `/root`

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,180 @@
...
+        self.make_args.set_env(
+            C_INCLUDE_PATH="$C_INCLUDE_PATH:" + str(self.source_dir / "compat_headers"),
+            # Ignore warning about the stack protector flag being unnecessary for purecap
+            CFLAGS=" ".join(
+                [
+                    *self.default_compiler_flags(),
+                    "-Wno-error=option-ignored",
+                    "-v",
+                    "-rtlib=compiler-rt",
+                    f"-resource-dir={compiler_rt_builtins_build_dir}",
```
heshamelmatary: You'd want to use the sysroot that should already have compiler-rt installed there if you just want to link against that library, instead of this build directory. If you want the headers, you may do something like https://github.com/CTSRD-CHERI/cheribuild/blob/main/pycheribuild/projects/cross/muslc.py#L112. Also https://github.com/CTSRD-CHERI/cheribuild/blob/main/pycheribuild/projects/cross/busybox.py#L210 might be relevant

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,180 @@
...
+            # This is not supported by Morello LLVM
+            MK_CHERI_CODEPTR_RELOCS="no",
+            MAKESYSPATH=str(self.source_dir / "mk"),
+            MAKEOBJDIRPREFIX=str(self.source_dir),
+            # This property was added to _ClangBasedTargetInfo to support this specific use case.
+            OBJCOPY=self.target_info.objcopy,
+            **self.env_make_args,
+        )
+
+        self.run_make(cwd=self.source_dir / "cheribsdtest")
```
heshamelmatary, line 115: Ideally we would want to build out-of-tree, in a separate build directory as other projects.

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def set_env_make_args(self, machine_cpuarch: str, machine_abi: str, machine_arch: str):
+        self.env_make_args = {
+            "MACHINE_CPUARCH": machine_cpuarch,
+            "MACHINE_ABI": machine_abi,
+            "MACHINE_ARCH": machine_arch,
+        }
+
+    def install(self, **kwargs):
```
heshamelmatary, line 124: Is that needed? if it's just `make install` I guess the `CrossCompileMakefileProject`or `CrossCompileAutotoolsProject` parent would do that for you, without you having to define install

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def process(self):
+        self.check_required_system_tool("bmake", homebrew="bmake", cheribuild_target="bmake")
+        super().process()
+
+
+class BuildRISCVCheriOSAPITestSuite(BuildCheriOSAPITestSuite):
+    _supported_architectures = (CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093,)
+    compiler_rt_dependency = "cheri-std093-compiler-rt-builtins"
+    dependencies = ("cheri-std093-libxo", "cheri-std093-muslc", "cheri-std093-compiler-rt-builtins")
```
heshamelmatary: I'd follow @arichardson 's clean dependencies commit here if that's possible https://github.com/CTSRD-CHERI/cheribuild/commit/a35b5d8ba466192065fc171b7c9a27fc40c2064f

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,180 @@
...
+    target = "cheri-os-api-test-suite"
+    _needs_sysroot = True
+
+    def setup(self) -> None:
+        # XXXPM: This won't work for CHERI-CVA6
```
heshamelmatary, line 161: why won't it work?

pycheribuild/projects/cross/libbsd.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @classproperty
+    def default_install_dir(self):
+        return DefaultInstallDir.ROOTFS_LOCALBASE
+
+    @property
+    def muslc_target(self) -> str:
+        return self.target_info.target_triple
+
+    def _apply_patch(self) -> None:
```
heshamelmatary, line 64: If this patch is public somewhere, we can probably follow something like https://github.com/CTSRD-CHERI/cheribuild/blob/main/pycheribuild/projects/cross/linux.py#L121

pycheribuild/projects/cross/libxo.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def install(self, **kwargs):
+        self.run_make_install()
+        super().install(**kwargs)
+
+
+class BuildRISCVLibxo(BuildLibxo):
+    _supported_architectures = (CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093,)
+    _default_architecture = CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093
+    compiler_rt_dependency = "compiler-rt-builtins"
```
heshamelmatary, line 120: For all dependencies here (and other targets as well) for compiler-rt, musl, and the linux-kernel, please refactor as per my other comment (following Alex's commit), if possible

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: I'm using git for now because this repository is not yet public. With https git would ask for a username and password when cheribuild clones the repository and probably for updates too.

heshamelmatary: No problem. We just need to make it https before merging. As for being private, we can manually just clone the repo once (using git) for the first time, that's what I usually do. Then cheribuild won't complain

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: Done

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: If I just add "compiler-rt-builtins" to the dependency list without manually setting  `-resource-dir=` I get:
`ld.lld: error: cannot open /home/pffm2/cheri/output/cheri-std093-sdk/lib/clang/19/lib/riscv64-unknown-linux-musl/libclang_rt.builtins.a: No such file or directory
ld.lld: error: cannot open /home/pffm2/cheri/output/cheri-std093-sdk/lib/clang/19/lib/riscv64-unknown-linux-musl/libclang_rt.builtins.a: No such file or directory`

When I moved the CheriBSD test suite from the CheriBSD repository into this repository I took the FreeBSD bmake files with it, and adapted them to build the stand-alone test suite. This is just an interim solution and the proper long-term solution is to eventually use cmake instead. Maybe this is a symptom of repurposing FreeBSD's build system.

paul-metzger: Though, looking at the path in the error message, it might actually be unrelated to that.

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: Yes, the stand-alone variant of the CheriBSD test suite includes a copy of CheriBSD's cheric.h because some test cases use macros and functions defined in it.

I removed this comment and the `-Wno-error=macro-redefined` that followed it and added `#undef cheri_is_subset` before the redefinition of `cheri_is_subset` in the copy of `cheric.h` instead.

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: Would it be useful to create matching issues in the CHERI Alliance's musl repository?

pycheribuild/projects/cross/libbsd.py
paul-metzger: This patch isn't public somewhere else, unlike the one at the link that you posted. I created this patch to be able to use libbsd on CHERI systems.
Are you suggesting to move the patch out of the Python code and to put it somewhere else?

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: This tells cheribuild that the makefile is not in the root directory of the repository but in `<repo_root>/cheribsdtest`. This makefile and the source code will move from `<repo_root>/cheribsdtest` to the repository's root directory in the future.

pycheribuild/projects/cross/cheritestsuite.py
heshamelmatary: Can't you just use -sysroot and/or -L and add `libclang_rt.builtins.a` to your libs? Instead of letting the build system use the resource directory for trying to find where compiler-rt lib is?

pycheribuild/projects/cross/cheritestsuite.py
heshamelmatary: I'd say yes, and probably link it here as well https://github.com/CHERI-Alliance/linux/milestone/1
Probably also the cheric.h for Linux too?

pycheribuild/projects/cross/libbsd.py
heshamelmatary: The patch looks small enough, I don't have a strong opinion to put it public, just thought it's already public somewhere and this could reduce the diff/maintenance

pycheribuild/projects/cross/libxo.py
paul-metzger: Thanks, that's done now.

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: Thanks, that's done now.

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: This doesn't work sadly. On Morello I tried with:
```
-rtlib=compiler-rt -L/home/pffm2/cheri/output/morello-sdk/linux/linux-morello-purecap/lib/ -lclang_rt.builtins-aarch64.a --sysroot=/home/pffm2/cheri/output/morello-sdk/linux/linux-morello-purecap
``` 
but still got this error message 
```
ld.lld: error: unable to find library -lclang_rt.builtins-aarch64.a
ld.lld: error: cannot open /home/pffm2/cheri/output/morello-sdk/lib/clang/17/lib/linux/libclang_rt.builtins-aarch64.a: No such file or directory
ld.lld: error: cannot open /home/pffm2/cheri/output/morello-sdk/lib/clang/17/lib/linux/libclang_rt.builtins-aarch64.a: No such file or directory
```
Even though it should have found the library:
```
ls -lah /home/pffm2/cheri/output/morello-sdk/linux/linux-morello-purecap/lib/libclang_rt.builtins-aarch64.a
lrwxrwxrwx 1 pffm2 pffm2 36 Apr 20 17:03 /home/pffm2/cheri/output/morello-sdk/linux/linux-morello-purecap/lib/libclang_rt.builtins-aarch64.a -> linux/libclang_rt.builtins-aarch64.a
```

However, setting `-resource-dir` to sysroot works instead of the more complicated code that I had previously. I updated the code in the branch accordingly.

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: I left a related comment above: https://github.com/CTSRD-CHERI/cheribuild/pull/479#discussion_r3112496883
`-resource-dir` is now set to the sysroot.

pycheribuild/projects/cross/libbsd.py
paul-metzger: Ok, I'd leave it in the Python code then, because it is very small and this is the only place where it is used.

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: The purpose of this line is to tell cheribuild that the makefile is no in the root of the repository but in <repository root>/cheribsdtest. However, I updated the build script so that it builds the test suite out-of-tree in the build directory, like you requested.

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,178 @@
...
+
+        # Musl libc's endian.h causes these warnings
+        self.cross_warning_flags.append("-Wno-error=shift-op-parentheses")
+        self.cross_warning_flags.append("-Wno-error=bitwise-op-parentheses")
+        # Muls libc's ucontext.h causes this warning
+        self.cross_warning_flags.append("-Wno-error=strict-prototypes")
+
+        # Overwrite dependencies because libsys and libcompiler_rt don't exist in
+        # the CHERI/Morello Linux sysroot
+        self.make_args.set(
```
arichardson, line 77: This seems rather awkward. I think you're bundling the makefiles so could you fix it in the source repo? Or alternatively it might make sense to write a CMake/meson based build system which should be quite simple for this small project.

pycheribuild/projects/cross/libxo.py
```diff
@@ -0,0 +1,106 @@
...
+        super().install(**kwargs)
+
+
+class BuildRISCVLibxo(BuildLibxo):
+    _supported_architectures = (CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093,)
+    _default_architecture = CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093
+    target = "cheri-std093-libxo"
+
+
+class BuildMorelloLibxo(BuildLibxo):
```
arichardson, line 103: Could you merge these classes into the base and just combine the supported_architectures to include both

arichardson: Using `supported_atchitectectures = (*CrossCompileAutotoolsProject.supported_architectures, ....)`

pycheribuild/projects/cross/libmd.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class BuildRISCVLibmd(BuildLibmd):
+    _supported_architectures = (CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093,)
+    _default_architecture = CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093
+    target = "cheri-std093-libmd"
+
+
+class BuildMorelloLibmd(BuildLibmd):
+    _supported_architectures = CompilationTargets.ALL_MORELLO_LINUX_TARGETS
```
arichardson, line 83: Similar comment to below, merge these classes

pycheribuild/projects/cross/libbsd.py
```diff
@@ -0,0 +1,117 @@
...
+        super().install(**kwargs)
+
+
+class BuildRISCVLibbsd(BuildLibbsd):
+    _supported_architectures = (CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093,)
+    _default_architecture = CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093
+    target = "cheri-std093-libbsd"
+
+
+class BuildMorelloLibbsd(BuildLibbsd):
```
arichardson, line 114: Merge these classes with the base. Since they use the same source dir there is no need to have a separate class

pycheribuild/projects/cross/libxo.py
paul-metzger: Thanks, that makes sense but I ran into several issues with this. The suggested code leads to a syntax error, because the *-operator cannot be used with supported_architectures:
```
Traceback (most recent call last):
  File "/home/pffm2/postdoc/repos/cheribuild/./cheribuild.py", line 41, in <module>
    from pycheribuild.__main__ import main  # "__main__" case  # noqa: E402, PLC2701, RUF100
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pffm2/postdoc/repos/cheribuild/pycheribuild/__main__.py", line 61, in <module>
    from .projects.cross import *  # noqa: F401, F403, RUF100
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pffm2/postdoc/repos/cheribuild/pycheribuild/projects/cross/libxo.py", line 37
    _supported_architectures = (*CrossCompileAutotoolsProject.supported_architectures)
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: cannot use starred expression here
```
My next attempt was `_supported_architectures = CrossCompileAutotoolsProject.supported_architectures()`, which doesn't lead to an error at first but supported_architectures() doesn't return any Linux targets causing other problems:
```
print(CrossCompileAutotoolsProject.supported_architectures())
(CheriBSDTargetInfo(RISCV64), CheriBSDTargetInfo(AARCH64), CheriBSDTargetInfo(X86_64), CheriBSDTargetInfo(RISCV64 purecap), CheriBSDMorelloTargetInfo(AARCH64 purecap), NativeTargetInfo(X86_64))
```
Finally I tried `_supported_architectures = (*CompilationTargets.ALL_CHERI_LINUX_TARGETS, *CompilationTargets.ALL_MORELLO_LINUX_TARGETS)` but this leads, surprisingly to this error:
```
  File "/home/pffm2/postdoc/repos/cheribuild/./cheribuild.py", line 41, in <module>
    from pycheribuild.__main__ import main  # "__main__" case  # noqa: E402, PLC2701, RUF100
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pffm2/postdoc/repos/cheribuild/pycheribuild/__main__.py", line 61, in <module>
    from .projects.cross import *  # noqa: F401, F403, RUF100
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pffm2/postdoc/repos/cheribuild/pycheribuild/projects/cross/libxo.py", line 34, in <module>
    class BuildLibxo(CrossCompileAutotoolsProject):
  File "/home/pffm2/postdoc/repos/cheribuild/pycheribuild/projects/simple_project.py", line 1752, in __init__
    target_manager.add_target(MultiArchTarget(new_name, new_cls, arch, base_target))
  File "/home/pffm2/postdoc/repos/cheribuild/pycheribuild/targets.py", line 393, in add_target
    assert target.name not in self._all_targets, "Duplicate target " + target.name
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: Duplicate target libxo-linux-morello-purecap
```
_supported_architectures doesn't contain any duplicates and so this might point at a bug elsewhere.
```
print((*CompilationTargets.ALL_CHERI_LINUX_TARGETS, *CompilationTargets.ALL_MORELLO_LINUX_TARGETS))
(CheriLinuxTargetInfo(RISCV64 purecap), CheriLinuxTargetInfo(RISCV64), CheriLinuxTargetInfo(AARCH64), CheriLinuxWithMorelloCompilerTargetInfo(AARCH64 purecap), MorelloLinuxTargetInfo(AARCH64 purecap), MorelloLinuxTargetInfo(AARCH64))
```

heshamelmatary: The problem is both CHERI_LINUX_MORELLO_PURECAP and MORELLO_LINUX_PURECAP target infos have the same `morello-purecap` name. I reported that before. The workaround is to not have both in the same target, but rather, have a separate `morello-xxx` targets that only support `MORELLO_LINUX_PURECAP`

heshamelmatary: See https://github.com/CTSRD-CHERI/cheribuild/pull/478#discussion_r3081082975

arichardson: I think the best way to fix this would be to give a unique suffix, maybe the `MORELLO_LINUX_PURECAP` should be `legacy-morello-purecap`?

heshamelmatary: @arichardson Agreed. Though I'd avoid renaming it now/soon and making it legacy until we finalise the convergence of Morello support into CHERI-Linux; this is still WIP

paul-metzger: I've seen @heshamelmatary comment only after creating a commit that changes the suffix to legacy-morello-purecap. The patch set of this PR includes now this commits and commits that unify the cheribuild targets. I'm happy to remove them if the consensus is that changing the suffix should be postponed.

pycheribuild/projects/cross/libmd.py
paul-metzger: Done, refer to https://github.com/CTSRD-CHERI/cheribuild/pull/479#discussion_r3130938859

pycheribuild/projects/cross/libbsd.py
paul-metzger: Done, refer to https://github.com/CTSRD-CHERI/cheribuild/pull/479#discussion_r3130938859

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: Done, I fixed this in the test suite repository instead, as suggested.

The bundled makefiles are just an interim solution that will be replaced by cmake. However, these tests would be very useful to @heshamelmatary and potentially others, including work on a CI setup for the CHERI Alliance Linux repositories, and so I created this PR before replacing the makefiles with cmake.

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: Ok, this is a https URL now.

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: This comment was a bit unclear. What I meant is that CHERI-CVA6 will likely need different arguments for `machine_cpuarch` and `machine_arch`. The script doesn't have CHERI-CVA6 in its supported targets yet, but it will be added once someone is working on CHERI-CVA6 Linux targets, and so I removed this comment.

pycheribuild/projects/cross/libxo.py
paul-metzger: @arichardson and @heshamelmatary, please let me know whether you'd like to keep the changed suffix and the unified targets. If you want to keep them then I'll squash the respective commits into others to clean up commits in this PR.

heshamelmatary: I wouldn't really want to mark the current Morello Linux as "legacy" until we have a proper Morello support in CHERI-Linux (which might take time)

heshamelmatary: Had a quick chat with @paul-metzger, maybe we could rename the current `CHERI_LINUX_MORELLO_PURECAP` target info to `wip-morello-purecap` or such in the time being? @arichardson what do you think?

arichardson: Yes that also sounds good. I think we should just make sure the suffixes are different, it doesn't really matter which one gets the extra component.

paul-metzger: Ok, I changed the suffix of CHERI Linux Morello targets to `wip-morello-purecap` and `wip-aarch64` and reverted the suffix of the Morello Linux targets back to `morello-purecap`. @arichardson @heshamelmatary would it make sense to merge this change first and then Hesham could remove the `morello-` prefix from the Morello Linux targets, if we want this prefix to be removed.

paul-metzger: I cleaned up the commits that unify my new targets by squashing them into other commits.

pycheribuild/projects/cross/cheritestsuite.py
heshamelmatary: I always remove `-rtlib=compiler-rt` and `-resource-dir` and just manually include either -sysroot or -I (for headers) -L (for libs) directly, when cross-compiling

pycheribuild/config/compilation_targets.py
```diff
@@ -1654,10 +1658,10 @@ class CompilationTargets(BasicCompilationTargets):
         _cheri_isa=RiscvCheriISA.EXPERIMENTAL_STD093,
     )
     UPSTREAM_LINUX_AARCH64 = CrossCompileTarget("aarch64", CPUArchitecture.AARCH64, UpstreamLinuxTargetInfo)
-    CHERI_LINUX_AARCH64 = CrossCompileTarget("aarch64", CPUArchitecture.AARCH64, CheriLinuxTargetInfo)
+    CHERI_LINUX_AARCH64 = CrossCompileTarget("wip-aarch64", CPUArchitecture.AARCH64, CheriLinuxTargetInfo)
```
heshamelmatary: I haven't tested this, and I don't think it's wip, theoretically. Though I understand it's just a rename to avoid duplicate names

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,159 @@
...
+from ...config.compilation_targets import CompilationTargets, LinuxTargetInfoBase
+from ...config.target_info import CPUArchitecture, RiscvCheriISA
+from ...utils import classproperty
+
+
+class BuildCheriAPITests(CrossCompileMakefileProject):
+    _always_add_suffixed_targets = True
+    _needs_sysroot = True
+    _supported_architectures = (
+        CompilationTargets.CHERI_LINUX_MORELLO_PURECAP,
```
heshamelmatary: Can we remove all `CHERI_LINUX_MORELLO_PURECAP` until we have that working and just use `MORELLO_LINUX_MORELLO_PURECAP` for all targets here?

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,159 @@
...
+        CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093,
+        CompilationTargets.MORELLO_LINUX_MORELLO_PURECAP,
+    )
+    _default_architecture = CompilationTargets.CHERI_LINUX_MORELLO_PURECAP
```
heshamelmatary: Please use `MORELLO_LINUX_MORELLO_PURECAP` instead

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,159 @@
...
+    repository = GitRepository("https://github.com/CTSRD-CHERI/portable-cheribsd-test-suite.git")
+
+    @classproperty
+    def default_install_dir(self):
+        return DefaultInstallDir.ROOTFS_LOCALBASE
+
+    @classmethod
+    def dependencies(cls, config) -> "tuple[str, ...]":
+        ti = typing.cast(typing.Type[LinuxTargetInfoBase], cls.get_crosscompile_target().target_info_cls)
+        return ti.compiler_rt_target, ti.musl_target, "libxo"
```
heshamelmatary: theoretically speaking, musl_target (with -d) will build compiler-rt and other stuff as well, but I haven't tested that.

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def setup(self) -> None:
+        if self.get_crosscompile_target().is_aarch64(include_purecap=True):
+            self.set_env_make_args(machine_cpuarch="aarch64c", machine_abi="purecap", machine_arch="aarch64c")
+        elif self.get_crosscompile_target().is_experimental_cheri093_std(self.config):
+            self.set_env_make_args(
+                machine_cpuarch="rv64imafdczcherihybrid_zcherilevels",
```
heshamelmatary:
```suggestion
                machine_cpuarch="rv64imafdc_zcherihybrid_zcherilevels",
```

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,159 @@
...
+                machine_abi="purecap",
+                machine_arch="rv64imafdczcherihybrid_zcherilevels",
```
heshamelmatary:
```suggestion
                machine_arch="rv64imafdc_zcherihybrid_zcherilevels",
```

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,159 @@
...
+        self.make_args.set_env(
+            C_INCLUDE_PATH="$C_INCLUDE_PATH:" + str(self.source_dir / "compat_headers"),
+            # Ignore warning about the stack protector flag being unnecessary for purecap
+            CFLAGS=" ".join(
+                [
+                    *self.default_compiler_flags(),
+                    # Don't link with libgcc_s
+                    "-rtlib=compiler-rt",
+                    # It won't find libclang_rt.builtins-riscv64.a without this
+                    f"-resource-dir={self.rootfs_dir}",
```
heshamelmatary: Please see my other comment about providing flags for finding libs, headers, etc without `-rtlib=` and `-resource-dir`

pycheribuild/projects/cross/libxo.py
```diff
@@ -0,0 +1,91 @@
...
+    repository = GitRepository("https://github.com/Juniper/libxo.git")
+
+    @classproperty
+    def default_install_dir(self):
+        return DefaultInstallDir.ROOTFS_LOCALBASE
+
+    @classmethod
+    def dependencies(cls, config) -> "tuple[str, ...]":
+        ti = typing.cast(typing.Type[LinuxTargetInfoBase], cls.get_crosscompile_target().target_info_cls)
+        return ti.compiler_rt_target, ti.musl_target, "libbsd", "libmd"
```
heshamelmatary: musl will have compiler_rt as dep so no need to add it here?

heshamelmatary: though that's dynamically linked musl to be specific

pycheribuild/projects/cross/libbsd.py
```diff
@@ -0,0 +1,102 @@
...
+    def _apply_patch(self) -> None:
+        patch = """
+diff --git a/src/merge.c b/src/merge.c
+index 3f1b3fb..f8cb602 100644
+--- a/src/merge.c
++++ b/src/merge.c
+@@ -84,8 +84,8 @@ static void insertionsort(unsigned char *, size_t, size_t,
+  */
+ /* Assumption: PSIZE is a power of 2. */
+ #define EVAL(p) (unsigned char **)					\\
```
arichardson: Could we send a patch upstream using something non-CHERI-specific like
```c
#include <stdint.h>

/* Assumption: PSIZE is a power of 2. */
#define EVAL(p) (unsigned char **)( \
    ((uintptr_t)(p) + PSIZE - 1) & ~(size_t)(PSIZE - 1))
```
or if uintptr_t is not possible
```c
/* Calculate the bytes needed to reach the next PSIZE boundary and add to p */
#define EVAL(p) (unsigned char **)( \
    (unsigned char *)(p) + ((PSIZE - ((size_t)(p) & (PSIZE - 1))) & (PSIZE - 1)))
```
There is also https://clang.llvm.org/docs/LanguageExtensions.html#alignment-builtins
which allows
```c
#ifdef __clang__
#define EVAL(p) (unsigned char **)__builtin_align_up(p, PSIZE)
#else
// old code
#endif
```

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: Thanks, neither `-rtlib` nor `-resource-dir` are now used.

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: As mentioned above, neither -rtlib nor -resource-dir are now used.

pycheribuild/projects/cross/cheritestsuite.py
heshamelmatary: Stale?

pycheribuild/projects/cross/libmd.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class BuildLibmd(CrossCompileAutotoolsProject):
+    _always_add_suffixed_targets = True
+    _can_use_autogen_sh = True
+    _supported_architectures = (
+        CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093,
+        *CompilationTargets.ALL_MORELLO_LINUX_TARGETS,
+    )
+    _default_architecture = CompilationTargets.CHERI_LINUX_MORELLO_PURECAP
```
heshamelmatary: stale?

pycheribuild/projects/cross/cheritestsuite.py
heshamelmatary: Sounds good. I wanted to know why CVA6-CHERI would need changing arguments? Same RVY flags seem to work on QEMU and hardware so far unless there's something we need to be aware of

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,147 @@
...
+    repository = GitRepository("https://github.com/CTSRD-CHERI/portable-cheribsd-test-suite.git")
+
+    @classproperty
+    def default_install_dir(self):
+        return DefaultInstallDir.ROOTFS_LOCALBASE
+
+    @classmethod
+    def dependencies(cls, config) -> "tuple[str, ...]":
+        ti = typing.cast(typing.Type[LinuxTargetInfoBase], cls.get_crosscompile_target().target_info_cls)
+        return ti.compiler_rt_target, ti.musl_target, "libxo", "libbsd"
```
heshamelmatary: same as my other comment, I don't think you need to add `ti.compiler_rt_target` as musl already has it as a dependency

pycheribuild/projects/cross/libmd.py
```diff
@@ -0,0 +1,77 @@
...
+    repository = GitRepository("https://git.hadrons.org/git/libmd.git")
+
+    @classproperty
+    def default_install_dir(self):
+        return DefaultInstallDir.ROOTFS_LOCALBASE
+
+    @classmethod
+    def dependencies(cls, config) -> "tuple[str, ...]":
+        ti = typing.cast(typing.Type[LinuxTargetInfoBase], cls.get_crosscompile_target().target_info_cls)
+        return ti.compiler_rt_target, ti.musl_target
```
heshamelmatary: same for `ti.compiler_rt_target`

pycheribuild/projects/cross/libmd.py
```diff
@@ -0,0 +1,77 @@
...
+
+    @property
+    def muslc_target(self) -> str:
```
heshamelmatary: Now sure what this is for?

pycheribuild/config/compilation_targets.py
paul-metzger: Wouldn't it be fair to call it WiP if its untested?

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: Yes, that's stale. I've changed the default architecture to `CHERI_LINUX_RISCV64_PURECAP_093` for all targets that I'm adding.

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: Thanks, `ti.compiler_rt_target` is not in the dependency tuple of any target added with this PR any more.

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: Thanks, that's changed.

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: Thanks, done (see above)

pycheribuild/projects/cross/libxo.py
paul-metzger: Thanks, it isn't in the dependency list any more.

pycheribuild/projects/cross/libmd.py
paul-metzger: Yes, that's now `CHERI_LINUX_RISCV64_PURECAP_093`.

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: Thanks, that's not part of the dependency list any more.

pycheribuild/projects/cross/libmd.py
paul-metzger: Thanks, that's done (see comment above).

pycheribuild/projects/cross/libmd.py
paul-metzger: Thanks for spotting this unused leftover. It's bee removed.

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: The wording in this comment wasn't precise enough. I assumed that CVA6 will have different arguments because its a different ISA extension for RISC-V and maybe doesn't have the same CHERI extensions or the extensions are named differently. However, this assumption might be wrong.

pycheribuild/projects/cross/libbsd.py
paul-metzger: @arichardson Yes, of course. Are you happy for me to use your code for an upstream patch? Or would you like to create the patch? I'd be happy to do this.

arichardson: Please do go ahead

pycheribuild/projects/cross/libbsd.py
```diff
@@ -0,0 +1,98 @@
...
+    _supported_architectures = (
+        CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093,
+        CompilationTargets.MORELLO_LINUX_MORELLO_PURECAP,
+    )
+    _default_architecture = CompilationTargets.CHERI_LINUX_RISCV64_PURECAP_093
+    make_kind = MakeCommandKind.GnuMake
+    # `default_branch` is set because the build scripts assume that
+    # a git tag is checked out. In more detail ./get-version returns an
+    # empty string if main is checked out.
+    repository = GitRepository("https://gitlab.freedesktop.org/libbsd/libbsd.git", default_branch="0.12.2")
```
arichardson: What I've done in cases like this in the past is use temporary_url_override to point it at a fork with the patch applied to avoid needing the logic for patching inside the target.

pycheribuild/projects/cross/libbsd.py
paul-metzger: I can add "Co-authored-by:  Alexander Richardson <email address>" at the end of the commit message. Could you let me know an email address you'd like me to use?

paul-metzger: @arichardson Could you let me know if you'd like to be named as co-author in the commit for the upstream libbsd merge request? I could add 'Co-authored-by: Alexander Richardson <your email address>' at the end of the commit. I'd just need the email address that you'd like me to use.
If accepted, this commit will then be in the libbsd git history.

This is the commit:
```
commit 753c35505b31c3947185308915bd43208e29a917 (HEAD -> fix_alignment_macro, origin/fix_alignment_macro)
Author: Paul Metzger <pffm2@cl.cam.ac.uk>
Date:   Tue Apr 28 12:05:57 2026 +0100

    Fix alignment macro in merge sort
    
    The implementation of this macro relies on undefined behaviour
    as it subtracts two pointers that don't point to two elements
    in the same array (see § 6.5.6 in the working draft from April
    2023).
    
    Additionally, the integer-to-pointer conversion is causing a
    compile-time error on CHERI systems, which implement pointers with
    capabilities. On these systems, a valid capability that can be
    used to access memory cannot be derived from an integer alone.

diff --git a/src/merge.c b/src/merge.c
index 3f1b3fb..8d62587 100644
--- a/src/merge.c
+++ b/src/merge.c
@@ -84,8 +84,7 @@ static void insertionsort(unsigned char *, size_t, size_t,
  */
 /* Assumption: PSIZE is a power of 2. */
 #define EVAL(p) (unsigned char **)                                     \
-        (((unsigned char *)p + PSIZE - 1 -                             \
-          (unsigned char *)0) & ~(PSIZE - 1))
+        (((uintptr_t)(p) + PSIZE - 1) & ~(size_t)(PSIZE - 1))
 
 /*
  * Arguments are as for qsort.
```

pycheribuild/projects/cross/libbsd.py
paul-metzger: Done - I've created a patched fork of libbsd on the freedesktop GitLab instance with a view to upstreaming the patch to libbsd soon (see my comment above).

pycheribuild/config/compilation_targets.py
heshamelmatary: It's not really wip, and it's tested and it works :)

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: I created an issue in the CHERI Alliance Musl repository for one of these: https://github.com/CHERI-Alliance/musl/issues/8
The other's were not CHERI-related and benign. I changed the use of -W flags in the test suite repository and removed the -Wno flags from the cheribuild target.

pycheribuild/config/compilation_targets.py
heshamelmatary: (just tested it on Jenkins today)

heshamelmatary:
It's looking way better to me. I built and tested it locally for Morello and RVY and it's working just fine. @paul-metzger please let us know once you're ready to make the repo public, then can add it to Jenkins tests after we merge this PR

pycheribuild/config/compilation_targets.py
arichardson: I guess this one is still an open - I don't mind which one rename. I'll let @heshamelmatary decide.

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,136 @@
+#
+# Copyright (c) 2025 Paul Metzger
```
arichardson:
```suggestion
# Copyright (c) 2026 Paul Metzger
```

pycheribuild/projects/cross/cheritestsuite.py
```diff
@@ -0,0 +1,136 @@
...
+import shutil
+import typing
+
+from .crosscompileproject import CrossCompileMakefileProject, DefaultInstallDir, GitRepository, MakeCommandKind
+from ...config.compilation_targets import CompilationTargets, LinuxTargetInfoBase
+from ...config.target_info import CPUArchitecture, RiscvCheriISA
+from ...utils import classproperty
+
+
+class BuildCheriAPITests(CrossCompileMakefileProject):
```
arichardson: Should we give this an explicit target name instead of the default?

E.g. `cheri-portable-tests`?

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: >  Probably also the cheric.h for Linux too?

@heshamelmatary,  `cheric.h` is deprecated. It may be better to not bundle `cheric.h` with CHERI Musl and just bundle a copy with the test cases instead, with the intention to remove it and adapt the test cases at a later point.

pycheribuild/config/compilation_targets.py
paul-metzger: Since aarch64 CHERI Linux is functional, shall we change the suffix of aarch64 Morello Linux then from `aarch64` to `legacy-aarch64`, and leave the CHERI Linux aarch64 suffix `aarch64` instead of changing it to `wip-aarch64`?

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: I've named them `cheri-portable-os-tests`. I think Robert would like to have OS or the like in the name.

pycheribuild/projects/cross/cheritestsuite.py
paul-metzger: Thanks, I've updated the copyright in all files.

pycheribuild/projects/cross/cheritestsuite.py
brooksdavis: For things we're not sure about, I might suggest creating a local macro in the  `cheritest_*` namespace.

paul-metzger: @brooksdavis I’m not quite sure what you’re suggesting. Could you elaborate a little? For what would you like me to create a macro?

brooksdavis: For non-portable macros, the test suite should define it's own version of them until and unless they are added to cheriintrin.h

paul-metzger: Ok, thanks

paul-metzger:
I renamed the branch from `cheritest` to `cheriostest` to make it easier to find, as someone struggled to find this branch with the cheriostest targets. This caused GitHub to close this PR.

PR #481, opened 2026-04-17
linux-kernel: remove no-longer necessary patch

heshamelmatary:
upstream-linux-kernel-[aarch64|riscv64] build/boot fine without this patch. It also no longer applies, so likely fixed upstream.


arichardson:
landed upstream as https://git.kernel.org/pub/scm/linux/kernel/git/riscv/linux.git/commit/?id=bdce162f2e57a969803e5e9375999a3e0546905f, sounds like it's new enough now.

PR #483, opened 2026-04-20
linux-kernel: allow specifying defconfigs

heshamelmatary:
Similar to CheriBSD's kernel-config, this commit allows the user to specify a defconfig that is not a default eg for CVA6-CHERI or Codasip's X730 by passing its name to --linux-kernel/defconfig $XXX.

### DIFF
--- a/pycheribuild/projects/cross/linux.py
+++ b/pycheribuild/projects/cross/linux.py
@@ -35,11 +35,13 @@ from typing import Optional
 
 from .crosscompileproject import CrossCompileAutotoolsProject
 from ..project import (
+    ComputedDefaultValue,
     DefaultInstallDir,
     GitRepository,
     MakeCommandKind,
 )
 from ..run_qemu import LaunchQEMUBase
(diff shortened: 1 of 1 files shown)


heshamelmatary:
my local `ruff` doesn't complain as with CI here. Probably versioning mismatch? Can't see why exactly ruff isn't happy here, it just says file changes/failed and I can't reproduce locally

heshamelmatary:
> my local `ruff` doesn't complain as with CI here. Probably versioning mismatch? Can't see why exactly ruff isn't happy here, it just says file changes/failed and I can't reproduce locally



> my local `ruff` doesn't complain as with CI here. Probably versioning mismatch? Can't see why exactly ruff isn't happy here, it just says file changes/failed and I can't reproduce locally

fixed by upgrading my local ruff version to 15.11

arichardson:
If you install pre-commit and run `pre-commit install && pre-commit run --all-files` it should run the same version as CI.

arichardson (approved):

PR #489, opened 2026-05-01
freebsd-morello-purecap and freebsd-riscv64-purecap targets

brooksdavis:
Add bare minimum targets for purecap freebsd.

A bunch of minor cleanups prior to the main change.

### DIFF
--- a/pycheribuild/config/compilation_targets.py
+++ b/pycheribuild/config/compilation_targets.py
@@ -468,7 +468,13 @@ class FreeBSDTargetInfo(_ClangBasedTargetInfo):
             CPUArchitecture.RISCV64: "riscv64",
             CPUArchitecture.X86_64: "amd64",
         }
-        return mapping[self.target.cpu_architecture]
+        base = mapping[self.target.cpu_architecture]
+
+        if self.target.is_cheri_purecap():
+            purecap_suffix = "c"
+        else:
+            purecap_suffix = ""
+        return base + purecap_suffix
 
     @classmethod
     def base_sysroot_targets(cls, target: "CrossCompileTarget", config: "CheriConfig") -> "list[str]":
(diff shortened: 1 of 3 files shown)


arichardson (approved):
LGTM with CI passing

arichardson:
Apologies for missing this PR

pycheribuild/config/compilation_targets.py
jrtc27, line 455: *If* we're going to remove FreeBSD/mips64 support then please remove it from all these mappings too so we explode on any attempt to use it rather than potentially silently give broken output. Though given it's very little code to support it might still be worth keeping FreeBSD/mips64 specifically around a bit longer (I hear there are people on IRC who sometimes poke at it and bait me into fixing things...). The CheriBSD bits can go though, which I think is most of the de-MIPSification diff.

pycheribuild/config/compilation_targets.py
```diff
@@ -392,7 +392,7 @@ def get_riscv_abi(cls, xtarget: CrossCompileTarget, *, softfloat: Optional[bool]
 
 class FreeBSDTargetInfo(_ClangBasedTargetInfo):
     shortname: str = "FreeBSD"
-    FREEBSD_VERSION: int = 13
+    FREEBSD_VERSION: int = 16
```
jrtc27, line 395: Commit says 15, but this does 16 for FreeBSD and 15 for CheriBSD

pycheribuild/config/compilation_targets.py
```diff
@@ -702,10 +708,20 @@ def _get_run_project(self, xtarget: "CrossCompileTarget", caller: AbstractProjec
         return typing.cast(LaunchFreeBSDInterface, result)
 
 
+class FreeBSDCheriTargetInfo(FreeBSDTargetInfo):
+    shortname: str = "FreeBSD-CHERI"
```
jrtc27, line 712: Set os_prefix for these to "freebsd-" otherwise targets other than freebsd/disk-image/run/etc themselves will be foo-freebsd-cheri-riscv64-purecap and foo-freebsd-morello-morello-purecap

pycheribuild/config/compilation_targets.py
```diff
@@ -702,10 +708,20 @@ def _get_run_project(self, xtarget: "CrossCompileTarget", caller: AbstractProjec
...
+    uses_upstream_llvm: bool = False
```
jrtc27, line 713:
```suggestion
    uses_upstream_llvm: bool = False
 
    @classmethod
    def essential_compiler_and_linker_flags_impl(cls, instance: "FreeBSDCheriTargetInfo", *args, xtarget, **kwargs):
        result = super().essential_compiler_and_linker_flags_impl(instance, *args, xtarget=xtarget, **kwargs)
        # Treat FreeBSD for CHERI the same as the latest CheriBSD
        if xtarget.is_cheri_purecap():
            result.extend(cheribsd_version_dependent_flags(None, xtarget))
        return result
```
so third-party software has the right ABI, though it seems we're lacking -cheri-tgot-tls(=compat) for Morello at the moment

PR #491, opened 2026-05-21
README.md: Remove mentions of MIPS and various other updates

bsdjhb:
While CHERI LLVM still retains MIPS support, CheriBSD does not, so it is not of interest to most users.

Add a few more mentions of Morello and Morello LLVM.

Update details about GDB while here to point to CTSRD-CHERI/gdb.


README.md
bsdjhb: I'm not sure how relevant this section is.  Can this be updated for RISC-V or should it be removed?

arichardson: This section can be deleted.

arichardson (approved):

jrtc27 (approved):

jrtc27:
Looks like this needs rebasing, both for conflicts in README.md and to pick up CI fixes

PR #493, opened 2026-05-26
Config option cleanup

arichardson:
Use class variables instead of overriding setup_config_options.

This will eventually allow turning on the bad-attribute type checks.

The repetitive class-level conversion of setup_config_options->class attributes and adding tests was done using Gemini, but most of the logic changes were manual.

Confirmed the difference in default config options against HEAD using:
```
(cd /tmp && diff -u <(uv run ~/cheri/cheribuild-tmp/cheribuild.py --dump-configuration | sort | grep -Ev "test-against-running-qemu-instance|build-tests") <(uv run ~/cheri/cheribuild/cheribuild.py --dump-configuration | sort | grep -Ev "test-against-running-qemu-instance|build-tests")) > options.diff
```

This actually includes a few minor fixes where config options were not inherited properly previously, but otherwise this has no functional changes.

### DIFF
--- a/pycheribuild/config/config_loader_base.py
+++ b/pycheribuild/config/config_loader_base.py
@@ -395,18 +395,18 @@ class ConfigOptionBase(AbstractConfigOption[T]):
         return self._is_default_value
 
     def __get__(self, instance, owner) -> T:
-        assert instance is not None or not callable(self.default), (
+        # If instance is None, we can pass the class (owner) as a fallback for instance
+        # so that ComputedDefaultValue can query class attributes/methods if it handles it.
+        lookup_instance = instance if instance is not None else owner
+        assert lookup_instance is not None or not callable(self.default), (
             f"Tried to access read config option {self.full_option_name} without an object instance. "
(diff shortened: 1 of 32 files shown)


arichardson:
These are the trivial ones, I'll create a new PR with CheriBSD/Project options since they have conditional logic

jrtc27:
> The repetitive class-level conversion of setup_config_options->class attributes and adding tests was done using Gemini, but most of the logic changes were manual.

This is going to need very painstaking manual review...

jrtc27:
(No, I do not for one minute trust an LLM to vomit up something correct without appropriate human review)

arichardson:
> > The repetitive class-level conversion of setup_config_options->class attributes and adding tests was done using Gemini, but most of the logic changes were manual.
> 
> This is going to need very painstaking manual review...

I agree, fortunately it's basically move out of function, unindent, replace .add_foo with FooConfigOption, and all arguments are basically the same. To complicated to write a sensible regex replace, but letting the LLM do it and manually review isn't too bad.

arichardson:
I'll split this into a NFC and functional changes PR.

PR #494, opened 2026-05-27
Add --sqlite/build-speedtest option to the sqlite target.

qwattash:
This builds the speedtest1 binary and installs it to the sqlite install directory.

### DIFF
--- a/pycheribuild/projects/cross/sqlite.py
+++ b/pycheribuild/projects/cross/sqlite.py
@@ -28,12 +28,15 @@
 # SUCH DAMAGE.
 #
 from .crosscompileproject import CrossCompileAutotoolsProject, GitRepository
+from ..simple_project import BoolConfigOption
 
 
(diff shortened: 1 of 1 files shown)


pycheribuild/projects/cross/sqlite.py
```diff
@@ -35,6 +35,13 @@ class BuildSQLite(CrossCompileAutotoolsProject):
         "https://github.com/CTSRD-CHERI/sqlite.git", default_branch="3.22.0-cheri", force_branch=True
     )
 
+    @classmethod
```
arichardson: Could you use BoolConfigOption?

arichardson (approved):
LGTM but I'm trying to avoid new setup_config_ootion overrides

pycheribuild/projects/cross/sqlite.py
qwattash: Oh sure, I'm not fully up to date with the New Way (tm)

PR #497, opened 2026-06-02
Fix bluespec-compiler for Ubuntu 24.04 to 26.04

tmarkettos:
This PR fixes the Haskell setup for building bluespec-compiler. It also corrects a dependency apt package name.

Tested on Ubuntu 24.04, 25.10 and 26.04.

### DIFF
--- a/pycheribuild/projects/bluespec_compiler.py
+++ b/pycheribuild/projects/bluespec_compiler.py
@@ -46,7 +46,7 @@ class BuildBluespecCompiler(Project):
         self.check_required_system_tool("ghc", apt="ghc", homebrew="ghc")
         self.check_required_system_tool("cabal", apt="cabal-install", homebrew="cabal-install")
         self.check_required_system_tool("gperf", homebrew="gperf", apt="gperf")
-        self.check_required_pkg_config("tcl", apt="tcl", homebrew="tcl-tk")
+        self.check_required_pkg_config("tcl", apt="tcl-dev", homebrew="tcl-tk")
         for i in ("autoconf", "bison", "flex"):
             self.check_required_system_tool(i, homebrew=i)
 
@@ -64,7 +64,7 @@ class BuildBluespecCompiler(Project):
 
     def install(self, **kwargs):
         try:
-            self.run_cmd("cabal", "v1-update")
+            self.run_cmd("cabal", "update")
             self.run_cmd("cabal", "v1-install", "regex-compat", "syb", "old-time", "split", cwd=self.source_dir)
             self.run_make("install-src", cwd=self.source_dir)
         except Exception:
--- a/pycheribuild/projects/bluespec_compiler.py
+++ b/pycheribuild/projects/bluespec_compiler.py
@@ -65,7 +65,16 @@ class BuildBluespecCompiler(Project):
     def install(self, **kwargs):
         try:
             self.run_cmd("cabal", "update")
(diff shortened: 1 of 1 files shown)


pycheribuild/projects/bluespec_compiler.py
```diff
@@ -64,7 +64,7 @@ def compile(self, **kwargs):
 
     def install(self, **kwargs):
         try:
-            self.run_cmd("cabal", "v1-update")
+            self.run_cmd("cabal", "update")
```
jrtc27, line 67: Hm, so v1-update has been removed but v1-install remains? I'd naively assume they'd go together

tmarkettos: It seems that way - 'update; v1-install' is what the upstream Bluespec repo recommends.

I tried v2-install but ran into problems as it needs to do `make GHC="ghc -package-env default"` and I couldn't get that to work (I don't know if cheribuild was failing to escape that correctly, but even when I did it by hand it didn't work).

jrtc27: Interesting. I assume this is also fine on 22.04, but could you please check that, given it's still supported and we have a bunch of machines (e.g. technos) on it?

pycheribuild/projects/bluespec_compiler.py
```diff
@@ -64,8 +64,9 @@ def compile(self, **kwargs):
 
     def install(self, **kwargs):
         try:
-            self.run_cmd("cabal", "v1-update")
-            self.run_cmd("cabal", "v1-install", "regex-compat", "syb", "old-time", "split", cwd=self.source_dir)
+            self.run_cmd("cabal", "update")
+            self.run_cmd("cabal", "v1-install", "regex-compat", "syb", "old-time", "split",
+                         "strict-concurrency", cwd=self.source_dir)
```
jrtc27: ruff will want this as one argument per line with trailing comma, I think:
```suggestion
            self.run_cmd(
                "cabal",
                "v1-install",
                "regex-compat",
                "syb",
                "old-time",
                "split",
                "strict-concurrency",
                cwd=self.source_dir,
            )
```

pycheribuild/projects/bluespec_compiler.py
tmarkettos: Looks like the cabal install worked on 22.04, but it won't build with the supplied ghc-8.8.4:
```
[165 of 227] Compiling RSchedule        ( RSchedule.hs, /home/atm26/cheri/bsc/src/comp/../../build/comp/RSchedule.o )

RSchedule.hs:51:10: error:
    • Illegal instance declaration for ‘PPrint RAT’
        (All instance types must be of the form (T t1 ... tn)
         where T is not a synonym.
         Use TypeSynonymInstances if you want to disable this.)
    • In the instance declaration for ‘PPrint RAT’
   |
51 | instance PPrint RAT where
   |          ^^^^^^^^^^
```

tmarkettos: Now reported as https://github.com/B-Lang-org/bsc/issues/963

jrtc27: Thanks; given it got that far the cabal side of things here is probably fine. Can you please squash the formatting commits into the second one, then we can merge this?

arichardson (approved):
LGTM if it works for you

pycheribuild/projects/bluespec_compiler.py
tmarkettos: Done

PR #500, opened 2026-06-10
linux: enable 9P filesystem support

heshamelmatary:
Enable the 9P networking and virtio transport options in the kernel configuration and regenerate the configuration with olddefconfig. This allows QEMU virtfs shares to be mounted inside the guest, enabling convenient host-target file sharing.

### DIFF
--- a/pycheribuild/projects/cross/linux.py
+++ b/pycheribuild/projects/cross/linux.py
@@ -144,6 +144,13 @@ class BuildLinux(CrossCompileAutotoolsProject):
     def configure(self, **kwargs):
         self.run_make(self.defconfig, cwd=self.source_dir, parallel=False)
 
+        # Enable 9P filesystem for sharing directories between host and target
+        self._set_config("CONFIG_NET_9P")
+        self._set_config("CONFIG_NET_9P_VIRTIO")
+        self._set_config("CONFIG_NET_9P_FD")
+        self._set_config("CONFIG_9P_FS")
+        self.run_make("olddefconfig")  # regen dependencies
+
     def install(self, **kwargs):
         self.install_file(self.build_dir / "vmlinux", self.install_dir / "boot/vmlinux")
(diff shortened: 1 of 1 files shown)


pycheribuild/projects/cross/linux.py
```diff
@@ -144,6 +144,13 @@ def _apply_patch_from_url(self, patch_output_path: Path, patch_url: str):
     def configure(self, **kwargs):
         self.run_make(self.defconfig, cwd=self.source_dir, parallel=False)
 
+        # Enable 9P filesystem for sharing directories between host and target
+        self._set_config("CONFIG_NET_9P")
+        self._set_config("CONFIG_NET_9P_VIRTIO")
+        self._set_config("CONFIG_NET_9P_FD")
+        self._set_config("CONFIG_9P_FS")
+        self.run_make("olddefconfig")  # regen dependencies
```
arichardson, line 152: The comment in this function says oldconfig not olddefconfig, but if it works for you this is fine

arichardson (approved):

pycheribuild/projects/cross/linux.py
arichardson: Should probably update the comment though

heshamelmatary: `olddefconfig` is the same, it regenerates deps. However, the only difference is that `olddefconfig` assigns deps/new symbols default values without prompting the user, while `oldconfig` does.

arichardson: Since we ran defconfig before there shouldn't be any prompts right? Anyway I don't this matters either way

heshamelmatary: There won't be prompts if no new symbols got visible by the new configs, but enabling 9P reveals new symbols (as it enables a submenu for 9P features)

PR #502, opened 2026-06-16
linux, rvy: add default QEMU -cpu flags

heshamelmatary:
Currently the launch target will add -cpu codasip-a730 only if --riscv-cheri-isa experimental_std093 is passed, and will discard -cpu if not. Discarding -cpu will result in a CHERI-enabled RVY QEMU, but with extended CHERI flags/features disabled such as cheri_pte, cheri_levels, etc, in the QEMU 7.1 branch.

This PR fixes this by:
1- Always pass -cpu rv64,cheri_pte,cheri_levels for CHERI-Linux by default, without having to check for the --riscv-cheri-isa flag.
2- Use a more generic -cpu rv64 and not codasip-a730.

### DIFF
--- a/pycheribuild/config/compilation_targets.py
+++ b/pycheribuild/config/compilation_targets.py
@@ -556,7 +556,7 @@ class FreeBSDTargetInfo(_ClangBasedTargetInfo):
         rootfs_xtarget = xtarget.get_rootfs_target()
         from ..qemu_utils import QemuOptions
 
-        qemu_options = QemuOptions(rootfs_xtarget, riscv_cheri_isa=self.config.riscv_cheri_isa)
+        qemu_options = QemuOptions(rootfs_xtarget, config=self.config)
         run_instance: LaunchFreeBSDInterface = self._get_run_project(rootfs_xtarget, self.project)
         if rootfs_xtarget.cpu_architecture not in (
             CPUArchitecture.MIPS64,
--- a/pycheribuild/projects/cross/bbl.py
+++ b/pycheribuild/projects/cross/bbl.py
@@ -162,7 +162,7 @@ class BuildBBLTestPayload(BuildBBLBase):
         self.configure_args.append("--enable-print-device-tree")
 
     def run_tests(self) -> None:
-        options = QemuOptions(self.crosscompile_target, riscv_cheri_isa=self.config.riscv_cheri_isa)
+        options = QemuOptions(self.crosscompile_target, config=self.config)
         self.run_cmd(
             options.get_commandline(
                 qemu_command=BuildQEMU.qemu_binary(self),
--- a/pycheribuild/projects/cross/opensbi.py
+++ b/pycheribuild/projects/cross/opensbi.py
@@ -197,9 +197,7 @@ class BuildOpenSBI(Project):
         return cls.get_hybrid_instance(caller)._fw_jump_path()
 
     def run_tests(self):
-        options = QemuOptions(
-            self.crosscompile_target, riscv_cheri_isa=self.crosscompile_target.riscv_cheri_isa(self.config)
-        )
+        options = QemuOptions(self.crosscompile_target, config=self.config)
         abi = self.target_info.get_riscv_abi(self.crosscompile_target, softfloat=True)
         self.run_cmd(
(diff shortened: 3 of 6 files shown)


arichardson (changes requested):
Can you do this globally instead of just for Linux?

heshamelmatary:
> Can you do this globally instead of just for Linux?

@arichardson I have now moved this to run_qemu's init. We can also probably remove [this check](https://github.com/CTSRD-CHERI/cheribuild/blob/ef97ca0eb7726b51c577117cc50fc8a67c7807a0/pycheribuild/qemu_utils.py#L74) in favour of the change in this PR

paul-metzger:
> > Can you do this globally instead of just for Linux?
> 
> @arichardson I have now moved this to run_qemu's init. We can also probably remove [this check](https://github.com/CTSRD-CHERI/cheribuild/blob/ef97ca0eb7726b51c577117cc50fc8a67c7807a0/pycheribuild/qemu_utils.py#L74) in favour of the change in this PR

It looks like we can't remove the check and the corresponding `if` and `else` branches entirely because `self._qemu_arch_suffix` wouldn't be set. If we want to remove the code that sets `-cpu` and drop l730 support, then we could refactor the `if` branch to
```
if riscv_cheri_isa is RiscvCheriISA.EXPERIMENTAL_STD093:
                self._qemu_arch_suffix = f"riscv{xlen}cheristd"
else:
                self._qemu_arch_suffix = f"riscv{xlen}cheri"
```
so that `_qemu_arch_suffix` is still set.

arichardson (commented):
Can you place this in the qemu options __init__ instead? 

It just needs to be updated to use 
xtarget.is_experimental_cheri093_std(self.config) instead of just checking the command line flag. 

And yes we can use the generic CPU instead of a730

arichardson:
> Can you place this in the qemu options **init** instead?
> 
> It just needs to be updated to use xtarget.is_experimental_cheri093_std(self.config) instead of just checking the command line flag.
> 
> And yes we can use the generic CPU instead of a730

If the problem is that the config argument is not available there we could probably just add it.

heshamelmatary:
> [this check](https://github.com/CTSRD-CHERI/cheribuild/blob/ef97ca0eb7726b51c577117cc50fc8a67c7807a0/pycheribuild/qemu_utils.py#L74)



> > > Can you do this globally instead of just for Linux?
> > 
> > 
> > @arichardson I have now moved this to run_qemu's init. We can also probably remove [this check](https://github.com/CTSRD-CHERI/cheribuild/blob/ef97ca0eb7726b51c577117cc50fc8a67c7807a0/pycheribuild/qemu_utils.py#L74) in favour of the change in this PR
> 
> It looks like we can't remove the check and the corresponding `if` and `else` branches entirely because `self._qemu_arch_suffix` wouldn't be set. If we want to remove the code that sets `-cpu` and drop l730 support, then we could refactor the `if` branch to
> 
> ```
> if riscv_cheri_isa is RiscvCheriISA.EXPERIMENTAL_STD093:
>                 self._qemu_arch_suffix = f"riscv{xlen}cheristd"
> else:
>                 self._qemu_arch_suffix = f"riscv{xlen}cheri"
> ```
> 
> so that `_qemu_arch_suffix` is still set.

I think that's not even needed here. I've been running CHERI-Linux without that flag and it picks the correct QEMU's suffix

heshamelmatary:
> Can you place this in the qemu options **init** instead?
> 
> It just needs to be updated to use xtarget.is_experimental_cheri093_std(self.config) instead of just checking the command line flag.

That's what I tried first thing, but `QemuOptions` doesn't have config eg:

`AttributeError: 'QemuOptions' object has no attribute 'config'`

heshamelmatary:
> > Can you place this in the qemu options **init** instead?
> > It just needs to be updated to use xtarget.is_experimental_cheri093_std(self.config) instead of just checking the command line flag.
> > And yes we can use the generic CPU instead of a730
> 
> If the problem is that the config argument is not available there we could probably just add it.

Yeah that was the issue. I just added it now (in the last push) as an argument if that's what you meant?

arichardson (approved):

PR #503, opened 2026-06-16
u-boot updates

heshamelmatary:


pycheribuild/projects/cross/u_boot.py
```diff
@@ -70,6 +70,19 @@ class BuildUBoot(Project):
         function=uboot_install_dir, as_string="$SDK_ROOT/u-boot/riscv{32,64}{,-hybrid,-purecap}"
     )
 
+    @classmethod
```
arichardson: Can you use StringConfigOption class member instead? I'm trying to remove all calls to setup_config_options().

arichardson (approved):
LGTM with one small change

pycheribuild/projects/cross/u_boot.py
heshamelmatary: I cannot see any other project using `StringConfigOption`, is that something new? I am not sure how and where to call it (eg under init or setup or what?)

heshamelmatary: Just checked https://github.com/CTSRD-CHERI/cheribuild/pull/493/changes, I have done that now for this change. I haven't touched the other calls to `setup_config_options()` in the same file to reduce your rebase conflicts for that PR

arichardson (approved):

PR #504, opened 2026-06-17
OpenSBI payloads and u-boot options

heshamelmatary:


pycheribuild/projects/cross/opensbi.py
```diff
@@ -117,6 +121,9 @@ def setup(self):
             self.make_args.set(FW_JUMP_ADDR=self.fw_jump_addr)
         if self.fw_jump_fdt_addr:
             self.make_args.set(FW_JUMP_FDT_ADDR=self.fw_jump_fdt_addr)
+        if self.fw_payload_path:
+            assert self.fw_payload_path.exists()
```
arichardson: This might not exist yet when setup() is called. Should add this as a fatal error to compile() instead

pycheribuild/projects/cross/opensbi.py
```diff
@@ -83,6 +84,9 @@ def setup_config_options(cls, **kwargs) -> None:
         cls.fw_jump_fdt_addr = cls.add_optional_config_option(
             "fw-jump-fdt-addr", kind=str, help="OpenSBI FW_JUMP_FDT_ADDR"
         )
+        cls.fw_payload_path = cls.add_optional_config_option(
```
arichardson: Does this really need to be a config option? Can't we use a field that defaults to none and then we override in the uboot class?

pycheribuild/projects/cross/opensbi.py
arichardson: But it looks like it's only used by the subclass so maybe just move everything there and not touch this class?

pycheribuild/projects/cross/opensbi.py
heshamelmatary: it doesn't need to be for our CI and cheribuild targets, but it was useful for others (eg lowRISC) during the development process to specify custom payload path with their own, baremetal, loaders, or OSes that doesn't have to be part of the cheribuild targets. Anyway, I removed it for now, if we need it again I can submit another PR.

arichardson (approved):

PR #505, opened 2026-06-17
opensbi: optionally support fw_payload

heshamelmatary:
Current cheribuild only uses the fw_jump variant of OpenSBI while there exists other variants as well such as fw_payload that could be used. This commit enables choosing the payload variant of opensbi (which could embed u-boot and/or other OSes) and install it as an optional (QEMU's) firmware variant besides fw_jump.


arichardson (commented):
Isn't this missing the logic to actually add a payload when building opensbi?

heshamelmatary:
> Isn't this missing the logic to actually add a payload when building opensbi?

Yeah but that's a separate issue/PR here [1]. I just decoupled them because even without [1], it will still build a dummy payload and run it, even if the user mistakenly/explicitly uses this new QEMU flag without building a proper payload.

[1] https://github.com/CTSRD-CHERI/cheribuild/pull/504/

PR #506, opened 2026-06-18
Support downloading and running Debian images on CHERI-Linux on QEMU

heshamelmatary:


pycheribuild/projects/cross/linux.py
```diff
@@ -270,3 +268,110 @@ class LaunchMorelloLinux(LaunchLinuxBase):
...
+        image_arch = self.linux_arch()
+
+        image_name = f"debian-13-nocloud-{image_arch}-{image_date}-{image_build}.raw"
+        base_url = f"https://cloud.debian.org/images/cloud/trixie/{image_date}-{image_build}"
+
+        image_path = BuildCheriAllianceLinux.get_install_dir(self) / image_name
+
+        # Download the checksum file
+        sha512sums_path = BuildCheriAllianceLinux.get_install_dir(self) / "SHA512SUMS"
+        self.download_file(sha512sums_path, f"{base_url}/SHA512SUMS")
```
arichardson: Can we just hardcode the two sha hashes here instead of downloading a file with the hashes?

heshamelmatary: Yeah can do that, I just wanted to make it a bit generic in case we want to change the downloaded image date/build in the future, but yeah we likely won't do that more often anyway

pycheribuild/projects/cross/linux.py
```diff
@@ -270,3 +269,121 @@ class LaunchMorelloLinux(LaunchLinuxBase):
...
+    def setup_config_options(cls, default_ssh_port: "Optional[int]" = None, **kwargs):
+        if default_ssh_port is None:
+            add_to_port = cls.get_cross_target_index()
+            default_ssh_port = get_default_ssh_forwarding_port(add_to_port)
+        super().setup_config_options(default_ssh_port=default_ssh_port, **kwargs)
+
+    def download_debian_disk_image(self) -> Path:
+        image_date = "20260601"
+        image_build = "2496"
+        # Default to the arm64 image and re-assign for riscv64 below
```
arichardson: Can we make this an arch -> sha dict to make it easier to add others in the future?

pycheribuild/projects/cross/linux.py
```diff
@@ -270,3 +269,121 @@ class LaunchMorelloLinux(LaunchLinuxBase):
...
+    def download_debian_disk_image(self) -> Path:
+        image_name = "morello-soc.tar.xz"
+        image_path = BuildCheriAllianceLinux.get_install_dir(self) / image_name
+
+        if self.download_file(
+            image_path,
+            (
+                "https://git.morello-project.org/morello/morello-rootfs-images/-/jobs/artifacts/"
+                "morello/mainline/raw/morello-soc.tar.xz?job=build-morello-rootfs-images"
+            ),
```
arichardson, line 382: Add the sha here?

arichardson (approved):
Two small comments. Feel free to merge once fixed. 

This is really great! Looking forward to testing it.

pycheribuild/projects/cross/linux.py
heshamelmatary: I didn't add the SHA here because this link fetches the latest image from CI, which theoretically could change. Though, the last build is from 2024, so we could get away with it if we just add this image's SHA. Also Linaro's gitlab is shutting down in September, so we may need to remove this target if there will be no alternative

pycheribuild/projects/cross/linux.py
heshamelmatary: done

pycheribuild/projects/cross/linux.py
heshamelmatary: I'll merge that but we can add the SHA for the Morello Debian image later if you still prefer that

PR #509, opened 2026-06-22
vim: add required -lrt needed to link

tmarkettos:
Vim now requires -lrt otherwise there are link failures for timer functions.

Building on FreeBSD without the patch:
```
$ ./cheribuild.py vim-morello-purecap
...
ld.lld: error: undefined symbol: timer_settime
>>> referenced by os_unix.c
>>>               objects/os_unix.o:(stop_timeout)
>>> referenced by os_unix.c
>>>               objects/os_unix.o:(start_timeout)
>>> did you mean: ktimer_settime
>>> defined in: ~/cheri/output/rootfs-morello-purecap/lib/libc.so.7

ld.lld: error: undefined symbol: timer_create
>>> referenced by os_unix.c
>>>               objects/os_unix.o:(start_timeout)
>>> did you mean: ktimer_create
>>> defined in: ~/cheri/output/rootfs-morello-purecap/lib/libc.so.7

ld.lld: error: undefined symbol: timer_delete
...
```

### DIFF
--- a/pycheribuild/projects/cross/vim.py
+++ b/pycheribuild/projects/cross/vim.py
@@ -66,6 +66,7 @@ class BuildVim(CrossCompileAutotoolsProject):
             )
             # Terminal library selection also uses AC_TRY_RUN
             self.configure_args.append("--with-tlib=tinfo")
+            self.LDFLAGS.append("-lrt")
 
     def needs_configure(self):
         # Makefile exists in the source (both at the top level and within


jrtc27:
Hm, should the upstream build system not be doing this automatically?

tmarkettos:
Hmm, good question. Without the fix, autoconf is saying:
```
checking for timer_create without -lrt... yes
checking if timer_create works... configure: WARNING: cross-compiling: consider setting 'vim_cv_timer_create_works'
yes
```
so it seems to be thinking that it doesn't need -lrt.  Possibly it does on the FreeBSD host but not on the FreeBSD cross-compile target?  It seems that a build of `vim-native` does work without the fix (but doesn't break with it).

I suppose we could preset `vim_cv_timer_create_works=no` instead - I'm not sure if that's better or worse than adding -lrt in the build script?

jrtc27:
Ok, upstream's configure script is broken: https://github.com/vim/vim/blob/e31ec2ef06c7cc4ec080f2ec64923e080680128f/src/configure.ac#L4038-L4106

It uses AC_COMPILE_IFELSE, not AC_LINK_IFELSE, so what should be the vim_cv_timer_create=no vim_cv_timer_create_with_lrt=yes case will end up as vim_cv_timer_create=yes, i.e. it won't detect that -lrt is needed. When not cross-compiling it will then go on to use AC_RUN_IFELSE, which will fail to link (let alone run) if -lrt is needed, and therefore set vim_cv_timer_create_works=no, but when cross-compiling it defaults to vim_cv_timer_create_works=yes. So I see two options:

1. Add vim_cv_timer_create_works=no to the environment above, mirroring what you end up with today when compiling natively
2. Add vim_cv_timer_create=no to the environment above, reflecting what the fixed configure script would determine

Ideally we'd go fix the upstream script too.

tmarkettos:
I'm not sure I'm confident enough about the autoconf semantics to fix upstream without being worried of breaking something else. Do you feel comfortable doing that?

jrtc27:
Sure, I can look later today

tmarkettos:
I can confirm that this patch to vim's configure.ac appears to fix it without needing my patch:

```
diff --git a/src/configure.ac b/src/configure.ac
index 710cb317f..b71e55339 100644
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -4037,7 +4037,7 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM(
 
 dnl Check for timer_create. It probably requires the 'rt' library.
 AC_CACHE_CHECK([for timer_create without -lrt], [vim_cv_timer_create], [
-  AC_COMPILE_IFELSE([AC_LANG_PROGRAM([
+  AC_LINK_IFELSE([AC_LANG_PROGRAM([
   #include <time.h>
   ], [
     timer_create(CLOCK_MONOTONIC, NULL, NULL);
@@ -4051,7 +4051,7 @@ if test "x$vim_cv_timer_create" = "xno" ; then
   save_LIBS="$LIBS"
   LIBS="$LIBS -lrt"
   AC_CACHE_CHECK([for timer_create with -lrt], [vim_cv_timer_create_with_lrt], [
-    AC_COMPILE_IFELSE([AC_LANG_PROGRAM([
+    AC_LINK_IFELSE([AC_LANG_PROGRAM([
     #include <time.h>
     ], [
       timer_create(CLOCK_MONOTONIC, NULL, NULL);
```

(annoyingly, the recommended 'make autoconf' to generate ./configure doesn't work out of the box. You have to `make config` from vim toplevel, let it start running ./configure, Ctrl-C, then `make -C src autoconf`)

jrtc27:
https://github.com/vim/vim/pull/20605

And don't know why you were having issues regenerating configure, make -C src autoconf worked on a clean checkout for me.

jrtc27:
Patch has been merged, so this should no longer be needed. Setting vim_cv_timer_create_works=yes as a potential follow-up to silence the warning may make sense, but shouldn't be required (and I doubt it's the only such warning...).

PR #512, opened 2026-07-08
run_qemu, linux: add guest instructions for mounting 9p and SMB filesystems (on Debian)

heshamelmatary:


arichardson (approved):

pycheribuild/projects/cross/linux.py
```diff
@@ -152,6 +152,9 @@ def configure(self, **kwargs):
         self._set_config("CONFIG_NET_9P_FD")
         self._set_config("CONFIG_9P_FS")
 
+        # Enable SMB filesystem for sharing directories between host and target
+        self._set_config("CONFIG_CIFS")
```
arichardson, line 156: I'd prefer if we just used 9P, but I don't really mind.

heshamelmatary: Yeah I was debating whether to remove SMB commands/prints for Linux or just support it for use cases I might not be aware of.  I am happy with either, I can remove it for Linux if you still prefer

arichardson: As long as we prioritize 9p I don't really mind. I think we can merge this as is and revisit later.

PR #513, opened 2026-07-08
mtree: escape METALOG filenames according to FreeBSD's mtree(8) / VIS_OCTAL

tmarkettos:
FreeBSD's mtree file format by default uses the VIS_OCTAL format of
strsvis(3) to escape non-printable characters in filenames, namely
anything not within chars 33-126 or shell globs. Attempt to handle
non-printable characters, including UTF-8, with an escaping function.
Some basic tests are included.

(A comment indicated that the VIS_CSTYLE format should be used, which
encodes whitespace as \t, \n and \s. According to the FreeBSD
mtree(8) manpage, that is only correct when using the 'netbsd6'
flavour of mtree output. Debian's mtree(5) also specifies the
VIS_OCTAL format)

### DIFF
--- a/pycheribuild/mtree.py
+++ b/pycheribuild/mtree.py
@@ -128,8 +128,7 @@ class MtreeEntry:
     def __str__(self) -> str:
         components = [MtreePath.escape(str(self.path))]
         for k, v in self.attributes.items():
-            print("k=" + str(k) + " v=" + str(v))
-            components.append(k + "=" + shlex.quote(v))
+            components.append(k + "=" + MtreePath.escape(v))
         return " ".join(components)
 
     def __repr__(self) -> str:
--- a/pycheribuild/mtree.py
+++ b/pycheribuild/mtree.py
@@ -50,6 +50,32 @@ class MtreePath(PurePosixPath):
             pathstr = "./" + pathstr
         return pathstr
 
+    @staticmethod
+    def escape(s):
+        # mtree uses strsvis(3) (in the default flavour, in VIS_OCTAL format)
+        # to encode path names containing non-printable characters.
+        # Escape a string according to the VIS_OCTAL rules, also escaping non-printable and glob characters.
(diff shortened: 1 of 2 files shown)


pycheribuild/mtree.py
```diff
@@ -50,6 +50,32 @@ def __str__(self):
...
+        # mtree uses strsvis(3) (in the default flavour, in VIS_OCTAL format)
+        # to encode path names containing non-printable characters.
+        # Escape a string according to the VIS_OCTAL rules, also escaping non-printable and glob characters.
+        # Note that multibyte support in strsvis depends on LC_CTYPE.  For now, assume everything is UTF-8.
+        s_dest = ""
+        for i in range(0, len(s)):
+            # split string into Unicode code points
+            c = ord(s[i])
+
+            # if a non-ASCII character, convert to 3 digit octal without the 0o prefix
```
arichardson:
```suggestion
            # if a non-ASCII or glob character, convert to 3 digit octal without the 0o prefix
```

I guess we are also escaping glob/shell special chars?

pycheribuild/mtree.py
```diff
@@ -50,6 +50,32 @@ def __str__(self):
...
+                c_utf8 = chr(c).encode("utf-8")
+                # format each byte as an octal string \123
+                for b in c_utf8:
+                    b_octal = format(b, "03o")
+                    s_dest += "\\" + b_octal
+            else:
+                # append an ASCII character
+                s_dest += s[i]
+
+        # return s.upper().replace(" ", "\\s")
```
arichardson:
```suggestion

```

arichardson (approved):

arichardson:
@tmarkettos do you plan to update this with the two minor changes?

tmarkettos:
Apologies, I missed that. I applied the suggested fixes. OK to merge?

arichardson (approved):

PR #515, opened 2026-07-14
utils: make query_yes_no robust against broken terminal settings

Flowdalic:
If cheribuild is launched from a shell where the terminal was left in a broken state (e.g., by a previous QEMU crash that disabled ICRNL), Python's built-in input() function hangs when prompting the user. This occurs because input() waits for a \n character, but without ICRNL, the terminal only sends raw \r (Enter) keystrokes.

Furthermore, Python's readline library (which underpins input()) takes a snapshot of the broken terminal state on startup. This causes it to actively revert any temporary tcsetattr fixes applied right before the prompt. This commit replaces input() with a custom character-by-character reading loop using tty.setcbreak(). This completely bypasses input() and the OS line-buffering, allowing the script to manually handle \r and cleanly accept user confirmation regardless of the inherited terminal state.

### DIFF
--- a/pycheribuild/utils.py
+++ b/pycheribuild/utils.py
@@ -265,7 +265,57 @@ def query_yes_no(
         return force_result
     if not sys.__stdin__.isatty():  # pyrefly: ignore
         return default_result  # can't get any input -> return the default
-    result = input(message + yes_no_str)
+    # In some cases (e.g. after QEMU has run), the terminal can be left in a state
+    # where ICRNL is disabled, causing input() to wait forever because Enter produces \r.
+    # We bypass this by reading character-by-character in cbreak mode.
+    sys.stdout.write(message + yes_no_str)
+    sys.stdout.flush()
+    try:
+        import termios
+        import tty
+        fd = sys.__stdin__.fileno()
+        old_attrs = termios.tcgetattr(fd)
+        tty.setcbreak(fd)
+
+        result_chars = []
+        while True:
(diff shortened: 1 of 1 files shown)


jrtc27 (changes requested):
There are other ways cheribuild asks for input using `input()`, and I don't think the answer to "your terminal has a broken configuration" is "replace a call to input with 51 lines of code to do it manually".

How are you in this state, anyway? cheribuild takes great care to save and restore your tty settings after running QEMU, regardless of what QEMU does, including crashing. And even with a broken config you're not stuck, ^J will send a linefeed (and there's always ^C, ^Z, etc).

Flowdalic:
Thanks for the feedback. You make a fair point. Adding that much complexity to wrap input() is definitely heavy-handed for what is probably an environment state issue.

Regarding how I ended up in that state: I hit this during a particularly messy debugging session where QEMU and the host environment interacted poorly, leaving the tty mangled. At the time, the standard control characters (IIRC including `^C` and `^Z`) were unresponsive on my end, which prompted this patch.

However, I have not been able to reproduce the anomaly since. Given that I can not trigger the broken state again, and I agree the workaround is disproportionate, I'll go ahead and close this. Thanks for the tip regarding `^J` as a fallback.

arichardson:
cheribuild does try quite hard to restore the terminal settings after running e.g. QEMU, but if you ran it manually it won't try to set it to something expected. Maybe we should warn if certain confusing terminal settings are active?

PR #516, opened 2026-07-15
Fix AttributeError due to shutil.which fallback in get_matching_binutil

Flowdalic:
This bug is triggered when a user provides a custom compiler path:(e.g., via --clang-path /custom/bin/clang) that does not have the corresponding linker (e.g., ld.lld) in the exact same directory. When this happens, get_matching_binutil falls back to searching the system PATH using shutil.which.

However, shutil.which() returns a raw string. When the build system subsequently tries to configure Link Time Optimization (LTO) flags, linker_override_flags attempts to access the .suffix property on the returned value. Since strings do not have a .suffix property, an AttributeError is raised, crashing the build.

Fix the issue by explicitly wrapping the return value of shutil.which in a Path object before returning it, matching the function's type hint and the expectations of its callers.

### DIFF
--- a/pycheribuild/processutils.py
+++ b/pycheribuild/processutils.py
@@ -723,7 +723,9 @@ class CompilerInfo:
                 warning_message("Could not find", binutil, "in expected path", result)
                 result = None
         if not result:
-            result = shutil.which(binutil)  # fall back to the default and assume clang can find the right one
+            fallback = shutil.which(binutil)  # fall back to the default and assume clang can find the right one
+            if fallback:
+                result = Path(fallback)
         return result
 
     @property


jrtc27:
That's a lot of words for something quite simple. Did you use an LLM at all in this?

Flowdalic:
Guilty as charged. I used an LLM to draft the commit message because I wanted to ensure the trigger conditions (custom compiler paths) were documented, but I agree it ended up way too verbose for a simple type mismatch.

At the end of the day, `shutil.which()` returns a `str`, violating the `Path` return type contract.

I can adjust the commit to something much more concise. How about: 

```
Fix return type of get_matching_binutil

shutil.which() returns a str, which violates the expected Path
return type. This caused an AttributeError when callers tried
to access Path-specific properties. Wrap the fallback result 
in a Path object to satisfy the API contract.

```
Let me know if you'd like me to force-push that update

arichardson (approved):
Thanks for the fix. Weird that the type checker did not catch this! I'll just go ahead and squash and merge with your suggested shorter commit message.

PR #517, opened 2026-07-15
Add compiler and linker flag options and warning-as-errors switch

Flowdalic:
I experimented a bit with running cheribuild with different compilers and C standard libcs. This is hindered by warnings-as-errors, so this proposes an opt-out mechanism (and a mechanism to inject custom flags).


jrtc27:
I don't think we want to be going down this road, it's already awkward enough to ensure cross-builds get the right set of flags. And I don't see how the wrappers you generate can possibly work reliably with BUILD_WITH_STRICT_TMPPATH.

pycheribuild/projects/project.py
```diff
@@ -658,6 +658,28 @@ def check_system_dependencies(self) -> None:
     @classmethod
     def setup_config_options(cls, install_directory_help="", **kwargs) -> None:
         super().setup_config_options(**kwargs)
+
+        cls.warnings_as_errors = cls.add_bool_option(
+            "warnings-as-errors",
+            help="Treat compiler warnings as errors (disabling this strips -Werror from default flags)",
+            default=True,
+        )
+        cls.cflags = cls.add_list_option(
```
arichardson, line 667: I like the idea of adding a way to add extra CFLAGS/CXXFLAGS/LDFLAGS flag, but we are starting to hit the limits of argparse scalability so adding new options for every projects is actually hurting startup (and testing) speed quite noticeably.

Are you setting this in a config file and building multiple targets at once? If not we could probably work around this by reading somethig like a CHERIBUILD_CFLAGS environment variable?

pycheribuild/projects/build_qemu.py
```diff
@@ -313,7 +313,8 @@ def setup(self):
         cflags = self.default_compiler_flags("c") + self.CFLAGS
         if ccinfo.compiler != "gcc":
             # QEMU's configure script adds --extra-cflags to CXXFLAGS and these flags are rejected by g++
-            cflags += ["-Werror=implicit-function-declaration", "-Werror=incompatible-pointer-types"]
+            if self.warnings_as_errors:
```
arichardson, line 316: Are you actually hitting those errors? I added those flags after spending too much time debugging issues that would have been caught. The first one should definitely not be happening, the second one has found some real problems.

pycheribuild/projects/build_qemu.py
```diff
@@ -323,7 +324,8 @@ def setup(self):
             ldflags.append("-lpcre")
         if ldflags:
             self.configure_args.append("--extra-ldflags=" + self.commandline_to_str(ldflags))
-        cflags += ["-Werror=implicit-function-declaration", "-Werror=incompatible-pointer-types"]
```
arichardson, line 326: Maybe we just delete this and only do it for clang?

Not sure why it's here twice...

pycheribuild/projects/build_qemu.py
Flowdalic: Sorry for the late reply.

> Are you actually hitting those errors?

Yes, as stated in one of the commit message I think two things happend
1. glibc started to use the `_Generic` macro on things like `strchr()`, causing `incompatible pointer type`
2. clang 22 (?) upgraded `incompatible pointer type` to an error

I understand and totally agree, from personal experience, with using strict compiler flags. However, as newer compilers get stricter and new code is using new language features, it is nice to be able to temporary disable certain features/warnings/errors, just to see if things still build with newer compilers/libc. Therefore, it is nice to have an opt-out mechanism.

pycheribuild/projects/project.py
Flowdalic: Environment variable would work for me as well. I wonder if it should even prefixed with `CHERIBUILD_`. Or maybe simply use `EXTRA_CFLAGS`.

pycheribuild/projects/build_qemu.py
arichardson: Hmm that's annoying. I agree that we will need to drop the werror once more distributions update glibc. Unless it's an easy fix?

Flowdalic: > Unless it's an easy fix?

More like tons of easy fixes. `strchr` is used a lot. :)

jrtc27: Those should just be fixed, though I'm a bit surprised QEMU doesn't pass -std= upstream.

PR #520, opened 2026-07-20
qemu: no longer make smbd a hard dependency by default

arichardson:
We now support p9fs on all supported operating systems, so instead of making smbd a hard requirement, only enable it if a usable version was found. I recently started using a new system without samba installed and getting build errors in QEMU by default is not a nice UX considering SMBD is not strictly required any more.

### DIFF
--- a/pycheribuild/projects/build_qemu.py
+++ b/pycheribuild/projects/build_qemu.py
@@ -48,7 +48,7 @@ from .project import (
     MakeCommandKind,
     Project,
 )
-from .simple_project import BoolConfigOption, SimpleProject
+from .simple_project import BoolConfigOption, OptionalBoolConfigOption, SimpleProject
 from ..config.compilation_targets import BaremetalFreestandingTargetInfo, CompilationTargets
 from ..config.config_loader_base import ConfigOptionHandle
 from ..processutils import cached_get_homebrew_prefix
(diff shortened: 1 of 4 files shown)


jrtc27:
Can we make sure Jenkins is passing this (I assume it isn't currently) before making this default change? Presumably it'll fall over testing the old CheriBSD release that doesn't have adequate 9pfs support so you'd notice, but I think we want to make sure it's enabled there for a while still.

arichardson:
> Can we make sure Jenkins is passing this (I assume it isn't currently) before making this default change? Presumably it'll fall over testing the old CheriBSD release that doesn't have adequate 9pfs support so you'd notice, but I think we want to make sure it's enabled there for a while still.

We should add it in Jenkins that's a good point. It will obviously work right now but if a future update removes smbd it will break so safer to include it. Which branch is currently being used to test cheribsd boot? Qemu-cheri?

jrtc27:
> > Can we make sure Jenkins is passing this (I assume it isn't currently) before making this default change? Presumably it'll fall over testing the old CheriBSD release that doesn't have adequate 9pfs support so you'd notice, but I think we want to make sure it's enabled there for a while still.
> 
> We should add it in Jenkins that's a good point. It will obviously work right now but if a future update removes smbd it will break so safer to include it. Which branch is currently being used to test cheribsd boot? Qemu-cheri?

qemu-cheri needs it, yes, and then dev of course so it's not behind. (Or to put it another way: land it in dev then fast-forward emu-cheri, they're in sync at the moment so that should be fine.)

arichardson:
> > > Can we make sure Jenkins is passing this (I assume it isn't currently) before making this default change? Presumably it'll fall over testing the old CheriBSD release that doesn't have adequate 9pfs support so you'd notice, but I think we want to make sure it's enabled there for a while still.
> > 
> > 
> > We should add it in Jenkins that's a good point. It will obviously work right now but if a future update removes smbd it will break so safer to include it. Which branch is currently being used to test cheribsd boot? Qemu-cheri?
> 
> qemu-cheri needs it, yes, and then dev of course so it's not behind. (Or to put it another way: land it in dev then fast-forward emu-cheri, they're in sync at the moment so that should be fine.)

done, and I'll merge this PR at some point tomorrow.

PR #521, opened 2026-07-25
cheri-sel4-excercises: pass on board as an argument for scripts

heshamelmatary:


pycheribuild/projects/cross/cheri_sel4_exercises.py
```diff
@@ -74,13 +75,27 @@ class BuildCheriseL4Excercises(CrossCompileProject):
     # ------------------------------------------------------------------
     # Project config
     # ------------------------------------------------------------------
+    @classmethod
+    def setup_config_options(cls, **kwargs):
+        super().setup_config_options(**kwargs)
+        cls.board: str = typing.cast(
+            str,
+            cls.add_config_option(
```
arichardson: Can you use a OptionalStringConfigOption class var instead?

arichardson (changes requested):
One comment otherwise LGTM

pycheribuild/projects/cross/cheri_sel4_exercises.py
```diff
@@ -74,13 +75,21 @@ class BuildCheriseL4Excercises(CrossCompileProject):
...
+
     def configure(self, **kwargs) -> None:
         if self.crosscompile_target.is_riscv(include_purecap=True):
-            self.board = "qemu_virt_riscv64"
+            self.board = "qemu_virt_riscv64" if self.board is None else self.board
             # Build and test both baseline and purecap variants
             self.targets = ["riscv64", "riscv64-purecap"]
         elif self.crosscompile_target.is_aarch64(include_purecap=True):
-            self.board = "morello_qemu"
+            self.board = "morello_qemu" if self.board is None else self.board
```
arichardson: Seems like we should just have a per class default using the computeddefaultvalue approach, then we don't need to use an optional config option

pycheribuild/projects/cross/cheri_sel4_exercises.py
```diff
@@ -74,13 +75,21 @@ class BuildCheriseL4Excercises(CrossCompileProject):
     # ------------------------------------------------------------------
     # Project config
     # ------------------------------------------------------------------
+    board = OptionalStringConfigOption(
+        "board",
+        metavar="BOARD",
+        show_help=True,
+        default=None,
```
arichardson: Use a lambda that reads a default class variable instead?

arichardson (commented):
Just noticed one more thing. Would be good to fix that but if it's too annoying we can go for this version.

pycheribuild/projects/cross/cheri_sel4_exercises.py
heshamelmatary: Yeah that's a cleaner approach. I have now pushed and followed the same approach as the defconfig one now

pycheribuild/projects/cross/cheri_sel4_exercises.py
```diff
@@ -133,7 +150,8 @@ def _build_single_elf(
         if self.crosscompile_target.is_riscv(include_purecap=True):
             flags.append("-G0")
 
-        cmd: "list[str | Path]" = [ccc, target]
+        assert self.board is not None
```
arichardson: This assertion should now no longer be needed.
```suggestion
```

arichardson (approved):
One tiny bit otherwise good to go

arichardson (approved):

PR #522, opened 2026-07-27
vim: Install real files, not symlink farm

tmarkettos:
vim's build uses the build_via_symlink_farm flag, to make a build tree that symlinks back to the source repository as it doesn't officially support out of tree builds.  Force its build to copy the real files during 'make install' rather than copying verbatim the symlinks from the farm.


arichardson (commented):
Seems fine to me, but would like @jrtc27 to confirm.

PR #525, opened 2026-08-01
ltp: fix new CI failure

heshamelmatary:
This fixes the following new failure:

```
  ty check
  shell: /bin/bash -e {0}
  env:
    RUFF_VERSION: 0.15.11
    PYREFLY_VERSION: 0.62.0
    TY_VERSION: 0.0.32
error[unsupported-operator]: Unsupported `+=` operation
  --> pycheribuild/projects/cross/ltp.py:68:9
   |
68 |         compflags += ["--sysroot", self.install_dir]
   |         ---------^^^^-------------------------------
   |         |            |
   |         |            Has type `list[str | Path]`
   |         Has type `list[str]`
   |
```

### DIFF
--- a/pycheribuild/projects/cross/ltp.py
+++ b/pycheribuild/projects/cross/ltp.py
@@ -65,8 +65,8 @@ class BuildLTP(CrossCompileAutotoolsProject):
 
     def install(self) -> None:
         compflags = [*self.essential_compiler_and_linker_flags]
-        compflags += ["--sysroot", self.install_dir]
-        compflags += ["-isystem", self.install_dir / "usr/include"]
+        compflags += ["--sysroot", str(self.install_dir)]
+        compflags += ["-isystem", str(self.install_dir / "usr/include")]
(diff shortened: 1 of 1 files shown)


pycheribuild/projects/cross/ltp.py
```diff
@@ -65,8 +65,8 @@ def compile(self) -> None:
 
     def install(self) -> None:
         compflags = [*self.essential_compiler_and_linker_flags]
-        compflags += ["--sysroot", self.install_dir]
-        compflags += ["-isystem", self.install_dir / "usr/include"]
+        compflags += ["--sysroot", str(self.install_dir)]
+        compflags += ["-isystem", str(self.install_dir) + "/usr/include"]
```
jrtc27:
```suggestion
        compflags += ["-isystem", str(self.install_dir / "usr/include")]
```
might as well still use the Path operator?

arichardson (approved):

PR #529, opened 2026-08-04
Add a (temporary) target for the CMake build of cheri-os-test

arichardson:
Most tests should pass with `cheribuild.py cheri-os-test-cmake-riscv64-purecap --clean --build-and-test`, but there are some known failures for the hybrid case.

### DIFF
--- a/pycheribuild/projects/cross/cheri_os_test.py
+++ b/pycheribuild/projects/cross/cheri_os_test.py
@@ -43,7 +43,7 @@ class BuildCheriOSTest(CrossCompileMakefileProject):
     )
     target = "cheri-os-test"
     make_kind = MakeCommandKind.BsdMake
-    repository = GitRepository("https://github.com/CTSRD-CHERI/cheri-os-test.git", default_branch="preview")
+    repository = GitRepository("https://github.com/CTSRD-CHERI/cheri-os-test.git", default_branch="dev")
 
     @classproperty
     def default_install_dir(self):
@@ -78,10 +78,18 @@ class BuildCheriOSTest(CrossCompileMakefileProject):
                 MACHINE_ARCH=self.target_info.get_riscv_arch_string(
                     self.crosscompile_target, self.config, softfloat=False
                 ),
+                MACHINE="riscv",
             )
         else:
             target = self.target_info.target
             raise NotImplementedError(f"Unsupported architecture: {target.cpu_architecture} {target._cheri_isa}")
+
+        if self.target_info.is_freebsd():
+            self.make_args.set(TARGET_OS="FreeBSD")
+        elif self.target_info.is_linux():
+            self.make_args.set(TARGET_OS="FreeBSD")
+        else:
+            raise NotImplementedError(f"Unsupported target OS: {self.target_info.shortname}")
 
         self.make_args.set_env(
             DESTDIR=str(self.destdir),
--- a/pycheribuild/projects/cross/cheri_os_test.py
+++ b/pycheribuild/projects/cross/cheri_os_test.py
@@ -26,13 +26,21 @@
 
 import os
 
-from .crosscompileproject import CrossCompileMakefileProject, DefaultInstallDir, GitRepository, MakeCommandKind
+from .crosscompileproject import (
+    CrossCompileCMakeProject,
+    CrossCompileMakefileProject,
+    DefaultInstallDir,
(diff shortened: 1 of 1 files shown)


pycheribuild/projects/cross/cheri_os_test.py
```diff
@@ -95,3 +105,22 @@ def install(self, **kwargs):
...
+
+
+class BuildCheriOSTestCmake(CrossCompileCMakeProject):
+    _supported_architectures = (
+        CompilationTargets.ALL_LINUX_PURECAP_TARGETS + CompilationTargets.ALL_CHERIBSD_PURECAP_TARGETS
+    )
+    target = "cheri-os-test-cmake"
+    repository = GitRepository(
+        "https://github.com/CTSRD-CHERI/cheri-os-test.git",
+        default_branch="dev",
```
paul-metzger, line 117: This should be 'preview'. While the test suite is published as a preview, the 'preview' branch serves a role similar to a 'main' branch: it is from time to time updated from dev after thorough testing. @jrtc27 and others suggested to use a 'preview' branch for now until the the bmake build system has been replaced and a few other issues have been resolved. I will update the 'preview' soon after the cmake build system has been merged.

pycheribuild/projects/cross/cheri_os_test.py
```diff
@@ -95,3 +105,22 @@ def install(self, **kwargs):
...
+        temporary_url_override="https://github.com/arichardson/cheri-os-test.git",
```
paul-metzger, line 118: This should point to the CTSRD cheri-os-test repository.

pycheribuild/projects/cross/cheri_os_test.py
```diff
@@ -43,7 +43,7 @@ class BuildCheriOSTest(CrossCompileMakefileProject):
     )
     target = "cheri-os-test"
     make_kind = MakeCommandKind.BsdMake
-    repository = GitRepository("https://github.com/CTSRD-CHERI/cheri-os-test.git", default_branch="preview")
+    repository = GitRepository("https://github.com/CTSRD-CHERI/cheri-os-test.git", default_branch="dev")
```
paul-metzger, line 46: This should be 'preview' for the same reason as above.

pycheribuild/projects/cross/cheri_os_test.py
```diff
@@ -78,11 +78,19 @@ def compile(self, **kwargs):
...
+                MACHINE="riscv",
             )
         else:
             target = self.target_info.target
             raise NotImplementedError(f"Unsupported architecture: {target.cpu_architecture} {target._cheri_isa}")
 
+        if self.target_info.is_freebsd():
+            self.make_args.set(TARGET_OS="FreeBSD")
+        elif self.target_info.is_linux():
+            self.make_args.set(TARGET_OS="FreeBSD")
```
paul-metzger, line 90: This should be 'Linux' on Linux. :)

paul-metzger (changes requested):

pycheribuild/projects/cross/cheri_os_test.py
arichardson: Well yes but it needs to be merged first

pycheribuild/projects/cross/cheri_os_test.py
arichardson: Oh okay that's not what I expected. To me preview sounded like "here's an early preview to try out until there is a main/stable branch". If you consider it the stable one i would suggest renaming it now while there aren't many users of preview.

arichardson: I other words: to me "dev" sounds more stable than "preview'

pycheribuild/projects/cross/cheri_os_test.py
arichardson: Oops not sure how it actually built

pycheribuild/projects/cross/cheri_os_test.py
paul-metzger: It is an uncommon naming scheme, but we want to keep it named 'preview' for now.

pycheribuild/projects/cross/cheri_os_test.py
paul-metzger: This is merged now.

PR #530, opened 2026-08-06
libxo 2.0 treats some failing configure tests as errors

paul-metzger:
Some failing configure tests related to `gettext()` functions and `msgfmt` are now treated as errors by version 2.0.0. As a fix does not appear to be straight forward, use the preceding version 1.7.5 instead for now to allow others to continue using cheriostest.

### DIFF
--- a/pycheribuild/projects/repository.py
+++ b/pycheribuild/projects/repository.py
@@ -187,6 +187,7 @@ class GitRepository(SourceRepository):
         old_urls: "Optional[list[str]]" = None,
         default_branch: "Optional[str]" = None,
         force_branch: bool = False,
+        force_tag: Optional[str] = None,
         temporary_url_override: "Optional[str]" = None,
         url_override_reason: "typing.Any" = None,
         per_target_branches: "Optional[dict[CrossCompileTarget, TargetBranchInfo]]" = None,
@@ -205,6 +206,7 @@ class GitRepository(SourceRepository):
             self.url = url
         self._default_branch = default_branch
         self.force_branch = force_branch
+        self.force_tag = force_tag
         if per_target_branches is None:
             per_target_branches = dict()
         self.per_target_branches = per_target_branches
@@ -419,7 +421,11 @@ class GitRepository(SourceRepository):
                 clone_cmd.extend(["--depth", "1", "--no-single-branch"])
             if not skip_submodules:
                 clone_cmd.append("--recurse-submodules")
-            clone_branch = self.get_default_branch(current_project, include_per_target=False)
+            if self.force_tag:
+                # --branch supports tags since at least git 2.0.5
+                clone_branch = self.force_tag
+            else:
+                clone_branch = self.get_default_branch(current_project, include_per_target=False)
             if clone_branch:
                 clone_cmd += ["--branch", clone_branch]
             current_project.run_cmd([*clone_cmd, self.url, base_project_source_dir], cwd="/")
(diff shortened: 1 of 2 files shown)


heshamelmatary:
Thanks! LGTM (once all checks are passed), this should fix a new CI failure that started to occur today.

arichardson (commented):
THis seems fine, but would another option be to set the expected environment variables?

Also this needs force_branch so that existing checkouts work.

paul-metzger:
Setting `force_branch` leads to this error:
```
cd /home/pffm2/postdoc/repos/cheribuild && git fetch
Warning: You are trying to build the (detached) branch. You should be using 1.7.5
Would you like to change to the 1.7.5 branch? y/[N] y
cd /home/pffm2/cheri/libxo && git remote -v
Fetching changes from remote origin
cd /home/pffm2/cheri/libxo && git fetch origin
Branch 1.7.5 does not exist locally. Checking it out from origin/1.7.5
cd /home/pffm2/cheri/libxo && git checkout --track origin/1.7.5
fatal: 'origin/1.7.5' is not a commit and a branch '1.7.5' cannot be created from it
Fatal error (in target libxo-linux-riscv64-purecap): Command `git checkout --track origin/1.7.5` failed with non-zero exit code 128. Working directory was /home/pffm2/cheri/libxo
```
I assume because '1.7.5' is not actually a branch but a tag.

Could we merge this until I or someone else finds the time to set the environment variables? I had a brief go at this today and it turned out to be not straight forward.

arichardson:
I don't think this will work, let me see if I can add something like "force_tag" that checks out a given tag.

arichardson:
If this actually works if you don't have libxo cloned yet then we can merge it. I'll try work on the proper fix for allowing tags in parallel

paul-metzger:
@arichardson, @heshamelmatary  Setting `default_branch` to a tag works if the repo hasn't been cloned locally yet, but it doesn't behave as intended or fails if a local clone already exists. Just setting `default_branch` has no effect on an already existing repo and setting `default_branch` and `force_branch` leads to this kind of error:
```
Branch 1.7.5 does not exist locally. Checking it out from origin/1.7.5
cd /home/pffm2/cheri/libxo && git checkout --track origin/1.7.5
fatal: 'origin/1.7.5' is not a commit and a branch '1.7.5' cannot be created from it
Fatal error (in target libxo-linux-riscv64-purecap): Command `git checkout --track origin/1.7.5` failed with non-zero exit code 128. Working directory was /home/pffm2/cheri/libxo
```
To address this issue I now added a `force_tag` parameter to the constructor of `GitRepository()` and the matching logic.

pycheribuild/projects/repository.py
```diff
@@ -879,6 +885,36 @@ def _do_git_pull(
...
+        # It is possible to switch from a tag to a branch by setting 'default_branch'
+        # and 'force_branch'. The logic for this is implemented by _handle_branch_switch().
+
+        if self.force_tag and self.is_target_tag(current_project, src_dir, self.force_tag):
+            return
+
+        if self.force_tag:
+            current_project.run_cmd(["git", "fetch"], cwd=src_dir)
+            checkout = current_project.run_cmd(
+                ["git", "checkout", self.force_tag], allow_unexpected_returncode=True, cwd=src_dir, capture_error=True
```
arichardson: Could just drop the allow_unexpected_returncode=True then the exception will propagate.

pycheribuild/projects/repository.py
```diff
@@ -879,6 +885,36 @@ def _do_git_pull(
...
+
+    def _handle_tag_switch(self, current_project: "Project", src_dir: Path):
+        # It is possible to switch from a tag to a branch by setting 'default_branch'
+        # and 'force_branch'. The logic for this is implemented by _handle_branch_switch().
+
+        if self.force_tag and self.is_target_tag(current_project, src_dir, self.force_tag):
+            return
+
+        if self.force_tag:
+            current_project.run_cmd(["git", "fetch"], cwd=src_dir)
```
arichardson: Should this be `git fetch --tags`?

pycheribuild/projects/repository.py
```diff
@@ -879,6 +885,36 @@ def _do_git_pull(
...
+    def is_target_tag(current_project: "Project", src_dir: Path, tag: str):
+        desc = current_project.run_cmd(
+            ["git", "describe", "--exact-match", "--tags", "HEAD"],
+            allow_unexpected_returncode=True,
+            capture_output=True,
+            capture_error=True,
+            cwd=src_dir,
+            run_in_pretend_mode=_PRETEND_RUN_GIT_COMMANDS,
+        )
+        if desc.returncode == 0:
```
arichardson: This can just be `return desc.returncode == 0 and desc.stdout.decode("utf-8") == tag`

pycheribuild/projects/repository.py
```diff
@@ -879,6 +885,36 @@ def _do_git_pull(
...
+        return False
+
+    def _handle_tag_switch(self, current_project: "Project", src_dir: Path):
+        # It is possible to switch from a tag to a branch by setting 'default_branch'
+        # and 'force_branch'. The logic for this is implemented by _handle_branch_switch().
+
+        if self.force_tag and self.is_target_tag(current_project, src_dir, self.force_tag):
+            return
+
+        if self.force_tag:
```
arichardson, line 912: We need an opt out prompt here similar to the branch switch to ensure you can still have local changes:

```
if not current_project.query_yes_no("Update to correct tag?"):
            return
```

arichardson (commented):
A few small comments otherwise looking great

pycheribuild/projects/repository.py
paul-metzger: Done

pycheribuild/projects/repository.py
paul-metzger: Thanks, done. Not sure why how I could have missed that.

pycheribuild/projects/repository.py
paul-metzger: Done

pycheribuild/projects/repository.py
paul-metzger: Done. I also removed the return code check, as it's not required if we propagate the exception.

arichardson (approved):

PR #531, opened 2026-08-07
Jenkins CI: update runs to supported baselines

tmarkettos:
The CI scripts were running tests on no-longer-supported versions such as Python 3.8 and Ubuntu 20.04.  Bump forward the baselines to the oldest currently supported by these projects.

Also, tests/test_git_branch_switching.py was overly sensitive to the exact behaviour of certain versions of git which varies by OS release; now look for undesired behaviour rather than fully specifying behaviour.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -26,7 +26,7 @@ jobs:
                 run: |
                     # Run as a non-root user to avoid cheribuild errors
                     adduser --disabled-password --gecos "Not Root" notroot
-                    su -c "python3 -m pip install --user --upgrade pip pytest pytest-xdist" notroot
+                    su -c "python3 -m pip install --user --upgrade --break-system-packages pip pytest pytest-xdist" notroot
             -   name: Run basic regression tests
                 run: su -c "tests/run_basic_tests.sh" notroot
     build:
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -72,7 +72,6 @@ jobs:
                 uses: actions/setup-python@v5
                 with:
                     python-version: ${{ matrix.version }}
-                    cache: 'pip'
             -   name: Install uv
                 uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
                 with:
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -26,7 +26,8 @@ jobs:
                 run: |
                     # Run as a non-root user to avoid cheribuild errors
                     adduser --disabled-password --gecos "Not Root" notroot
-                    su -c "python3 -m pip install --user --upgrade --break-system-packages pip pytest pytest-xdist" notroot
+                    su -c "python3 -m pip config set global.break-system-packages true || true" notroot
+                    su -c "python3 -m pip install --user --upgrade pip pytest pytest-xdist" notroot
             -   name: Run basic regression tests
                 run: su -c "tests/run_basic_tests.sh" notroot
     build:
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -13,7 +13,7 @@ jobs:
     build-docker:
         strategy:
             matrix:
-                image: [ 'python:3.10', 'python:3.12', 'python:latest', 'ubuntu:22.04', 'ubuntu:24.04', 'ubuntu:26.04', 'ubuntu:latest' ]
+                image: [ 'python:3.9.6', 'python:3.12', 'python:latest', 'ubuntu:22.04', 'ubuntu:24.04', 'ubuntu:26.04', 'ubuntu:latest' ]
         runs-on: ubuntu-latest
         container:
           image: ${{ matrix.image }}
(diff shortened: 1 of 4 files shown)


tmarkettos:
I also removed python:3.8 and ubuntu:20.04 from the required checks in Settings -> Branches -> Branch protection -> 'main', to avoid blocking merging PRs for checks that no longer exist. Possibly that list wants to be expanded to include the above checks if this PR is merged.

arichardson:
Thanks for this! I do wonder if we need to keep supporting 3.9.6 for macos system python? Or should we just assume homebrew or similar is available since it is kinda needed to build anything useful. @jrtc27 what do you think?

jrtc27:
I’d rather not require homebrew Python when the system one can work just fine for the purposes of cheribuild

arichardson (approved):

tests/test_git_branch_switching.py
```diff
@@ -382,10 +382,12 @@ def test_negative_fetch_refspecs(tmp_path: Path):
...
     # In the new clone, we should not have fetched the revert-123 or users/someone/wip branches
-    assert "revert-123" not in get_all_branches(local_dir)
-    assert "users/someone/wip" not in get_all_branches(local_dir)
+    local_branches = get_all_branches(local_dir)
+    assert "revert-123" not in local_branches
+    assert "users/someone/wip" not in local_branches
+
     fetch_config = (
-        subprocess.check_output(["git", "config", "get", "--all", "remote.origin.fetch"], cwd=local_dir)
+        subprocess.check_output(["git", "config", "--get-all", "remote.origin.fetch"], cwd=local_dir)
```
arichardson, line 390: This one is a bit weird, I thought we were running this test against an older version of git, did they break backwards compatibility?

arichardson: Or do we happen to now run some version between 2.29 and whatever we were using before that didn't have this?

tmarkettos: Python 3.10 (and 3.9) Docker images are based on Debian bullseye, and that does:
```
root@bullseye:/tmp# git --version
git version 2.30.2
root@bullseye:/tmp# mkdir foo
root@bullseye:/tmp# cd foo
root@bullseye:/tmp/foo# git init
...
root@bullseye:/tmp/foo# git config get --all remote.origin.fetch
error: key does not contain a section: get
root@bullseye:/tmp/foo# git config --get-all remote.origin.fetch
root@bullseye:/tmp/foo# 
```

Bookworm:
```
root@bookworm:/tmp/foo# git --version
git version 2.39.5
root@bookworm:/tmp/foo# git config get --all remote.origin.fetch
error: key does not contain a section: get
root@bookworm:/tmp/foo# git config --get-all remote.origin.fetch
root@bookworm:/tmp/foo#
```

Trixie:
```
root@trixie:/tmp/foo# git --version
git version 2.47.3
root@trixie:/tmp/foo# git config get --all remote.origin.fetch
root@trixie:/tmp/foo# git config --get-all remote.origin.fetch
root@trixie:/tmp/foo# 
```

So looks like anything before Trixie (released July 2026) is affected.

tmarkettos:
> I’d rather not require homebrew Python when the system one can work just fine for the purposes of cheribuild

I've dropped `python-baseline` from 3.10.x to 3.9.6 (specifically, ie not 3.9.x) to include this. I suspect it's not going to make a big difference compared with 3.10.

arichardson:
Thanks! 3.10 has a few new typing improvements but really doesn't make much of a difference so this is great. I'd like give Jessica another chance to comment here before merging but this looks good to me. 

The history is a bit messy so if you're ok with it I'd just use squash and merge instead of rebase.

arichardson:
Oh actually we might also need to update the Jenkins file although I'm not sure it really adds much (any?) extra coverage over the GitHub CI so maybe it's time to drop it.

tmarkettos:
I also added a fix for the Python 3.11 Typecheck GH workflow, which is currently blocking other PRs.  @jrtc27 do you have any further comments before merging?

arichardson:
Unless @jrtc27 has any further comments I'll merge this on Monday.

jrtc27:
This needs some squashing before landing? Probably about 2 commits in the end?

.github/workflows/python-app.yml
```diff
@@ -13,19 +13,20 @@ jobs:
...
             -   uses: actions/checkout@v7
             -   name: Install Python
                 if: startsWith(matrix.image, 'ubuntu:')
-                run: apt-get update && apt-get install -y python3 python3-pip
+                run: apt-get update && apt-get install -y python3 python3-pip adduser
             -   name: Install dependencies
                 run: |
                     # Run as a non-root user to avoid cheribuild errors
                     adduser --disabled-password --gecos "Not Root" notroot
+                    su -c "python3 -m pip config set global.break-system-packages true || true" notroot
```
jrtc27, line 29: Should we switch this over to uv like the later build jobs in the file?

arichardson: Yeah sounds good I'll do that as a follow up.

arichardson:
> This needs some squashing before landing? Probably about 2 commits in the end?

I was thinking of just squash and merge for the whole series?

arichardson:
> > This needs some squashing before landing? Probably about 2 commits in the end?
> 
> I was thinking of just squash and merge for the whole series?

Or maybe better to move the test updates to a separate one

PR #532, opened 2026-08-21
Relax InstallInstructions equality assertions in SimpleProject.

qwattash:
When the install instructions for a tool have already been checked, these are cached in a private dictionary in SimpleProjectBase. The dictionary key does not take into account the compat_abi for which the tool is checked. As a result, it may occur that checking for the `ninja` tool from a project targeting native will generate an install hint using pkg64c, while the same call for a project targeting native-hybrid will generate the same hint for pkg64.
This results into incompatible hints that trigger an assertion.

This patch removes the install hint assertion, as the hint equality does not seem to be critical.


arichardson:
This seems fine, I've been meaning to clean this up anyway and drop the assertion.

arichardson (approved):

PR #535, opened 2026-09-03
RVY 0.9.9 support

qwattash:
General support for RVY draft 0.9.9.

I think the only major change is the removal of all "_093"-suffixed compilation targets.
Instead this relies on the `--riscv-cheri-isa` to be set, and incompatible targets will already trigger an early error according to `supported_riscv_cheri_standard`.
The only real difference is that for these the default ISA will no longer be std-093 and you have to explicitly specify `--riscv-cheri-isa`.
In my opinion, this is better than refactoring the target registration code to support multiple targets with the same name but different ISAs. Ultimately this will all collapse to at most 2 options, the RVY base and the CHERI ISAv10 on top of RVY.

### DIFF
--- a/pycheribuild/config/chericonfig.py
+++ b/pycheribuild/config/chericonfig.py
@@ -126,6 +126,7 @@ class AArch64FloatSimdOptions(Enum):
 class RiscvCheriISA(Enum):
     V9 = "v9"
     EXPERIMENTAL_STD093 = "experimental-std093"
+    RVY = "rvy"
 
 
 class RiscvFloatAbi(Enum):
--- a/pycheribuild/config/compilation_targets.py
+++ b/pycheribuild/config/compilation_targets.py
@@ -41,7 +41,7 @@ from functools import cached_property
 from pathlib import Path
 from typing import Optional
 
-from .chericonfig import CheriConfig, RiscvCheriISA
+from .chericonfig import CheriConfig
 from .config_loader_base import ConfigLoaderBase, ConfigOptionHandle
 from .target_info import (
     AArch64FloatSimdOptions,
@@ -1588,13 +1588,6 @@ class CompilationTargets(BasicCompilationTargets):
         is_cheri_purecap=True,
         hybrid_target=FREESTANDING_RISCV32_HYBRID,
     )
-    FREESTANDING_RISCV32_PURECAP_093 = CrossCompileTarget(
-        "riscv32-purecap",
-        CPUArchitecture.RISCV32,
-        BaremetalFreestandingTargetInfo,
-        is_cheri_purecap=True,
-        _cheri_isa=RiscvCheriISA.EXPERIMENTAL_STD093,
-    )
     FREESTANDING_RISCV64 = CrossCompileTarget("riscv64", CPUArchitecture.RISCV64, BaremetalFreestandingTargetInfo)
     FREESTANDING_RISCV64_HYBRID = CrossCompileTarget(
         "riscv64-hybrid",
@@ -1609,13 +1602,6 @@ class CompilationTargets(BasicCompilationTargets):
         BaremetalFreestandingTargetInfo,
         is_cheri_purecap=True,
         hybrid_target=FREESTANDING_RISCV64_HYBRID,
-    )
-    FREESTANDING_RISCV64_PURECAP_093 = CrossCompileTarget(
-        "riscv64-purecap",
-        CPUArchitecture.RISCV64,
-        BaremetalFreestandingTargetInfo,
-        is_cheri_purecap=True,
-        _cheri_isa=RiscvCheriISA.EXPERIMENTAL_STD093,
     )
     ALL_FREESTANDING_TARGETS = (
         FREESTANDING_AARCH64,
@@ -1702,12 +1688,8 @@ class CompilationTargets(BasicCompilationTargets):
     # Linux targets
     UPSTREAM_LINUX_RISCV64 = CrossCompileTarget("riscv64", CPUArchitecture.RISCV64, UpstreamLinuxTargetInfo)
     CHERI_LINUX_RISCV64 = CrossCompileTarget("riscv64", CPUArchitecture.RISCV64, CheriLinuxTargetInfo)
-    CHERI_LINUX_RISCV64_PURECAP_093 = CrossCompileTarget(
-        "riscv64-purecap",
-        CPUArchitecture.RISCV64,
-        CheriLinuxTargetInfo,
-        is_cheri_purecap=True,
-        _cheri_isa=RiscvCheriISA.EXPERIMENTAL_STD093,
+    CHERI_LINUX_RISCV64_PURECAP = CrossCompileTarget(
+        "riscv64-purecap", CPUArchitecture.RISCV64, CheriLinuxTargetInfo, is_cheri_purecap=True
     )
     UPSTREAM_LINUX_AARCH64 = CrossCompileTarget("aarch64", CPUArchitecture.AARCH64, UpstreamLinuxTargetInfo)
     CHERI_LINUX_AARCH64 = CrossCompileTarget("aarch64", CPUArchitecture.AARCH64, CheriLinuxTargetInfo)
@@ -1733,7 +1715,7 @@ class CompilationTargets(BasicCompilationTargets):
     )
     ALL_UPSTREAM_LINUX_TARGETS = (UPSTREAM_LINUX_AARCH64, UPSTREAM_LINUX_RISCV64)
     ALL_CHERI_LINUX_TARGETS = (
-        CHERI_LINUX_RISCV64_PURECAP_093,
+        CHERI_LINUX_RISCV64_PURECAP,
         CHERI_LINUX_RISCV64,
         CHERI_LINUX_AARCH64,
         CHERI_LINUX_MORELLO_PURECAP,
@@ -1741,7 +1723,7 @@ class CompilationTargets(BasicCompilationTargets):
     ALL_MORELLO_LINUX_TARGETS = (MORELLO_LINUX_MORELLO_PURECAP, MORELLO_LINUX_AARCH64)
     ALL_CHERI_AND_MORELLO_LINUX_TARGETS = (*ALL_CHERI_LINUX_TARGETS, *ALL_MORELLO_LINUX_TARGETS)
     ALL_LINUX_PURECAP_TARGETS = (
-        CHERI_LINUX_RISCV64_PURECAP_093,
+        CHERI_LINUX_RISCV64_PURECAP,
         CHERI_LINUX_MORELLO_PURECAP,
         MORELLO_LINUX_MORELLO_PURECAP,
     )
--- a/pycheribuild/config/compilation_targets.py
+++ b/pycheribuild/config/compilation_targets.py
@@ -364,12 +364,18 @@ class _ClangBasedTargetInfo(TargetInfo, ABC):
     def get_riscv_arch_string(cls, xtarget: CrossCompileTarget, config: CheriConfig, softfloat: Optional[bool]) -> str:
         assert xtarget.is_riscv(include_purecap=True)
         # Use the insane RISC-V arch string to enable CHERI
-        arch_string = "rv" + str(xtarget.cpu_architecture.word_bits()) + "ima"
+        if xtarget.is_riscv_y(config):
+            base = "y"
+        else:
+            base = "i"
+        arch_string = "rv" + str(xtarget.cpu_architecture.word_bits()) + base + "ma"
         if not softfloat:
             arch_string += "fd"
         arch_string += "c"
         if xtarget.is_hybrid_or_purecap_cheri():
-            if xtarget.is_experimental_cheri093_std(config):
+            if xtarget.is_riscv_y(config):
+                arch_string += "zyhybrid_zylevels1b"
+            elif xtarget.is_experimental_cheri093_std(config):
                 arch_string += "zcherihybrid_zcherilevels"
             else:
                 arch_string += "xcheri"
--- a/pycheribuild/config/target_info.py
+++ b/pycheribuild/config/target_info.py
@@ -1081,9 +1081,14 @@ class CrossCompileTarget:
         return self.is_riscv32(include_purecap) or self.is_riscv64(include_purecap)
 
     def is_experimental_cheri093_std(self, config: "CheriConfig") -> bool:
+        return self.is_hybrid_or_purecap_cheri(
+            [CPUArchitecture.RISCV32, CPUArchitecture.RISCV64]
+        ) and self.riscv_cheri_isa(config) in [RiscvCheriISA.RVY, RiscvCheriISA.EXPERIMENTAL_STD093]
+
+    def is_riscv_y(self, config: "CheriConfig") -> bool:
         return (
             self.is_hybrid_or_purecap_cheri([CPUArchitecture.RISCV32, CPUArchitecture.RISCV64])
-            and self.riscv_cheri_isa(config) == RiscvCheriISA.EXPERIMENTAL_STD093
+            and self.riscv_cheri_isa(config) == RiscvCheriISA.RVY
         )
 
     def riscv_cheri_isa(self, config: "CheriConfig") -> RiscvCheriISA:
--- a/pycheribuild/projects/cross/busybox.py
+++ b/pycheribuild/projects/cross/busybox.py
@@ -233,7 +233,7 @@ class BuildAllianceBusyBox(BuildBusyBox):
     repository = GitRepository("https://github.com/CHERI-Alliance/busybox.git")
     _supported_architectures = CompilationTargets.ALL_CHERI_LINUX_TARGETS
     _default_architecture = CompilationTargets.CHERI_LINUX_MORELLO_PURECAP
-    supported_riscv_cheri_standard = RiscvCheriISA.EXPERIMENTAL_STD093
+    supported_riscv_cheri_standard = [RiscvCheriISA.EXPERIMENTAL_STD093]
 
     def configure(self) -> None:
         self.run_make("morello_busybox_defconfig", cwd=self.source_dir)
--- a/pycheribuild/projects/cross/cheribsd.py
+++ b/pycheribuild/projects/cross/cheribsd.py
@@ -772,20 +772,20 @@ class BuildFreeBSD(BuildFreeBSDBase):
         if cls._xtarget and not cls.universe_target:
             # KERNCONF always depends on the target, so we don't inherit this config option. The only exception is
             # the global --kernel-config option that is provided for convenience and backwards compat.
-            cls.kernel_config: Optional[str] = cls.add_config_option(
+            cls.kernel_config: "Optional[list[str]]" = cls.add_list_option(
                 "kernel-config",
                 metavar="CONFIG",
                 show_help=True,
                 extra_fallback_config_names=["kernel-config"],
                 default=ComputedDefaultValue(
                     function=lambda _, p: (
-                        p.default_kernel_config() if p.has_default_buildkernel_kernel_config else None
+                        [p.default_kernel_config()] if p.has_default_buildkernel_kernel_config else None
                     ),
                     as_string="target-dependent, usually GENERIC",
                 ),
                 use_default_fallback_config_names=False,  #
                 help="The kernel configuration to use for `make buildkernel`",
-            )  # ty:ignore[invalid-assignment]
+            )
 
         if cls._xtarget is not None and cls._xtarget.is_hybrid_or_purecap_cheri():
             # When targeting CHERI we have to use CHERI LLVM
@@ -1348,7 +1348,9 @@ class BuildFreeBSD(BuildFreeBSDBase):
 
     def kernconf_list(self) -> "list[str]":
         assert self.kernel_config is not None
-        return [self.kernel_config, *self.extra_kernels]
+        assert type(self.kernel_config) is list
+        kcl = [*self.kernel_config, *self.extra_kernels]
+        return kcl
 
     def compile(self, *, mfs_root_image: "Optional[Path]" = None, sysroot_only=False, **kwargs) -> None:
         # The build seems to behave differently when -j1 is passed (it still complains about parallel make failures)
@@ -1760,14 +1762,16 @@ class BuildFreeBSD(BuildFreeBSDBase):
         Get the installed kernel path for the given kernel configuration. If no kernel config
         is given, the default kernel configuration is selected.
         """
-        return self._get_kernel_rootfs_install_path(kernconf, self.install_dir, self.kernel_config)
+        default_kernconf = self.kernel_config[0] if self.kernel_config else None
+        return self._get_kernel_rootfs_install_path(kernconf, self.install_dir, default_kernconf)
 
     def get_kern_module_path(self, kernconf: "Optional[str]" = None) -> "Optional[str]":
         """
         Get the path to provide to kern.module_path for the given kernel
         configuration if needed (i.e. the kernel is not the default one).
         """
-        if kernconf is None or kernconf == self.kernel_config:
+        default_kernconf = self.kernel_config[0] if self.kernel_config else None
+        if kernconf is None or kernconf == default_kernconf:
             return None
         return "/boot/kernel." + kernconf
 
@@ -1789,7 +1793,7 @@ class BuildFreeBSD(BuildFreeBSDBase):
         The filter parameters in kwargs are mapped to CheriBSDConfig fields.
         """
         assert self.kernel_config is not None
-        config = CheriBSDConfigTable.get_entry(self.config, self.crosscompile_target, self.kernel_config)
+        config = CheriBSDConfigTable.get_entry(self.config, self.crosscompile_target, self.kernel_config[0])
         assert config is not None, "Invalid configuration name"
         return [c.kernconf for c in filter_kernel_configs([config], platform=platform, kernel_abi=None)]
 
@@ -2161,7 +2165,7 @@ class BuildCheriBsdMfsKernel(BuildCHERIBSD):
 
     def get_kernel_configs(self, platform: "Optional[ConfigPlatform]") -> "list[str]":
         if self.kernel_config is not None:
-            return [self.kernel_config]
+            return self.kernel_config
         configs = self._get_all_kernel_configs()
         return [c.kernconf for c in filter_kernel_configs(configs, platform=platform, kernel_abi=None)]
 
--- a/pycheribuild/projects/cross/cheribsd.py
+++ b/pycheribuild/projects/cross/cheribsd.py
@@ -923,8 +923,11 @@ class BuildFreeBSD(BuildFreeBSDBase):
                 # FIXME: still needed?
                 result["WITH_CHERI"] = "1"
             else:
-                if self.crosscompile_target.is_experimental_cheri093_std(self.config):
+                if self.crosscompile_target.is_riscv_y(self.config):
                     result["TARGET_CPUTYPE"] = "rvy"
+                elif self.crosscompile_target.is_experimental_cheri093_std(self.config):
+                    result["TARGET_CPUTYPE"] = "rvy"
+                    result["CHERI_RISCV_STD_093"] = "1"
                 else:
                     result["TARGET_CPUTYPE"] = "cheri"
                 if self.compiling_for_mips(include_purecap=True):
--- a/pycheribuild/projects/cross/cheribsd.py
+++ b/pycheribuild/projects/cross/cheribsd.py
@@ -2055,6 +2055,10 @@ class BuildCHERIBSD(BuildFreeBSD):
         elif self.auto_var_init is AutoVarInit.PATTERN:
             self.make_args.set_with_options(INIT_ALL_PATTERN=True)
 
+        if self.crosscompile_target.is_riscv_y(self.config):
+            # Hybrid compat not supported yet
+            self.make_args.set_with_options(LIB64=False)
+
     def compile(self, **kwargs) -> None:
         # We could also just pass all values in KERNCONF to build all those kernels. However, if MFS_ROOT is set
         # that will apply to all those kernels and embed the rootfs even if not needed
--- a/pycheribuild/projects/cross/opensbi.py
+++ b/pycheribuild/projects/cross/opensbi.py
@@ -254,9 +254,9 @@ class BuildAllianceOpenSBI(BuildOpenSBI):
     )
     _supported_architectures = (
         CompilationTargets.FREESTANDING_RISCV32,
-        CompilationTargets.FREESTANDING_RISCV32_PURECAP_093,
+        CompilationTargets.FREESTANDING_RISCV32_PURECAP,
         CompilationTargets.FREESTANDING_RISCV64,
-        CompilationTargets.FREESTANDING_RISCV64_PURECAP_093,
+        CompilationTargets.FREESTANDING_RISCV64_PURECAP,
     )
     supported_riscv_cheri_standard = [RiscvCheriISA.RVY, RiscvCheriISA.EXPERIMENTAL_STD093]
 
@@ -275,9 +275,9 @@ class BuildAllianceOpenSBI(BuildOpenSBI):
     def get_cheri_bios(cls, caller, xtarget: CrossCompileTarget):
         assert xtarget.is_riscv(include_purecap=True), "Should only call this for RISC-V"
         if xtarget.is_riscv32(include_purecap=True):
-            bios_xtarget = CompilationTargets.FREESTANDING_RISCV32_PURECAP_093
+            bios_xtarget = CompilationTargets.FREESTANDING_RISCV32_PURECAP
         else:
-            bios_xtarget = CompilationTargets.FREESTANDING_RISCV64_PURECAP_093
+            bios_xtarget = CompilationTargets.FREESTANDING_RISCV64_PURECAP
         # This version of OpenSBI requires a purecap build to support CHERI
         proj = cls.get_instance(caller, cross_target=bios_xtarget)
         assert isinstance(proj, BuildOpenSBI)
--- a/pycheribuild/projects/cross/opensbi.py
+++ b/pycheribuild/projects/cross/opensbi.py
@@ -69,7 +69,7 @@ class BuildOpenSBI(Project):
     _default_install_dir_fn = ComputedDefaultValue(
         function=opensbi_install_dir, as_string="$SDK_ROOT/opensbi/riscv{32,64}{-hybrid,-purecap,}"
     )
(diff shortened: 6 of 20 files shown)


paul-metzger (commented):
The changes touching cheri-os-test are just minimal and look good to me. It also built the test suite fine for Linux after I added `--riscv-cheri-isa=experimental_std093`.
However, when I then try a clean Linux build to test run the test suite I get this error when building QEMU:
```
...
ERROR: Unknown target name 'riscv32y-softmmu'

Fatal error (in target cheri-std093-qemu): Command `/home/pffm2/cheri/cheri-std093-qemu/configure --disable-sdl --disable-gtk --disable-opengl --disable-virglrenderer --disable-stack-protector --disable-strip --enable-virtfs --disable-xen --disable-docs --disable-rdma --disable-werror --cxx=/usr/bin/clang++ --cc=/usr/bin/clang --make=gmake --python=/usr/bin/python3 '--extra-cflags=-O3 -Werror=format -Wno-sign-compare -Wno-unused-parameter -Wno-missing-field-initializers -Werror=incompatible-pointer-types-discards-qualifiers -Wno-address-of-packed-member -Wno-missing-field-initializers -Wno-null-pointer-subtraction -Wno-bitwise-instead-of-logical -Werror=implicit-function-declaration -Werror=incompatible-pointer-types' --disable-capstone --disable-bsd-user --disable-linux-user --enable-slirp=git --smbd=/usr/sbin/smbd --target-list=arm-softmmu,aarch64-softmmu,morello-softmmu,riscv64-softmmu,riscv64xcheri-softmmu,riscv64cheristd-softmmu,riscv32-softmmu,riscv32xcheri-softmmu,riscv32cheristd-softmmu,riscv32y-softmmu,riscv64y-softmmu,x86_64-softmmu --prefix=/home/pffm2/cheri/output/cheri-std093-sdk` failed with non-zero exit code 1
```
This is how I build Linux:
```
./cheribuild.py run-minimal-cheri-linux-riscv64-purecap --linux-kernel-riscv64-purecap/skip-update -d --qemu/no-use-smbd --riscv-cheri-isa=experimental_std093 --clean
```

qwattash:
> The changes touching cheri-os-test are just minimal and look good to me. It also built the test suite fine for Linux after I added `--riscv-cheri-isa=experimental_std093`. However, when I then try a clean Linux build to test run the test suite I get this error when building QEMU:
> 
> ```
> ...
> ERROR: Unknown target name 'riscv32y-softmmu'
> 
> Fatal error (in target cheri-std093-qemu): Command `/home/pffm2/cheri/cheri-std093-qemu/configure --disable-sdl --disable-gtk --disable-opengl --disable-virglrenderer --disable-stack-protector --disable-strip --enable-virtfs --disable-xen --disable-docs --disable-rdma --disable-werror --cxx=/usr/bin/clang++ --cc=/usr/bin/clang --make=gmake --python=/usr/bin/python3 '--extra-cflags=-O3 -Werror=format -Wno-sign-compare -Wno-unused-parameter -Wno-missing-field-initializers -Werror=incompatible-pointer-types-discards-qualifiers -Wno-address-of-packed-member -Wno-missing-field-initializers -Wno-null-pointer-subtraction -Wno-bitwise-instead-of-logical -Werror=implicit-function-declaration -Werror=incompatible-pointer-types' --disable-capstone --disable-bsd-user --disable-linux-user --enable-slirp=git --smbd=/usr/sbin/smbd --target-list=arm-softmmu,aarch64-softmmu,morello-softmmu,riscv64-softmmu,riscv64xcheri-softmmu,riscv64cheristd-softmmu,riscv32-softmmu,riscv32xcheri-softmmu,riscv32cheristd-softmmu,riscv32y-softmmu,riscv64y-softmmu,x86_64-softmmu --prefix=/home/pffm2/cheri/output/cheri-std093-sdk` failed with non-zero exit code 1
> ```
> 
> This is how I build Linux:
> 
> ```
> ./cheribuild.py run-minimal-cheri-linux-riscv64-purecap --linux-kernel-riscv64-purecap/skip-update -d --qemu/no-use-smbd --riscv-cheri-isa=experimental_std093 --clean
> ```

Good point. I think I need to update the qemu branch being used, since those are rvy specific.

pycheribuild/config/chericonfig.py
```diff
@@ -126,6 +126,7 @@ def clang_march_flag(self) -> str:
 class RiscvCheriISA(Enum):
     V9 = "v9"
     EXPERIMENTAL_STD093 = "experimental-std093"
+    RVY = "rvy"
```
heshamelmatary, line 129: std093 is still rvy, just another version?

pycheribuild/config/compilation_targets.py
```diff
@@ -364,12 +364,18 @@ def essential_compiler_and_linker_flags_impl(
     def get_riscv_arch_string(cls, xtarget: CrossCompileTarget, config: CheriConfig, softfloat: Optional[bool]) -> str:
         assert xtarget.is_riscv(include_purecap=True)
         # Use the insane RISC-V arch string to enable CHERI
-        arch_string = "rv" + str(xtarget.cpu_architecture.word_bits()) + "ima"
+        if xtarget.is_riscv_rvy(config):
```
heshamelmatary: I wonder if we should leave this as `is_riscv_cheri` for readability and being more explicit while still adding the "y" to the ISA string? `also "RVY" in my mind means "RISC-V with Y extension AKA CHERI-RISC-V", so riscv_rvy makes the "riscv" part redundant. So maybe either `is_riscv_cheri` or just `is_rvy`?

pycheribuild/config/compilation_targets.py
```diff
@@ -364,12 +364,18 @@ def essential_compiler_and_linker_flags_impl(
...
+        else:
+            base = "i"
+        arch_string = "rv" + str(xtarget.cpu_architecture.word_bits()) + base + "ma"
         if not softfloat:
             arch_string += "fd"
         arch_string += "c"
         if xtarget.is_hybrid_or_purecap_cheri():
-            if xtarget.is_experimental_cheri093_std(config):
+            if xtarget.is_riscv_rvy(config):
+                arch_string += "zyhybrid_zylevels1b"
```
heshamelmatary, line 377: I haven't caught up with the .9.9 spec, but does this mean it only supports one level? eg no local? We did support/assume 2 levels in the previous 0.9.3 release

pycheribuild/projects/cross/busybox.py
```diff
@@ -233,7 +233,7 @@ class BuildAllianceBusyBox(BuildBusyBox):
     repository = GitRepository("https://github.com/CHERI-Alliance/busybox.git")
     _supported_architectures = CompilationTargets.ALL_CHERI_LINUX_TARGETS
     _default_architecture = CompilationTargets.CHERI_LINUX_MORELLO_PURECAP
-    supported_riscv_cheri_standard = RiscvCheriISA.EXPERIMENTAL_STD093
+    supported_riscv_cheri_standard = [RiscvCheriISA.EXPERIMENTAL_STD093]
```
heshamelmatary, line 236: do we expect a target to support more than a RISC-V standard/release at a time? I am hoping this is a no (ie new targets only support rvy not xcheri, and rvy targets only need to support newer rvy specs unless it's stale and we didn't have time to bring it up to a newer s[ec), in which case `supported_riscv_cheri_standard` does not have to be a list?

pycheribuild/projects/cross/opensbi.py
```diff
@@ -254,11 +254,11 @@ class BuildAllianceOpenSBI(BuildOpenSBI):
...
     _supported_architectures = (
         CompilationTargets.FREESTANDING_RISCV32,
-        CompilationTargets.FREESTANDING_RISCV32_PURECAP_093,
+        CompilationTargets.FREESTANDING_RISCV32_PURECAP,
         CompilationTargets.FREESTANDING_RISCV64,
-        CompilationTargets.FREESTANDING_RISCV64_PURECAP_093,
+        CompilationTargets.FREESTANDING_RISCV64_PURECAP,
     )
-    supported_riscv_cheri_standard = RiscvCheriISA.EXPERIMENTAL_STD093
+    supported_riscv_cheri_standard = [RiscvCheriISA.RVY, RiscvCheriISA.EXPERIMENTAL_STD093]
```
heshamelmatary, line 261: Probably not the right place for this comment, but would it make sense to rename all `EXPERIMENTAL_STD093` to something like `RVY_093`? `EXPERIMENTAL_STD093` is still RVY. I am thinking ahead in case we have to support yet a more recent RVY release (eg > .9.9), so we just rename the old targets that only support .9.9 to `RVY_099`, while we could still have some targets that support `RVY_093`. Also a comment and/or a (big) print what the current `RiscvCheriISA.RVY` spec/release is would be really useful

pycheribuild/qemu_utils.py
```diff
@@ -68,7 +68,10 @@ def __init__(self, xtarget: CrossCompileTarget, want_debugger=False, config: Opt
             # Note: we always use the CHERI QEMU
             xlen = 32 if xtarget.is_riscv32(include_purecap=True) else 64
             self.machine_flags = ["-M", "virt"]
-            if config is not None and xtarget.is_experimental_cheri093_std(config):
+            if config is not None and xtarget.is_riscv_rvy(config):
+                self._qemu_arch_suffix = self._qemu_arch_suffix = f"riscv{xlen}y"
+                self.machine_flags.extend(["-cpu", "rv64,Svyrg=true,Zylevels1=true"])
```
heshamelmatary, line 73: ditto with the levels1 thing, why not 2?

heshamelmatary (commented):
Thanks Alfredo! Looking forward to make this the default and also bring other targets to this spec! I left some comments and also tried to build both CHERI-Linux and CHERI-seL4 (and all of their deps). I do get some build errors:

`
[1;31mFatal error [1;33m(in target cheri-std093-opensbi-baremetal-riscv64-purecap)[0m[1;31m: Project cheri-std093-opensbi-baremetal-riscv64-purecap is not compatible with the RiscvCheriISA.V9 RISC-V-CHERI variant. Supported CHERI standards: ['rvy', 'experimental_std093']. Try building with --riscv-cheri-isa=rvy[0m
[1;36mCould not find version-suffixed ld.lld in expected path /usr/lib/llvm-18/bin/ld.lld[0m
`
I think all current existing targets with "std093" in their name should support and build RVY_093 by default, without having to provide `--riscv-cheri-isa=experimental_std093` as currently the case?

pycheribuild/config/chericonfig.py
arichardson: Well we didn't call it that back then it was still zcheripurecap

heshamelmatary: Yeah seems like the "y" renaming happened in .9.5? The current ISAs split makes it feel like v9 vs "experimental-std093" vs "rvy" are completely different standards

qwattash: yeah, but it needs the distinction because there are different `-march` flags to pass.

pycheribuild/config/compilation_targets.py
qwattash: The `is_riscv_cheri` is probably too generic, because it could mean a variety of things among xcheri, zcheri, RVY. I think I prefer to keep this Y-specific, so I can perhaps rename this as `is_riscv_y`.

pycheribuild/config/compilation_targets.py
qwattash: this means there is 1 level bit, so local/global.

pycheribuild/projects/cross/busybox.py
qwattash: I expect targets to support multiple ISA standard, at least some of them. CheriBSD will support xcheri, zcheri and RVY at the same time for a while, ideally it will wind down to just xcheri and rvy. I suspect in the long run we will still have to make a distinction between RVY and CHERI ISA v10.

pycheribuild/projects/cross/opensbi.py
qwattash: I think I really don't want to proliferate ISA versions. I think we are stuck on 0.9.3 because it was hard to move the entirety of the systems and toolchains in lockstep without breaking everything at once, but I think RVY should be the "latest draft" and it should be kept in sync.
The latter about the RVY ISA version is likely useful to keep around.

pycheribuild/qemu_utils.py
qwattash: see above, that is the name of the extension in the spec.

qwattash:
> Thanks Alfredo! Looking forward to make this the default and also bring other targets to this spec! I left some comments and also tried to build both CHERI-Linux and CHERI-seL4 (and all of their deps). I do get some build errors:
> 
> `�[1;31mFatal error �[1;33m(in target cheri-std093-opensbi-baremetal-riscv64-purecap)�[0m�[1;31m: Project cheri-std093-opensbi-baremetal-riscv64-purecap is not compatible with the RiscvCheriISA.V9 RISC-V-CHERI variant. Supported CHERI standards: ['rvy', 'experimental_std093']. Try building with --riscv-cheri-isa=rvy�[0m �[1;36mCould not find version-suffixed ld.lld in expected path /usr/lib/llvm-18/bin/ld.lld�[0m` I think all current existing targets with "std093" in their name should support and build RVY_093 by default, without having to provide `--riscv-cheri-isa=experimental_std093` as currently the case?

I commented a bit on that in the PR description.
To me it feels cleaner to drive towards a rename that removes the cheri-std093 prefixes in favor of a more general cheri-alliance prefixed ones. I prefer the `--riscv-cheri-isa` to be explicit because I want to know what I'm building, but I guess I could _probably_ get the first one of the supported ISAs from the target. This is very weird though if you have multiple targets that have different ordering of supported ISAs.
Ultimately, this is a side effect of having the riscv-cheri-isa flag and not encoding this as a different target (e.g. cheribsd-riscv64zcheri-purecap or cheribsd-riscv64y-purecap). We may yet go down that path, but it is a larger refactor.

pycheribuild/config/compilation_targets.py
qwattash: Done

tests/test_git_branch_switching.py
```diff
@@ -365,7 +365,7 @@ def get_all_branches(repo_dir: Path) -> "list[str]":
     )
 
 
-@pytest.mark.skipif(GIT_VERSION < (2, 29), reason="negative refspecs require git >= 2.29")
+@pytest.mark.skipif(GIT_VERSION < (2, 45), reason="git config get requires git >= 2.45")
```
arichardson, line 368: This should be fixed by #531

pycheribuild/config/target_info.py
```diff
@@ -1081,9 +1081,14 @@ def is_riscv(self, include_purecap: Optional[bool] = None) -> bool:
         return self.is_riscv32(include_purecap) or self.is_riscv64(include_purecap)
 
     def is_experimental_cheri093_std(self, config: "CheriConfig") -> bool:
+        return self.is_hybrid_or_purecap_cheri(
+            [CPUArchitecture.RISCV32, CPUArchitecture.RISCV64]
+        ) and self.riscv_cheri_isa(config) in [RiscvCheriISA.RVY, RiscvCheriISA.EXPERIMENTAL_STD093]
```
arichardson, line 1086: Why is RiscvCheriISA.RVY accepted here? It's not 093?

pycheribuild/projects/cross/busybox.py
arichardson, line 236: Probably makes sense to rename this to "supported_riscv_cheri_standards" now that it's a list. Also I think we should use tuples to avoid this being a mutable variable

pycheribuild/config/compilation_targets.py
```diff
@@ -1588,13 +1588,6 @@ class CompilationTargets(BasicCompilationTargets):
         is_cheri_purecap=True,
         hybrid_target=FREESTANDING_RISCV32_HYBRID,
     )
-    FREESTANDING_RISCV32_PURECAP_093 = CrossCompileTarget(
```
arichardson, line 1591: The advantage of this approach was that you could build a 093-only target such as Linux without having to pass --riscv-cheri-isa. Is this still possible or do we get a hard error?

pycheribuild/projects/cross/opensbi.py
heshamelmatary: Yeah I agree. Hopefully there won't be disruptive/breaking changes in later revisions and the toolchain will be kept in sync with the latest drafts, and the assembly bits in our OSes as well.

pycheribuild/projects/cross/busybox.py
heshamelmatary: I see, it makes sense for CheriBSD given Codasip's HW (and others?) are still on .9.3, but hopefully they will upgrade. I am still not sure about the zcheri vs rvy naming that gives the impression they are separate standards, but hopefully this is just temporary and will get sorted out sooner than later.

pycheribuild/qemu_utils.py
heshamelmatary: The above is `zylevels1b` which is a bit less misleading (given the 1 bit in the name), but his is `Zylevels1` so maybe QEMU should match that naming to avoid confusions? `Zylevels1` to me reads as 1 level.

pycheribuild/config/chericonfig.py
heshamelmatary: Yeah I understand we need a distinction, just wasn't sure about the ISA naming divergence (`experimental-std093` vs `rvy`) that makes them read as different standards.

qwattash: in a sense they kind of are, but I understand what you mean. It would make sense to call it experimental_rvy093 perhaps, so it is clear it is an older version. Is this what you were thinking?

pycheribuild/projects/cross/busybox.py
qwattash: Well zcheri should go away entirely, I hope.

pycheribuild/qemu_utils.py
qwattash: The RVY spec explicitly calls this `Zylevels1`, if I understand correctly, the `b` was added to pacify Clang/LLVM because extensions ending by a number make life harder there. While I agree the `1b` is clearer, I'm not sure how much we should deviate from the spec spelling. My personal opinion is that the Z extensions naming is already so confusing that you have to look up what they mean anyway, but I'm happy to take other people's opinion on this.

pycheribuild/config/chericonfig.py
heshamelmatary: yeah something like that, but not very opinionated if it causes lots of work especially it if will go away

tests/test_git_branch_switching.py
qwattash: Yes, I failed to notice the existing fix. I'm happy to revert it or rebase this PR if #531 lands first.

pycheribuild/config/target_info.py
qwattash: I have not fully duplicated everything of the 093 stuff. In particular, I'm not creating a new sdk directory. Perhaps I should though, at the cost of some duplication. It would certainly be cleaner.

pycheribuild/projects/cross/busybox.py
qwattash: Sure, I can do this.

pycheribuild/config/compilation_targets.py
qwattash: I get an hard error because the target registry finds things with the same name and complains that we have duplicated targets. I could teach it to keep track of this, I'm just not sure how much of a diff that would be.

qwattash: I believe if we go this way it may make sense to just drop the `--riscv-cheri-isa` in favor of explicit target naming. E.g. `cheribsd-riscv64y-purecap` and `cheribsd-riscv64zcheri-purecap`.

tests/test_git_branch_switching.py
qwattash: Removed the commit.

pycheribuild/config/compilation_targets.py
qwattash: This now refactors the target backend fairly significantly to support the explicit targets.

pycheribuild/config/chericonfig.py
```diff
@@ -785,6 +785,10 @@ def default_cheri_sdk_directory_name(self) -> str:
     def default_morello_sdk_directory_name(self) -> str:
         return "morello-sdk"
 
+    @property
+    def default_rvy_sdk_directory_name(self) -> str:
+        return "cheri-rvy-sdk"
```
arichardson, line 790: Maybe just rvy-sdk?

pycheribuild/projects/cross/cheribsd.py
```diff
@@ -340,6 +340,16 @@ def make_all(self) -> "list[CheriBSDConfig]":
         return configs
 
 
+class RISCVStdKernelConfigFactory(KernelConfigFactory):
```
arichardson, line 343: RISCVCheri093KernelConfigFactory(KernelConfigFactory)

Or something like this?

pycheribuild/projects/run_qemu.py
```diff
@@ -283,7 +283,7 @@ def get_chosen_qemu(cls, config: CheriConfig) -> ChosenQEMU:
                 supported_qemu_classes += [BuildUpstreamQEMU, None]
         elif xtarget.is_riscv(include_purecap=True):
             can_provide_src_via_smb = True
-            if xtarget.is_experimental_cheri093_std(config):
+            if xtarget.is_riscv_y(config) or xtarget.is_experimental_cheri093_std(config):
```
arichardson, line 286: We seem to be adding A lot of these y||zcheri093, can we add one helper function that handles both?

pycheribuild/config/compilation_targets.py
```diff
@@ -1404,7 +1404,7 @@ class CompilationTargets(BasicCompilationTargets):
         cheri_isa=RiscvCheriISA.EXPERIMENTAL_STD093,
     )
     CHERIBSD_RISCV_XCHERI_PURECAP = CrossCompileTarget(
-        "riscv64xcheri-purecap",
+        "riscv64-purecap",
```
arichardson, line 1407: Thanks I agree this is the easiest solution for now, adding aliases would be a huge pain. Probably makes sense to squash this one into the original commit that adds the new target names to avoid broken intermediate state.

pycheribuild/config/chericonfig.py
qwattash: good point. I think I also spotted some redundant suffixes around, since the target itself is now riscv64y-cheribsd, there is not point in having riscv64y-cheribsd-build-rvy in the build directory name.

pycheribuild/config/compilation_targets.py
```diff
@@ -312,7 +312,7 @@ def essential_compiler_and_linker_flags_impl(
             result.append(
                 "-mrelax" if _linker_supports_riscv_relaxations(instance.linker, config, xtarget) else "-mno-relax"
             )
-            if xtarget.is_cheri_purecap() and xtarget.is_experimental_cheri093_std(config):
+            if xtarget.is_cheri_purecap() and (xtarget.is_riscv_y() or xtarget.is_experimental_cheri093_std()):
```
arichardson, line 315:
```suggestion
            if xtarget.is_cheri_purecap() and xtarget.is_riscv_y_or_zcheri093():
```

There are so many cases of this I think it makes sense to add a helper.

pycheribuild/config/target_info.py
```diff
@@ -889,7 +889,7 @@ def __init__(
         non_cheri_for_purecap_rootfs_target: "Optional[CrossCompileTarget]" = None,
         hybrid_for_purecap_rootfs_target: "Optional[CrossCompileTarget]" = None,
         purecap_for_hybrid_rootfs_target: "Optional[CrossCompileTarget]" = None,
-        _cheri_isa: Optional[RiscvCheriISA] = None,
+        cheri_isa: Optional[RiscvCheriISA] = None,
```
arichardson, line 892:
```suggestion
        riscv_cheri_isa: Optional[RiscvCheriISA] = None,
```

Maybe this to make it clear that it's riscv only?

arichardson (commented):
A few minor comments, but this looks good to me. Much cleaner than the curretnt solution.

pycheribuild/projects/cross/cheribsd.py
qwattash: Fixed.

pycheribuild/config/target_info.py
qwattash: Done.

pycheribuild/config/compilation_targets.py
qwattash: Added the helper.

pycheribuild/config/compilation_targets.py
qwattash: Done

pycheribuild/config/target_info.py
```diff
@@ -959,55 +971,41 @@ def _set_for(self, other_target: "Optional[CrossCompileTarget]", also_set_other=
...
-                    other_target._purecap_for_hybrid_rootfs_target = self
-                    self._purecap_for_hybrid_rootfs_target = self
+                    # assert self not in other_target._purecap_for_hybrid_rootfs_target, "Already set?"
+                    other_target._purecap_for_hybrid_rootfs_target.add(self)
+                    self._purecap_for_hybrid_rootfs_target.add(self)
                 else:
-                    assert other_target._purecap_target is None or other_target._purecap_target is self, "Already set?"
-                    other_target._purecap_target = self
-                    self._purecap_target = self
+                    # assert self not in other_target._purecap_target, "Already set?"
```
arichardson: I think we can just delete these comments?

pycheribuild/config/defaultconfig.py
```diff
@@ -242,6 +242,16 @@ def __init__(self, loader: "ConfigLoaderBase[typing.Self]", available_targets: "
             group=loader.path_group,
             help="The directory to find/install the Morello SDK",
         )
+        default_rvy_sdk = ComputedDefaultValue(
+            function=lambda p, cls: p.tools_root / p.default_rvy_sdk_directory_name,
+            as_string="'<TOOLS_ROOT>/cheri-rvy-sdk'",
+        )
+        self.cheri_rvy_sdk_dir = loader.add_path_option(
+            "cheri-rvy-sdk-root",
```
arichardson: I think I'd do a global cheri-rvy -> rvy search replace but don't feel strongly about this

arichardson (approved):
Two minor comments otherwise I think we should go and merge this once CI passes!

pycheribuild/config/target_info.py
qwattash: Indeed, they slipped through, sorry.

pycheribuild/config/defaultconfig.py
qwattash: Done

pycheribuild/config/target_info.py
arichardson, line 1086: this looks wrong, but might be fixed in a later commit.

arichardson: Oh I see this should really be a TODO comment, please ignore.

pycheribuild/projects/cross/opensbi.py
```diff
@@ -187,7 +187,9 @@ def _qemu_fw_install_path(self, is_payload=False) -> Path:
 
         if self.crosscompile_target.is_cheri_purecap():
             suffix = "cheri"
-            if self.crosscompile_target.is_experimental_cheri093_std(self.config):
+            if self.crosscompile_target.is_riscv_y(
+                self.config
+            ) or self.crosscompile_target.is_experimental_cheri093_std(self.config):
                 suffix += "std"
```
arichardson, line 193: can we add a y suffix or is that not possible in the current build system?

pycheribuild/config/compilation_targets.py
```diff
@@ -1469,7 +1495,7 @@ class CompilationTargets(BasicCompilationTargets):
         extra_target_suffix="-for-purecap-rootfs",
         is_cheri_hybrid=True,
         rootfs_target=CHERIBSD_RISCV_ZCHERI093_PURECAP,
-        purecap_target=CHERIBSD_RISCV_ZCHERI093_PURECAP,
+        # purecap_target=CHERIBSD_RISCV_ZCHERI093_PURECAP,
```
arichardson, line 1498: why is this commented out?

pycheribuild/config/compilation_targets.py
```diff
@@ -1459,7 +1485,7 @@ class CompilationTargets(BasicCompilationTargets):
         extra_target_suffix="-for-purecap-rootfs",
         is_cheri_hybrid=True,
         rootfs_target=CHERIBSD_RISCV_Y_PURECAP,
-        purecap_target=CHERIBSD_RISCV_Y_PURECAP,
+        # purecap_target=CHERIBSD_RISCV_Y_PURECAP,
```
arichardson, line 1488: why is this commented out?

pycheribuild/config/target_info.py
```diff
@@ -1150,22 +1138,36 @@ def is_libcompat_target(self) -> bool:
...
         if self._is_cheri_hybrid and self._rootfs_target is None:
             return self
-        elif self._hybrid_target is not None:
-            return self._hybrid_target
+        if self.is_hybrid_or_purecap_cheri([CPUArchitecture.RISCV64, CPUArchitecture.RISCV32]):
+            for t in self._hybrid_target:
+                if t.riscv_cheri_isa() == self.riscv_cheri_isa():
+                    return t
+        elif len(self._hybrid_target) == 1:
+            return set(self._hybrid_target).pop()
```
arichardson, line 1146: next(iter(self._hybrid_target))

pycheribuild/config/target_info.py
```diff
@@ -1150,22 +1138,36 @@ def is_libcompat_target(self) -> bool:
...
         if self._is_cheri_purecap and self._rootfs_target is None:
             return self
-        elif self._purecap_target is not None:
-            return self._purecap_target
+        if self.is_hybrid_or_purecap_cheri([CPUArchitecture.RISCV64, CPUArchitecture.RISCV32]):
+            for t in self._purecap_target:
+                if t.riscv_cheri_isa() == self.riscv_cheri_isa():
+                    return t
+        elif len(self._purecap_target) == 1:
+            return set(self._purecap_target).pop()
```
arichardson, line 1159: next(iter(self._purecap_target))

pycheribuild/config/target_info.py
```diff
@@ -1150,22 +1138,36 @@ def is_libcompat_target(self) -> bool:
...
 
     def get_non_cheri_target(self) -> "CrossCompileTarget":
         if not self._is_cheri_purecap and not self._is_cheri_hybrid and self._rootfs_target is None:
             return self
-        elif self._non_cheri_target is not None:
-            return self._non_cheri_target
+        elif self._non_cheri_target:
+            # There can only be one non-cheri associated to a hybrid or purecap target
+            assert len(self._non_cheri_target) == 1, "Too many non-cheri targets"
+            return set(self._non_cheri_target).pop()
```
arichardson, line 1170: next(iter(self._non_cheri_target))

pycheribuild/config/target_info.py
```diff
@@ -1176,8 +1178,15 @@ def get_non_cheri_for_hybrid_rootfs_target(self) -> "CrossCompileTarget":
...
-            return self._non_cheri_for_hybrid_rootfs_target
+        if self.is_riscv(include_purecap=True):
+            # Distinguis between rootfs ISAs
+            for t in self._non_cheri_for_hybrid_rootfs_target:
+                assert t._rootfs_target is not None
+                if t._rootfs_target.riscv_cheri_isa() == self.riscv_cheri_isa():
+                    return t
+        elif self._non_cheri_for_hybrid_rootfs_target:
+            assert len(self._non_cheri_for_hybrid_rootfs_target) == 1
+            return set(self._non_cheri_for_hybrid_rootfs_target).pop()
```
arichardson, line 1189: next(iter(self._non_cheri_for_hybrid_rootfs_target))

pycheribuild/config/target_info.py
```diff
@@ -1188,22 +1197,39 @@ def get_non_cheri_for_purecap_rootfs_target(self) -> "CrossCompileTarget":
...
-            return self._non_cheri_for_purecap_rootfs_target
+        if self.is_riscv(include_purecap=True):
+            # Distinguis between rootfs ISAs
+            for t in self._non_cheri_for_purecap_rootfs_target:
+                assert t._rootfs_target is not None
+                if t._rootfs_target.riscv_cheri_isa() == self.riscv_cheri_isa():
+                    return t
+        elif self._non_cheri_for_purecap_rootfs_target:
+            assert len(self._non_cheri_for_purecap_rootfs_target) == 1
+            return set(self._non_cheri_for_purecap_rootfs_target).pop()
```
arichardson, line 1208: next(iter(self._non_cheri_for_purecap_rootfs_target))

pycheribuild/config/target_info.py
```diff
@@ -1188,22 +1197,39 @@ def get_non_cheri_for_purecap_rootfs_target(self) -> "CrossCompileTarget":
...
             return self
-        elif self._purecap_for_hybrid_rootfs_target is not None:
-            return self._purecap_for_hybrid_rootfs_target
+        if self.is_riscv(include_purecap=True):
+            for t in self._purecap_for_hybrid_rootfs_target:
+                if t.riscv_cheri_isa() == self.riscv_cheri_isa():
+                    return t
+        elif self._purecap_for_hybrid_rootfs_target:
+            assert len(self._purecap_for_hybrid_rootfs_target) == 1
+            return set(self._purecap_for_hybrid_rootfs_target).pop()
```
arichardson, line 1232: next(iter(self._purecap_for_hybrid_rootfs_target))

pycheribuild/config/target_info.py
```diff
@@ -1188,22 +1197,39 @@ def get_non_cheri_for_purecap_rootfs_target(self) -> "CrossCompileTarget":
...
             return self
-        elif self._hybrid_for_purecap_rootfs_target is not None:
-            return self._hybrid_for_purecap_rootfs_target
+        if self.is_riscv(include_purecap=True):
+            for t in self._hybrid_for_purecap_rootfs_target:
+                if t.riscv_cheri_isa() == self.riscv_cheri_isa():
+                    return t
+        elif self._hybrid_for_purecap_rootfs_target:
+            assert len(self._hybrid_for_purecap_rootfs_target) == 1
+            return set(self._hybrid_for_purecap_rootfs_target).pop()
```
arichardson, line 1220: next(iter(self._hybrid_for_purecap_rootfs_target))

arichardson (approved):
I will make these minor changes then merge. Thank you for this great PR!

PR #536, opened 2026-09-09
Fix installation directories for cheri-os-test's CMake targets

paul-metzger:
@arichardson, 

as I mentioned to you, @brooks would like cheri-os-test to be installed into
```
/usr/local/bin/
/usr/local/lib/
/usr/local/libexec/instead of
```
instead of
```
/usr/local/riscv64-purecap/bin/
/usr/local/riscv64-purecap/lib/
/usr/local/riscv64-purecap/libexec/
```
on CheriBSD. The build for Linux needs to use a different directory too, specifically a subdirectory of rootfs.

I also created a draft for a new installation directory kind for this situation, as suggested by you. However, distinguishing between two 'local' installation directory kinds, one with an ABI specific subdirectory and one without, only makes sense for some compilation targets. The solution that I found to support both compilation target types feels hacky. I'm referring specifically to [this](https://github.com/CTSRD-CHERI/cheribuild/commit/ca56f3b3cfd52473bdd2a2ed1032f51e9a41fd70#diff-c3a94bb0138716ffa300bebb1b45903e4b7f64f9d35e65fb63d24517a92fc0baR979-R980). The property `sysroot_install_prefix_relative_no_abi_dir` is only defined by compilation targets for which the distinction make sense. It would be great to get your feedback on this.

Changing `ROOTFS_LOCALBASE` to mean `/usr/local` on CheriBSD and introducing a `ROOTFS_LOCALBASE_WITH_ABI_DIR` installation target would also make sense, but this will require reviewing and potentially testing targets that support CheriBSD and use `ROOTFS_LOCALBASE`, to make sure they don't break.


pycheribuild/config/compilation_targets.py
```diff
@@ -739,6 +741,10 @@ def base_sysroot_targets(cls, target: "CrossCompileTarget", config: "CheriConfig
     def sysroot_install_prefix_relative(self) -> Path:
         return Path("usr/local", self.install_prefix_dirname)
 
+    @property
+    def sysroot_install_prefix_relative_no_abi_dir(self) -> Path:
```
arichardson:
```suggestion
    def sysroot_install_prefix_relative_without_abi_subdir(self) -> Path:
```

Maybe this more explicit name is better?

pycheribuild/config/compilation_targets.py
```diff
@@ -169,6 +169,8 @@ def default_install_dir(self, install_dir: DefaultInstallDir) -> Path:
             return Path(self._rootfs_path(), "opt", self.install_prefix_dirname, "kde")
         elif install_dir == DefaultInstallDir.ROOTFS_LOCALBASE:
             return self.sysroot_dir
+        elif install_dir == DefaultInstallDir.ROOTFS_LOCALBASE_NO_ABI_DIR:
```
arichardson:
```suggestion
        elif install_dir == DefaultInstallDir.ROOTFS_LOCALBASE_WITHOUT_ABI_SUBDIR:
```

pycheribuild/projects/project.py
```diff
@@ -975,6 +975,13 @@ def __init__(self, *args, **kwargs) -> None:
             if install_dir_kind == DefaultInstallDir.ROOTFS_LOCALBASE:
                 self._install_prefix = Path("/", self.target_info.sysroot_install_prefix_relative)
                 self.destdir = self._install_dir
+            elif install_dir_kind == DefaultInstallDir.ROOTFS_LOCALBASE_NO_ABI_DIR:
+                if hasattr(self.target_info, "sysroot_install_prefix_relative_no_abi_dir"):
```
arichardson: I think it would be cleaner to just add a default implmentation of this method to the base class that calls `sysroot_install_prefix_relative`

pycheribuild/projects/cross/cheri_os_test.py
```diff
@@ -116,19 +118,29 @@ def process(self):
 
 
 class BuildCheriOSTestCmake(CrossCompileCMakeProject):
-    _supported_architectures = (
-        CompilationTargets.ALL_LINUX_PURECAP_TARGETS + CompilationTargets.ALL_CHERIBSD_PURECAP_TARGETS
-    )
+    _supported_architectures = (CompilationTargets.ALL_CHERIBSD_PURECAP_TARGETS)
```
arichardson, line 121: Why is this change needed? Can't we allow it for both Linux and FreeBSD?

pycheribuild/config/compilation_targets.py
paul-metzger: Yes, that's better - done.

pycheribuild/config/compilation_targets.py
paul-metzger: Done

pycheribuild/projects/project.py
paul-metzger: Yes, of course - done.

pycheribuild/projects/cross/cheri_os_test.py
paul-metzger: Linux needs a different installation directory (see [here](https://github.com/CTSRD-CHERI/cheribuild/pull/536/changes#diff-b9925a51c0a8a4d820cd719e75a195c35c0f6a5fdb898443acb9fb565e78419eR141-R146)). Isn't creating a separate Linux-specific class the most idiomatic way to implement this in cheribuild, instead of adding branches for per-compilation-target logic to a single class?

PR #537, opened 2026-09-10
cheri-std093-llvm: use CHERI Alliance's repo

heshamelmatary:
PR #29 is now merged with the fix the temp branch had. https://github.com/CHERI-Alliance/llvm-project/pull/29


jrtc27 (approved):
Thanks

PR #539, opened 2026-09-17
OpenSBI: explicitly disable RELRO

heshamelmatary:
OpenSBI executes in M-mode and performs its own relocations without a loader to enforce PT_GNU_RELRO afterwards. Disable RELRO explicitly since the generated RELRO metadata would not provide protection by itself.

This PR also un-breaks CI after the latest [CHERI-Alliance LLVM-18 merge ](https://github.com/CHERI-Alliance/llvm-project/pull/31)as OpenSBI fails with the following error:

```
ld.lld: error: section: .got is not contiguous with other relro sections
ld.lld: error: section: .pad.cheri.pcc is not contiguous with other relro sections
```


arichardson (commented):
No opinion on whether this is the right solution since I've not looked into it. The cheribuild change is fine but I'd want @jrtc27 to approve this one

jrtc27:
I've gone and investigated this properly, and the real issue is a bunch of things colluding to cause rather dodgy layout and PCC bounds within OpenSBI after the Codasip LLVM merge. Disabling relro squashes this error but does nothing to fix the dodgy layout. Manually laying out .got (and .dynamic? though not convinced that one is necessarily required) gets a more sensible layout, but does nothing to fix the PCC bounds. As a way to move forwards whilst I work through the changes to get sensible layout and PCC bounds in OpenSBI, this approach is fine, but this is just that, and it should be removed once no longer needed, thus **needs an appropriate comment**.

There is a separate question as to whether upstream OpenSBI should build with -znorelro or other such flags to get a more efficient layout, but that's not our problem to solve. Given upstream's... choice... to put .data and .bss on separate pages it's not clear they are all that aware of how to lay out ELF binaries in efficient ways and/or don't care about this kind of thing.

heshamelmatary:
@jrtc27 Thanks for investigating this. I don't think upstream cares about paging as OpenSBI executes in M-Mode without paging? Apart from fixing the error above, I do think upstream OpenSBI should explicitly build with -z norelro for intentionality; RELRO will still be useless and can't enforce perms in M-Mode anyway as things currently stand (i.e., no paging, no PMP enforcement in M-Mode, no loader before OpenSBI, etc)

jrtc27:
Well, it *could* enforce it with the PMP if it wanted to burn another entry... but that's probably an ePMP thing?

jrtc27:
And besides, that just gets to my point about it being a separate upstream question that we shouldn't be patching, and even if CHERI OpenSBI wants to do that, it should be done there, not in cheribuild. This is only here in cheribuild as a temporary workaround.

jrtc27:
Given we already override the OpenSBI branch, I'd also be ok with just chucking a linker script diff into that branch instead of working around it in cheribuild. Either way, whatever changes are made should be documented as temporary (although the linker script ones might be sensible to have anyway rather than relying on orphan section placement, but with the changes I'm working on won't be strictly necessary, and so can be dropped again to reduce upstream divergence).

heshamelmatary:
yeah it could, though they don't currently use the PMP to enforce protection in M-Mode, but rather just to protect itself from H/S modes. Anyway, I will add a comment to point it this (norelro) is temporary in cheribuild

heshamelmatary:
@jrtc27 I added a comment to make it temporary

pycheribuild/projects/cross/opensbi.py
```diff
@@ -93,6 +93,12 @@ def setup(self):
         super().setup()
         compflags = " " + self.commandline_to_str(self.essential_compiler_and_linker_flags)
         compflags += " -Qunused-arguments"  # -mstrict-align -no-pie
+        # OpenSBI performs its own relocations in M-mode and has no loader
+        # to enforce PT_GNU_RELRO after relocation. The resulting RELRO
+        # metadata therefore does not provide protection by itself.
+        # TODO: this fix temporarily fixes a CI build failure after an LLVM
+        # merge; remove when fixed in LLVM.
+        compflags += " -z norelro"
```
jrtc27, line 101: Huh, TIL -z exists as a compiler driver flag that gets passed through to the linker. That's one of the things I'd always prefix with `-Wl,`.

jrtc27 (approved):

PR #544, opened 2026-09-21
CHERI Alliance targets cleanups and fixes

heshamelmatary:
This makes CI almost happy (with the relro temp fix https://github.com/CTSRD-CHERI/cheribuild/pull/539).


pycheribuild/projects/cross/cheri_microkit.py
```diff
@@ -338,10 +337,10 @@ def dependencies(cls, config: CheriConfig) -> "tuple[str, ...]":
...
         result += ("cheri-microkit",)
         if cls.get_crosscompile_target().is_hybrid_or_purecap_cheri([CPUArchitecture.RISCV64]):
-            result += ("cheri-std093-llvm",)
-            result += ("cheri-std093-opensbi",)
-            result += ("cheri-std093-gdb-native",)
-            result += ("cheri-std093-qemu",)
+            result += ("alliance-llvm",)
+            result += ("alliance-opensbi",)
+            result += ("alliance-gdb-native",)
+            result += ("alliance-qemu",)
```
arichardson: hmm I think we don't need the alliance qemu anymore.

heshamelmatary: I removed it in the last commit and used QEMU instead. However, the CHERI-seL4 expect `qemu-system-riscv64` to be in the PATH (to extract the DTB from), and this is causing an error as building `cheri-microkit-riscv64zcheri093` doesn't add output/sdk in the PATH, only output/alliance-sdk

heshamelmatary: I am now leaning to keep `alliance-qemu` to keep all SDK and tools in one `alliance-sdk` directory instead of having to deal with both `sdk` and `alliance-sdk`

arichardson:
Alliance- as a prefix looks a bit odd to me, maybe we should be verbose and use cheri-alliance?

heshamelmatary:
> Alliance- as a prefix looks a bit odd to me, maybe we should be verbose and use cheri-alliance?

My personal preference was to keep names as short as possible and since this is cheribuild, I went for alliance just as an identifier rather than a recognition/marketing. We also currently mix *CheriAlliance* and *Alliance* in class names for that reason as well (to keep it shorter). For example, `cheri-alliance-cheri-linux-riscv64zcheri093-purecap` to build with `cheribuild` has too many `cheri`s and is rather long

I am happy to go for cheri-alliance- prefix still

